PR #1, opened 2015-11-10
Sprint Solutions:Zug PHX-200 PHX-201 XBlock LTI Consumer

douglashall:
This PR migrates the LTI XModule component which lives in edx-platform (https://github.com/edx/edx-platform/blob/master/common/lib/xmodule/xmodule/lti_module.py) to this new repository and converts the component to the XBlock API.

There is additional refactoring of the LTI XModule code to increase code quality and improve extensibility for when/if we want to add additional LTI spec functionality.

The modal window feature described in PHX-200 is also implemented as a part of this PR.

Related PRs:
https://github.com/edx/xblock-utils/pull/32
https://github.com/edx/edx-platform/pull/10600

Sandbox URLs:
https://lti-consumer.sandbox.edx.org/courses/course-v1:edX+Test101+course/courseware/039df3d8458844e68635ebe68681d9d3/0927a6a523e7435aa5a1ad07e901ccc8/
https://studio-lti-consumer.sandbox.edx.org/container/block-v1:edX+Test101+course+type@vertical+block@f0c6a0e3b41643a58b4302d2e01cb925


douglashall:
@chrisndodge

LICENSE
```diff
@@ -0,0 +1,671 @@
+                    GNU AFFERO GENERAL PUBLIC LICENSE
```
chrisndodge: You might want to check in with @nedbat or @jbarciauskas about this LICENSE. I'm not sure if it is our policy to mark our authored xBlocks as AGPL or Apache...

requirements.txt
```diff
@@ -0,0 +1,17 @@
+ddt
+lxml
+mock
+selenium
+pep8
+pylint
+coverage==3.7
+coveralls
+django==1.4.22
```
chrisndodge: Just be aware that we are in the midst of a massive Django 1.8 upgrade so we'll have to swing back to this. FYI @nedbat @macdiesel

README.rst
```diff
@@ -0,0 +1,85 @@
+LTI Consumer XBlock |Build Status|
```
chrisndodge: Can you add in the Travis/Coveralls status indicators up at the top. You can look over at edx-proctoring/README to see how that is done.

lti_consumer/lti_consumer.py
```diff
@@ -0,0 +1,817 @@
...
+
+    @property
+    def context_id(self):
+        """
+        Return context_id.
+
+        context_id is an opaque identifier that uniquely identifies the context (e.g., a course)
+        that contains the link being launched.
+        """
+        return self.course_id.to_deprecated_string()
```
chrisndodge: I think the new convention is to just unicode(self.course_id) now

lti_consumer/lti_consumer.py
```diff
@@ -0,0 +1,817 @@
...
+                _ = self.runtime.service(self, "i18n").ugettext
+                msg = _('Could not parse LTI passport: {lti_passport}. Should be "id:key:secret" string.').format(
+                    lti_passport='{0!r}'.format(lti_passport)
+                )
+                raise LtiError(msg)
+
+            if lti_id == self.lti_id.strip():
+                return key, secret
+
+        return '', ''
```
chrisndodge: I think Dave O was suggesting we come up with a means to have these secrets not have to be stored in the clear in our courseware. I'm not sure I have a proposal for you, but I just wanted to bring it up.

LICENSE
jbarciauskas: I think we want to default to apache wherever possible

README.rst
douglashall: Thanks. I have added the status indicators.

lti_consumer/lti_consumer.py
douglashall: Good to know. I have make this change.

lti_consumer/lti_consumer.py
douglashall: @chrisndodge I will create a Jira task to retire that debt.

douglashall:
@doctoryes @sarina Could you or someone from your team take a look at this and the related PRs when you get a chance?

chrisndodge:
@douglashall looking at the sandbox, the "action" button will likely need to get some updated styling as I believe the UX team has a standard set of button stylings. I'm not sure if the policy is to simply define classes on the HTML in an xBlock and assume that the style definition will be defined from the host LMS runtime container app. Or if xBlocks should manage its own style definitions through it's own .css.

FYI, @frrrances @talbs

douglashall:
@benpatterson Could you take a look at this from a testing perspective when you get a chance? Unit tests should be good. I'm still working on getting some integration tests up and running.

.pylintrc
```diff
@@ -0,0 +1,27 @@
...
+	maybe-no-member,
+	protected-access,
+	star-args,
+	too-few-public-methods,
+	too-many-ancestors,
+	too-many-arguments,
+	too-many-branches,
+	too-many-locals,
+	too-many-public-methods,
+	too-many-statements,
```
benpatterson: We should not disable this linter in particular. The point of that linter is not just clean code, but good code. too-many-statements means that the cyclomatic complexity of the method is very high. We should avoid that pattern & therefore it should be removed from this file so it can be enforced.

.travis.yml
```diff
@@ -0,0 +1,22 @@
...
+  - "2.7"
+
+before_install:
+  - "export DISPLAY=:99"
+  - "sh -e /etc/init.d/xvfb start"
+
+install:
+  - "pip install -r requirements.txt"
+  - "pip install -r test_requirements.txt"
+  - "pip install -r $VIRTUAL_ENV/src/xblock-sdk/requirements/base.txt"
```
benpatterson: Is there a reason these requirements aren't just captured in the test_requirements.txt? similarly, is XBlock SDK not on pip? I would think you'd just install XBlock SDK and it would include whatever you need to develop an XBlock. That would include testing it, since testing is a part of development.

README.rst
```diff
@@ -0,0 +1,91 @@
...
+Access it at `http://localhost:8000/ <http://localhost:8000>`__.
+
+Running tests
+-------------
+
+From the lti_consumer directory, run the tests with the following command:
+
+.. code:: bash
+
+    $ DJANGO_SETTINGS_MODULE="workbench.settings" nosetests tests/*
```
benpatterson: This appears to conflict slightly with what is contained in the .travis.yml file.

benpatterson:
@douglashall I've looked at it from the perspective of travis config, etc, like we talked about last week. 

I think for pylint you want to adopt what is used in the edx-platform repo, and potentially the auto-generated part of it. I'm less familiar with how to auto-generate the edx-specific linting file..perhaps that's something @sarina or @singingwolfboy could comment on. However, just taking the one on edx-platform and retrofitting it for this repo might be a good idea.

Taking a look at how you've structured your tests, etc, will take me more time (~2 days due to scheduling on my side). (It is generally something you'd rely on your team to do; however, I also understand the value since this is your first PR.) Did you want me to look at that as well?

.travis.yml
```diff
@@ -0,0 +1,22 @@
+language: python
+
+python:
+  - "2.7"
+
```
benpatterson: you'll also want a 'branches' stanza so that you don't get two builds for every pull request. See example [here](https://github.com/edx/edx-analytics-data-api/blob/master/.travis.yml#L12-L14)

benpatterson: You also want to add a `sudo: false` line. That will make these builds run on Travis's docker infrastructure, which will be much faster.

lti_consumer/templates/html/student.html
```diff
@@ -0,0 +1,91 @@
...
+% endif
+
+<div
+    id="${element_id}"
+    class="${element_class}"
+    data-ask-to-send-username="${ask_to_send_username}"
+    data-ask-to-send-email="${ask_to_send_email}"
+>
+
+% if launch_url and launch_url !=  'http://www.example.com' and not hide_launch:
```
chrisndodge: was this logic particularly around launch_url != 'http://www.example.com' part of the xModule implementation? I'm guessing that URL is the field default. If so, I think it might be cleaner to scrub that in the python side, e.g. perhaps set launch_url to None and then keep this Mako conditional just be %if launch_url:

douglashall:
@benpatterson Thanks for the initial review of the travis setup. I have made changes based on your feedback. I would appreciate a review of the test structure for the first PR, if you have time. Thank you in advance.

lti_consumer/templates/html/student.html
douglashall: Yes, this was copied over. I have removed the default value from launch_url so that we could simplify this logic.

sarina:
> I think for pylint you want to adopt what is used in the edx-platform repo, and potentially the auto-generated part of it. I'm less familiar with how to auto-generate the edx-specific linting file

I'm in the same boat. You should talk to Ned, or maybe Cale, about edx-lint.

sarina:
@douglashall I can get to a review of this in the later half of next week, will that be ok?

smarnach:
@douglashall Nice to see this work!  I've tested this on the sandbox with an LTI component we wrote for Harvard, and it works fine.  I've got a few comments on the Studio interface:

* Adding custom parameters to the LTI consumer XBlock.  The interface was nicer in the old version, where you could edit the list more interactively.  I don't think the difference matters much.  However, it would be nice to give a hint in what format the key/value pairs are supposed to be entered directly in the help text;  it's probably easiest to give an example, like `["key1=value1", "key2=value2"]`.

* The "Modal Height/Width" fields. The names are misleading.  They can also be used to change the height of the iframe when using inline display (which is an important feature to have).  The documentation is also misleading.  It states "This value must be expressed as a percentage of the browser viewport height", while the value actually seems to be the height in pixels (at least for an iframe, which is what I tested).

* The outcome URL transferred to the LTI provider.  When used in Studio, the outcome URL is of the form `/preview/xblock/block-v1:...`, i.e. it's not an absolute URL including protocol and server name.  The old LTI comsumer seems to have behaved differently.  All works fine in the LMS.

I'll do a pass over the code now.

.travis.yml
```diff
@@ -0,0 +1,18 @@
+language: python
+python: "2.7"
+before_install:
+  - "export DISPLAY=:99"
+  - "sh -e /etc/init.d/xvfb start"
```
smarnach: Are you planning to add integrations tests as well?  If not, ther's no point in starting xvfb before running the tests.

lti_consumer/lti.py
```diff
@@ -0,0 +1,288 @@
...
+        }
+
+        key, secret = self.xblock.lti_provider_key_secret
+        oauth_signature = get_oauth_request_signature(key, secret, self.xblock.launch_url, headers, lti_parameters)
+
+        # Parse headers to pass to template as part of context:
+        oauth_signature = dict([param.strip().replace('"', '').split('=') for param in oauth_signature.split(',')])
+
+        oauth_signature[u'oauth_nonce'] = oauth_signature[u'OAuth oauth_nonce']
+        del oauth_signature[u'OAuth oauth_nonce']
```
smarnach: Style nit: You could simply use

    oauth_signature[u'oauth_nonce'] = oauth_signature.pop(u'OAuth oauth_nonce')

To get the value and remove the key from the dictionary in one go.

lti_consumer/lti.py
```diff
@@ -0,0 +1,288 @@
...
+
+        # Username and email can't be sent in studio mode, because the user object is not defined.
+        # To test functionality test in LMS
+
+        if callable(self.xblock.runtime.get_real_user):
+            real_user_object = self.xblock.runtime.get_real_user(self.xblock.runtime.anonymous_student_id)
+            try:
+                self.xblock.user_email = real_user_object.email
+            except AttributeError:
+                self.xblock.user_email = ""
```
smarnach: Style nit:  how about

    xblock.user_email = getattr(real_user, "email", "")

instead?

lti_consumer/lti.py
```diff
@@ -0,0 +1,288 @@
...
+            try:
+                self.xblock.user_username = real_user_object.username
+            except AttributeError:
+                self.xblock.user_username = ""
```
smarnach: Same here.

lti_consumer/lti_consumer.py
```diff
@@ -0,0 +1,829 @@
...
+    'tool_consumer_info_version',
+    'tool_consumer_instance_guid',
+    'tool_consumer_instance_name',
+    'tool_consumer_instance_description',
+    'tool_consumer_instance_url',
+    'tool_consumer_instance_contact_email',
+]
+
+
+LaunchTargetOption = namedtuple('LaunchTargetOption', ['display_name', 'value'])
```
smarnach: For better readability, I suggest moving this line down to just before the definition of the `LaunchTarget` class.

lti_consumer/lti_consumer.py
```diff
@@ -0,0 +1,829 @@
...
+    )
+
+    @property
+    def descriptor(self):
+        """
+        Returns this XBlock object.
+
+        This is for backwards compatibility with the XModule API.
+        Some LMS code still assumes a descriptor attribute on the XBlock object.
+        See courseware.module_render.rebind_noauth_module_to_user.
```
smarnach: I don't think we ever had to add this to an XBlock.  If this is only required in `rebind_noauth_module_to_user()`, it might make sense to fix that function instead.  If there are more places where this is still needed, we could add the function to the `XBlock` base class, so we only have to remove it in a single place after edx-platform is fixed.

lti_consumer/lti_consumer.py
```diff
@@ -0,0 +1,829 @@
...
+        This is an opaque unique identifier that the TC guarantees will be unique
+        within the TC for every placement of the link.
+
+        If the tool / activity is placed multiple times in the same context,
+        each of those placements will be distinct.
+
+        This value will also change if the item is exported from one system or
+        context and imported into another system or context.
+
+        This parameter is required.
```
smarnach: I don't understand this sentence.  What paramter?

lti_consumer/lti_consumer.py
```diff
@@ -0,0 +1,829 @@
...
+        """
+        user_id = self.runtime.anonymous_student_id
+        if user_id is None:
+            raise LtiError("Could not get user id for current request")
+        return unicode(urllib.quote(user_id))
+
+    @property
+    def resource_link_id(self):
+        """
+        This is an opaque unique identifier that the TC guarantees will be unique
```
smarnach: What is TC?

lti_consumer/lti_consumer.py
```diff
@@ -0,0 +1,829 @@
...
+        return fragment
+
+    @XBlock.handler
+    def lti_launch_handler(self, request, suffix=''):  # pylint: disable=unused-argument
+        """
+        XBlock handler for launching the LTI provider.
+
+        Displays a form which is submitted via Javascript
+        to send the LTI launch POST request to the LTI
+        provider.
```
smarnach: What is the rationale of using a fake form here, as opposed to sending the POST request directly from JavaScript?

lti_consumer/lti_consumer.py
```diff
@@ -0,0 +1,829 @@
...
+            user (django.contrib.auth.models.User):  Actual user whose module state is to be set
+            score (float):  user's numeric score to set.  Must be in the range [0.0, 1.0]
+            max_score (float):  max score that could have been achieved on this module
+            comment (unicode):  comments provided by the grader as feedback to the student
+
+        Returns:
+            nothing
+        """
+        if score is not None and max_score is not None:
+            scaled_score = score * max_score
```
smarnach: I think the score in the published event is the raw score, which is later scaled with the weight automatically.  See https://openedx.atlassian.net/wiki/display/TNL/edX+Grading+Guide+for+Developers to make sure that I remember correctly. :)

benpatterson:
@douglashall have you considered a Makefile? A lot of the "run tests this way" or "install" kinds of instructions would be easier for developers if they were `make` targets.

lti_consumer/static/css/student.css
```diff
@@ -0,0 +1,41 @@
+.xblock-student_view.xblock-student_view-lti_consumer h2.problem-header {
```
smarnach: This seems to be generated from the SCSS source below.  Do we need to check this file in to version control?

benpatterson: Ah I was going to make a similar comment, so there's another vote for you, @douglashall,  in that particular improvement.

lti_consumer/tests/unit/test_utils.py
```diff
@@ -0,0 +1,53 @@
...
+    xblock.category = Mock()
+    xblock.location = Mock(
+        html_id=Mock(return_value='sample_element_id'),
+    )
+    xblock.runtime = Mock(
+        hostname='localhost',
+    )
+    xblock.course_id = 'course-v1:edX+DemoX+Demo_Course'
+    for key, value in attributes.iteritems():
+        setattr(xblock, key, value)
```
smarnach: I'd write this loop as `vars(xblock).update(attributes)`.

lti_consumer/tests/unit/test_utils.py
```diff
@@ -0,0 +1,53 @@
...
+from webob import Request
+from mock import Mock
+
+from xblock.fields import ScopeIds
+from xblock.runtime import KvsFieldData, DictKeyValueStore
+
+from workbench.runtime import WorkbenchRuntime
+
+
+def make_xblock(xblock_name, xblock_cls, attributes):
```
smarnach: Nice!  I think we should add a version of this function (and the next one) to xblock-utils.

smarnach:
@douglashall I'm done with my pass over the code now.  Looks good!  I've added a few minor comments.

.gitignore
```diff
@@ -0,0 +1,6 @@
+*.egg-info
```
benpatterson: We should add IDE-kinds of files, like those in the [platform](https://github.com/edx/edx-platform/blob/master/.gitignore#L13-L21).

benpatterson:
@douglashall you may also want to use jshint. I can point you to some folks that have put that in place in other repos. Would be good to get that as part of CI now rather than later.

test_requirements.txt
```diff
@@ -0,0 +1,5 @@
+coveralls
```
benpatterson: you should start this with `-r requirements.txt` so someone preparing for development will get everything s/he needs with one target.

lti_consumer/lti.py
```diff
@@ -0,0 +1,288 @@
...
+    present, in which case it must be a string.
+
+    Arguments:
+        json_str (unicode):  The body of the LTI 2.0 results service request, which is a JSON string]
+
+    Returns:
+        (float, str):  (score, [optional]comment) if verification checks out
+
+    Raises:
+        LtiError (with message) if verification fails
```
benpatterson: you can remove `(with message)`. That would be implied IMO.

lti_consumer/tests/unit/test_lti.py
```diff
@@ -0,0 +1,287 @@
...
+    "@type": "Result",
+}
+
+
+class TestParseResultJson(unittest.TestCase):
+    """
+    Unit tests for `lti_consumer.lti.parse_result_json`
+    """
+
+    def test_invalid_json(self):
```
benpatterson: You should use ddt for these tests. There are examples in edx-platform. ddt has a few different ways to use data...here is [one example file](https://github.com/edx/edx-platform/blob/master/lms/djangoapps/discussion_api/tests/test_permissions.py); however, I suggest you [read the docs](https://ddt.readthedocs.org/en/latest/).

lti_consumer/tests/unit/test_lti.py
```diff
@@ -0,0 +1,287 @@
...
+class TestLti1Consumer(TestLtiConsumerXBlock):
+    """
+    Unit tests for Lti1Consumer
+    """
+
+    def setUp(self):
+        super(TestLti1Consumer, self).setUp()
+        self.lti_consumer = Lti1Consumer(self.xblock)
+
+    @patch('lti_consumer.lti.get_oauth_request_signature', Mock(
```
benpatterson: Are these patch values made-up? They look pretty real. I'd make them more-obviously-fake. The timestamp also has me scratching my head...should you use datetime.now() or something for that?

lti_consumer/tests/unit/test_lti.py
```diff
@@ -0,0 +1,287 @@
...
+        'lti_consumer.lti_consumer.LtiConsumerXBlock.prefixed_custom_parameters',
+        PropertyMock(return_value={u'custom_param_1': 'custom1', u'custom_param_2': 'custom2'})
+    )
+    @patch('lti_consumer.lti_consumer.LtiConsumerXBlock.lti_provider_key_secret', PropertyMock(return_value=('t', 's')))
+    @patch('lti_consumer.lti_consumer.LtiConsumerXBlock.user_id', PropertyMock(return_value='4knx2j'))
+    def test_get_signed_lti_parameters(self):
+        """
+        Test `get_signed_lti_parameters` returns the correct dict
+        """
+        expected_lti_parameters = {
```
benpatterson: I'm confused  why some of the mocked values are being asserted.

lti_consumer/tests/unit/test_lti.py
```diff
@@ -0,0 +1,287 @@
...
+            'oauth_timestamp="1378916897", oauth_version="1.0", oauth_signature_method="HMAC-SHA1", '
+            'oauth_consumer_key="t", oauth_signature="frVp4JuvT1mVXlxktiAUjQ7%2F1cw%3D"'
+        )
+    ))
+    @patch(
+        'lti_consumer.lti_consumer.LtiConsumerXBlock.prefixed_custom_parameters',
+        PropertyMock(return_value={u'custom_param_1': 'custom1', u'custom_param_2': 'custom2'})
+    )
+    @patch('lti_consumer.lti_consumer.LtiConsumerXBlock.lti_provider_key_secret', PropertyMock(return_value=('t', 's')))
+    @patch('lti_consumer.lti_consumer.LtiConsumerXBlock.user_id', PropertyMock(return_value='4knx2j'))
```
benpatterson: Should this be in a factory that creates a user for you?

lti_consumer/tests/unit/test_lti.py
```diff
@@ -0,0 +1,287 @@
...
+from mock import Mock, PropertyMock, patch
+
+from lti_consumer.tests.unit.test_utils import make_request
+from lti_consumer.tests.unit.test_lti_consumer import TestLtiConsumerXBlock
+
+from lti_consumer.lti import parse_result_json, Lti1Consumer, Lti2Consumer
+from lti_consumer.exceptions import LtiError
+
+
+INVALID_JSON_INPUTS = [
```
benpatterson: I think you should refactor this into dicts, with your comments as strings. That way, if more are introduced, they will be listed that way. Just makes it clearer why they should be failing. It can stand in for test case name, basically, since these are being organized into an array rather than test methods.

lti_consumer/tests/unit/test_lti.py
```diff
@@ -0,0 +1,287 @@
...
+        }
+        self.lti_consumer.xblock.has_score = True
+        self.lti_consumer.xblock.ask_to_send_username = True
+        self.lti_consumer.xblock.ask_to_send_email = True
+
+        self.lti_consumer.xblock.runtime.get_real_user.return_value = Mock(email='edx@example.com', username='edx')
+        self.assertEqual(self.lti_consumer.get_signed_lti_parameters(), expected_lti_parameters)
+
+        self.lti_consumer.xblock.runtime.get_real_user.return_value = {}
+        del expected_lti_parameters['lis_person_sourcedid']
```
benpatterson: Why remove?

lti_consumer/tests/unit/test_lti.py
benpatterson: Also same thing: this looks like a real value. You'll want to make it obviously-fake.

lti_consumer/tests/unit/test_lti_consumer.py
```diff
@@ -0,0 +1,665 @@
...
+
+class TestLtiConsumerXBlock(unittest.TestCase):
+    """
+    Unit tests for LtiConsumerXBlock.max_score()
+    """
+
+    def setUp(self):
+        super(TestLtiConsumerXBlock, self).setUp()
+        self.xblock_attributes = {
+            'launch_url': 'http://localhost:8005/lti_launch',
```
benpatterson: But, this isn't actually being used as a running process during tests, is it? I wouldn't expect a running xblock to be necessary for the unit tests...

lti_consumer/tests/unit/test_lti.py
```diff
@@ -0,0 +1,287 @@
...
+        Test `get_result` returns valid json response
+        """
+        self.xblock.module_score = 0.9
+        self.xblock.score_comment = 'Great Job!'
+        response = dict(GET_RESULT_RESPONSE)
+        response.update({
+            "resultScore": self.xblock.module_score,
+            "comment": self.xblock.score_comment
+        })
+        self.assertEqual(self.lti_consumer.get_result(Mock()), response)
```
benpatterson: This is confusing me, too. Sorry if I'm not connecting something mentally here... but I'm surprised one could send in a mock object and get a valid response value?

lti_consumer/tests/unit/test_lti.py
```diff
@@ -0,0 +1,287 @@
...
+
+        self.xblock.module_score = None
+        self.xblock.score_comment = ''
+        self.assertEqual(self.lti_consumer.get_result(Mock()), GET_RESULT_RESPONSE)
+
+    @patch('lti_consumer.lti_consumer.LtiConsumerXBlock.clear_user_module_score')
+    def test_delete_result(self, mock_clear):
```
benpatterson: Is there a reason you can't actually clear data as part of this test? Like perhaps mock the underlying data store? It's great we assert that the delete is called, but I'd want to confirm it actually deletes...

benpatterson:
@douglashall I'm only partway through and I can't spend more time on it today. I can get back to it tomorrow. However I wanted to ask about the use of mocks. There are a lot of assertions that a given method is called. That makes sense so long as the underlying method is actually being used in another test. (Or, you actually exercise the method instead of mocking it out---but perhaps you mock out the underlying structure, such as the database or file so the test is better cleaned between runs, faster b/c it's in memory, etc). So could you take a spin through your tests and revisit which methods you might be mocking and never actually using (i.e. covering)?

.gitignore
douglashall: Cool. Wasn't sure if it was standard practice to keep those in global gitignore.

.travis.yml
douglashall: Thanks. I will remove.

lti_consumer/lti.py
douglashall: Yes. Sorry, much of this was existing code copied over from edx-platform. I did not do a thorough doc cleanup.

lti_consumer/lti.py
douglashall: Yes. Thanks.

lti_consumer/lti.py
douglashall: Yes. Thank you.

.gitignore
smarnach: I tend to think files IDEs and editors create don't belong here, unless they are produced by tools virtually everyone working with this repo is using.  For other tools, it's up to the people using these tools to put adequate patterns in `.git/info/exclude` or `~/.config/git/ignore`.

lti_consumer/lti_consumer.py
douglashall: Yes.

lti_consumer/lti_consumer.py
douglashall: Yes, this is a hold-over from the xmodule world and is only used in rebind_noauth_module_to_user(). I would like to limit the changes to edx-platform for this refactoring at this time, but I will create a Jira task to the backlog for addressing this at a later time. Fair enough?

https://openedx.atlassian.net/browse/LTI-3

lti_consumer/lti_consumer.py
douglashall: LTI Tool Consumer, in this case, edx-platform running this XBlock. Updating the docs.

lti_consumer/lti_consumer.py
douglashall: `resource_link_id is a required LTI launch parameter.` Updating.

.gitignore
benpatterson: I understand the argument, but that might be relying too much on local set-up. Not everyone will have their global git config setup that way.

lti_consumer/lti_consumer.py
douglashall: The way LTI works is that the LTI consumer (in this case, this xblock) sends a POST request with signed oauth parameters to the LTI provider (the tool that will be embedded on the courseware page) in order to "launch" the user into the provider application. The form referenced here is rendered within an iframe and sends this "launch" POST request to the LTI provider, the provider responds with a HTML (usually) rendering of itself. Does that shed more light on this?

lti_consumer/tests/unit/test_lti.py
douglashall: In this case we are passing in a user object which `get_result` uses to re-attach that user to the current request (these requests are made using an oauth-signed header for authentication). The code that does the re-attaching lives in edx-platform, so all I really want to test here is that I get the expected response object with the appropriate score and comment.

douglashall:
@benpatterson Re: jshint, okay to make that incremental improvement in a subsequent PR? I would like to keep this focused on porting the code over from edx-platform. Also, the amount of JS for this XBlock is limited.

benpatterson:
@douglashall I'm on the same page as you. I didn't realize this relied on methods in edx-platform (I'm not as familiar with some of the areas of code you are interacting with). So we agree on the high-level concept; that's really what I was looking for.

lti_consumer/tests/unit/test_utils.py
douglashall: Actually, it looks like the XBlock object's __dict__ attribute is not updating it's attributes. I'm going to leave this as is for now.

lti_consumer/tests/unit/test_lti.py
douglashall: I'm not sure I need a factory here. I am just looking to mock out the `LtiConsumerXBlock.user_id` property. I will make user_id look fake.

lti_consumer/tests/unit/test_lti.py
douglashall: Here we are testing the case when a user cannot be found `self.lti_consumer.xblock.runtime.get_real_user.return_value = {}`. In this case, `lis_person_sourcedid` and `lis_person_contact_email_primary` are NOT expected to appear in the returned LTI parameters. I will add a comment indicating this case.

lti_consumer/tests/unit/test_lti.py
douglashall: I just want to make sure that `get_signed_lti_parameters` is returning the dictionary that I would expect.

lti_consumer/tests/unit/test_lti.py
douglashall: I will make these look less real.

lti_consumer/tests/unit/test_lti.py
douglashall: These tests and the test data above are largely copied over from edx-platform. I would like to refactor them in a follow up PR, if that works for you?

https://openedx.atlassian.net/browse/LTI-4

lti_consumer/lti_consumer.py
smarnach: Thanks, that's perfectly fine.  We just need to make sure it gets cleaned up, to avoid things like this accumulating.

lti_consumer/lti_consumer.py
smarnach: @douglashall I know roughly how it works.  What I didn't understand is why we need a fake form to send a POST request; `$.post()` seems more straight-forward.  I think the answer is that we want to redirect the browser to the result of the POST request, and the only way to achieve this is to use a fake form.

lti_consumer/static/css/student.css
smarnach: @douglashall I don't think edx-platform helps you in any way with compiling the CSS, but you can always add the compilation to `setup.py` to make sure it gets compiled on installation.

lti_consumer/tests/unit/test_utils.py
smarnach: Ah, right, we XBlock fields have setters that needs to be explicitly called.  I should have thought of this.

smarnach:
@douglashall Thanks for the updates.  Just to help you keep track, the following of my comments are still open: https://github.com/edx/xblock-lti-consumer/pull/1#issuecomment-157062618, https://github.com/edx/xblock-lti-consumer/pull/1#discussion_r44939374.  (I wish Github had a feature to mark comments as done.)

douglashall:
Thanks @smarnach. Regarding https://github.com/edx/xblock-lti-consumer/pull/1#issuecomment-157062618:

* Yes, it would be nice to have the old version of the custom parameters widget in the studio edit view. I am using the studio_view offered by `StudioEditableXBlockMixin` for this XBlock which does not support the slick List widget that you get for XModules in Studio. I could implement such a feature in xblock-utils. Ideally, Studio would be able to generate the edit form from XBlock Fields the way it does for XModules (there is an open Jira task outlining this work) and we wouldn't need to use `StudioEditableXBlockMixin` at all. In any case, I think neither option is in scope for this work.

* I have fixed the Modal Height/Width field descriptions to indicate that the unit is pixels. I have also added another field to input iframe height when the Inline launch target option is selected. I don't think we want to enable the width of the inline iframe to be changed because then we would start overlapping with LMS chrome.

* I am seeing lis_outcome_service_url come over as a relative path with the XModule version of this component as well when rendered in Studio. Unclear why this might be, but I think changing that is out of scope for this PR.

lti_consumer/static/css/student.css
douglashall: Wouldn't that require us to make an assumption that sass is installed in the environment which is consuming this package? I'm not sure I would want to do that.

smarnach: @douglashall Good point.  I'm not sure whether this would be a good idea either, so let's just keep this file here for now.

One thing we could do to prevent this file from getting out of sync with the SCSS source is to add a test to the Travis build that checks whether the file is still current.  Something like this might do:
```yaml
addons:
  apt:
    packages:
      - ruby-sass
script:
  - sass lti_consumer/static/sass/student.scss | cmp - lti_consumer/static/css/student.css
```
Apart from this being completely untested, it might also unreasonably extent build times or be unpractical for another reason.  If you decide trying something like this, I'd be interested in how it goes. :)

smarnach:
@douglashall Thanks for the comments.  I agree that adding list support to `StudioEditableXBloxkMixin` and fixing the resource URLs in Studio are out of scope for this PR.  Could you please create a Jira ticket for fixing the resorce URLs in Studio?  It would be nice if this eventually gets fixed.

Regarding the Modal Height – the field that existed could already be used to set the height of an iframe; it just wasn't documented that it worked that way.  I'm not sure we need separate fields for the height of the modal and the iframe, since only one of the two is useful in any configuration, but having the separate fields makes clear what can be used in what context, so I'm fine with it.

douglashall:
@benpatterson I have added a makefile. Thanks.

douglashall:
@smarnach Re: Modal Height vs. Inline Height fields, I think ideally we would customize the studio edit view so that fields that depend on other fields would only become visible when needed (out of scope for this PR). To me it was clearer from a UX perspective to keep the two heights separate.

smarnach:
@douglashall Agreed, in particular regarding the future option to hide unneeded fields.

benpatterson:
Hmm the new github GUI didn't like me commenting on a particular commit. But basically you want to have `set -e` at the top of each of your .sh files.

douglashall:
@smarnach I created the following Jira to investigate the lis_outcome_service_url discrepancy between LMS and Studio.

https://openedx.atlassian.net/browse/LTI-5

benpatterson:
@douglashall not sure if you need an official thumbs from me, but LGTM.

Thanks for going the Makefile route. I think that'll be easier for CI and for developers.

douglashall:
Thanks @benpatterson. I appreciate the help.

README.rst
```diff
@@ -0,0 +1,68 @@
...
+
+    $ make quality
+
+License
+-------
+
+The LTI Consumer XBlock is available under the Apache Version 2.0 License.
+
+
+.. |Build Status| image:: https://travis-ci.org/edx/xblock-lti-consumer.svg
```
benpatterson: I came back to this PR to doublecheck badges are here. Done-already. Nice.

douglashall:
@ormsbee @cpennington Could you guys take a look at this PR when you get a chance?

frrrances:
@douglashall Thanks for getting the sandbox running again. I'll do a UX review on Monday.

douglashall:
@frrrances Thanks Frances! I appreciate the time.

douglashall:
@macdiesel Can you take a look at this PR when you have a chance?

sarina:
This looks reasonable to me @douglashall at a glance, but you should definitely get a qualified review from a platform team engineer.

LICENSE
```diff
@@ -0,0 +1,202 @@
+
+                                 Apache License
```
sarina: What's the reasoning for this block being apache vs agpl?

LICENSE
```diff
@@ -0,0 +1,202 @@
...
+      To apply the Apache License to your work, attach the following
+      boilerplate notice, with the fields enclosed by brackets "[]"
+      replaced with your own identifying information. (Don't include
+      the brackets!)  The text should be enclosed in the appropriate
+      comment syntax for the file format. We also recommend that a
+      file or class name and description of purpose be included on the
+      same "printed page" as the copyright notice for easier
+      identification within third-party archives.
+
+   Copyright [yyyy] [name of copyright owner]
```
sarina: If you move forward w/ this license you need to do this.

lti_consumer/lti.py
```diff
@@ -0,0 +1,281 @@
...
+        content_type = request.headers.get('Content-Type')
+        if verify_content_type and content_type != Lti2Consumer.CONTENT_TYPE_RESULT_JSON:
+            log.info("[LTI]: v2.0 result service -- bad Content-Type: %s", content_type)
+            raise LtiError(
+                "For LTI 2.0 result service, Content-Type must be %s.  Got %s",
+                Lti2Consumer.CONTENT_TYPE_RESULT_JSON,
+                content_type
+            )
+
+        _, secret = self.xblock.lti_provider_key_secret
```
sarina: General nit: I prefer we use `__` to mean 'unused variable' because the `_` semantic gets overloaded w/ the underscore gettext function. It's less of a concern in this xblock but it's a habit I like to follow everywhere.

LICENSE
douglashall: Joel had commented previously on this PR that we should be using Apache.

lti_consumer/lti.py
douglashall: Makes sense. I will update.

LICENSE
douglashall: Does this need to go into every file?

LICENSE
sarina: OK, I find that interesting - why @jbarciauskas ? XBlock API is Apache to encourage xblocks of all license types, but I don't quite understand why an xblock we author shouldn't be agpl

jbarciauskas: There are ways in which we think the AGPL was the wrong decision for the platform. I don't want to be the final word on this if this is a significant policy change, I thought we were moving towards Apache in general because the AGPL is overly limiting.

sarina: OK, was just trying to understand the reasoning. I guess I'm not feeling
one way or another, but I think AGPL was chosen in part on feedback from
our community (because can't you choose to close source a fork of Apache
licensed code, and continue forward development there?)

frrrances:
@douglashall I completed my UX review and the latest sandbox looks good! The Studio side seems to have almost all the settings I would expect and have appropriate strings to describe what each field does. The LMS side looks good and fits in with the surrounding courseware. I think there's just one thing that should be addressed before merge: the modal should have an explicit close button that can be navigated to with a keyboard. 

Once that's addressed, I think this is good from an experience perspective.

douglashall:
@frrrances Thanks for taking a look. Can you point me to an example of what the close button should look like in LMS?

LICENSE
antoviaque: @sarina @jbarciauskas +1 to Sarina's point - this requirement from the AGPL is very central to getting most of our clients to accept contributing the changes they pay for. From past experience on projects without that requirement, if we move away from the AGPL, it will become much harder to get clients to agree, and will likely significantly diminish the amount of contributions we are able to do.

jbarciauskas: If that's the case, then I'm fine making this AGPL. I've mostly heard the opposite, though - most organizations want the option to close source their changes.

sarina: I would suspect many orgs do want that option the question is do we want to
encourage that? Xavier's point is that if orgs CAN close source their
changes, they will, which would mean less contributing back to the
community. He would like to see the opposite happen.

sarina: FWIW my opinion is that our code should be AGPL; our tooling, however, may
make sense to be Apache so that more people can make use of the tools (eg
xblock sdk or i18n-tools).

jbarciauskas: OK. I'm not advocating for a significant policy change here, so if AGPL matches our past work then let's go with that

douglashall: Thanks for working that out folks. Switching back to AGPL.

lti_consumer/lti.py
```diff
@@ -0,0 +1,281 @@
...
+
+    The json_str must be loadable.  It can either be an dict (object) or an array whose first element is an dict,
+    in which case that first dict is considered.
+    The dict must have the "@type" key with value equal to "Result",
+    "resultScore" key with value equal to a number [0, 1],
+    The "@context" key must be present, but we don't do anything with it.  And the "comment" key may be
+    present, in which case it must be a string.
+
+    Arguments:
+        json_str (unicode):  The body of the LTI 2.0 results service request, which is a JSON string]
```
doctoryes: Errant `]`

douglashall: Removed. Thanks.

lti_consumer/lti.py
```diff
@@ -0,0 +1,281 @@
...
+
+    Raises:
+        LtiError: if verification fails
+    """
+    try:
+        json_obj = json.loads(json_str)
+    except (ValueError, TypeError):
+        msg = "Supplied JSON string in request body could not be decoded: {}".format(json_str)
+        log.info("[LTI] %s", msg)
+        raise LtiError(msg)
```
doctoryes: Any concerns about what might wind up in this log message? This JSON will be submitted from a third-party LTI app, correct? Are there security concerns here?

lti_consumer/lti.py
```diff
@@ -0,0 +1,281 @@
...
+
+    # the standard supports a list of objects, who knows why. It must contain at least 1 element, and the
+    # first element must be a dict
```
doctoryes: `The JSON object must be a dict. If a non-empty list is passed in, use the first element - but only if it's a dict.`

doctoryes: Or something like that.

lti_consumer/lti.py
```diff
@@ -0,0 +1,281 @@
...
+    # '@type' must be "Result"
+    result_type = json_obj.get("@type")
+    if result_type != "Result":
+        msg = "JSON object does not contain correct @type attribute (should be 'Result', is z{})".format(result_type)
+        log.info("[LTI] %s", msg)
+        raise LtiError(msg)
+
+    # '@context' must be present as a key
+    REQUIRED_KEYS = ["@context"]  # pylint: disable=invalid-name
+    for key in REQUIRED_KEYS:
```
doctoryes: Nit: No need for `REQUIRED_KEYS` here.

doctoryes: Also, if there were multiple `REQUIRED_KEYS`, this would only complain about the first one found.

lti_consumer/lti.py
```diff
@@ -0,0 +1,281 @@
...
+    # 'resultScore' is not present.  If this was a PUT this means it's actually a DELETE according
+    # to the LTI spec.  We will indicate this by returning None as score, "" as comment.
+    # The actual delete will be handled by the caller
+    if "resultScore" not in json_obj:
+        return None, json_obj.get('comment', "")
+
+    # if present, 'resultScore' must be a number between 0 and 1 inclusive
+    try:
+        score = float(json_obj.get('resultScore', "unconvertable"))  # Check if float is present and the right type
+        if not 0 <= score <= 1:
```
doctoryes: Nit: This code will work - but I'd prefer `0.0 <= score <= 1.0` to keep all type conversions out of the picture.

lti_consumer/lti.py
```diff
@@ -0,0 +1,281 @@
...
+        msg = "Supplied JSON string in request body could not be decoded: {}".format(json_str)
+        log.info("[LTI] %s", msg)
+        raise LtiError(msg)
+
+    # the standard supports a list of objects, who knows why. It must contain at least 1 element, and the
+    # first element must be a dict
+    if not isinstance(json_obj, dict):
+        if isinstance(json_obj, list) and len(json_obj) >= 1 and isinstance(json_obj[0], dict):
+            json_obj = json_obj[0]
+        else:
```
macdiesel: this was a speed bump while I was reading through the code. Perhaps something like this:

 if isinstance(json_obj, list) and len(json_obj) >= 1
    json_obj = json_obj[0]

if not isinstance(json_obj, dict):
    #error stuff here
    
A little more clear, covers the same cases and only checks if the object is a dict once.

lti_consumer/lti.py
```diff
@@ -0,0 +1,281 @@
...
+
+    # '@context' must be present as a key
+    REQUIRED_KEYS = ["@context"]  # pylint: disable=invalid-name
+    for key in REQUIRED_KEYS:
+        if key not in json_obj:
+            msg = "JSON object does not contain required key {}".format(key)
+            log.info("[LTI] %s", msg)
+            raise LtiError(msg)
+
+    # 'resultScore' is not present.  If this was a PUT this means it's actually a DELETE according
```
macdiesel: can we flush out the first sentence "'resultScore' is not present" a bit more, while technically a sentence fragment, it could use more context.

lti_consumer/lti.py
```diff
@@ -0,0 +1,281 @@
...
+    # to the LTI spec.  We will indicate this by returning None as score, "" as comment.
+    # The actual delete will be handled by the caller
+    if "resultScore" not in json_obj:
+        return None, json_obj.get('comment', "")
```
macdiesel: I'm confused, if result score is not present and this is a PUT request it's actually a delete, but this code does not check for the the HTTP Verb. Question, if this method is parsing the JSON, should this piece of the logic be in this method at all?

lti_consumer/lti.py
```diff
@@ -0,0 +1,281 @@
...
+
+    # if present, 'resultScore' must be a number between 0 and 1 inclusive
+    try:
+        score = float(json_obj.get('resultScore', "unconvertable"))  # Check if float is present and the right type
+        if not 0 <= score <= 1:
+            msg = 'score value outside the permitted range of 0-1.'
```
macdiesel: The above line should also be updated to reflect @doctoryes' comment above about floats.

lti_consumer/lti.py
```diff
@@ -0,0 +1,281 @@
...
+            raise LtiError(msg)
+    except (TypeError, ValueError) as err:
+        msg = "Could not convert resultScore to float: {}".format(err.message)
+        log.info("[LTI] %s", msg)
+        raise LtiError(msg)
+
+    return score, json_obj.get('comment', "")
+
+
+class Lti1Consumer(object):
```
macdiesel: Documentation below says: 
We have also begun to add support for LTI 1.2/2.0

Name of this class is Lti1Consumer. Is there a better name we could use for this so it's obvious that this is the 1.1 specification?

lti_consumer/lti_consumer.py
```diff
@@ -0,0 +1,839 @@
...
+from xblock.fields import Boolean, Float, Integer
+from xblock.fragment import Fragment
+
+from xblockutils.resources import ResourceLoader
+from xblockutils.studio_editable import StudioEditableXBlockMixin
+
+from .exceptions import LtiError
+from .oauth import log_authorization_header
+from .lti import Lti1Consumer, Lti2Consumer
+from .outcomes import OutcomeService
```
macdiesel: @nedbat remind me, we decided that full path imports are better than relative during the 1.8 upgrade?

lti_consumer/lti_consumer.py
```diff
@@ -0,0 +1,839 @@
...
+
+        First part of location is tag-org-course-category, i4x-2-3-lti.
+
+        Location.name itself does not change on import to another course,
+        but org and course_id change.
+
+        So together with org and course_id in a form of
+        i4x-2-3-lti-31de800015cf4afb973356dbe81496df this part of resource_link_id:
+        makes resource_link_id to be unique among courses inside same system.
+        """
```
doctoryes: The docstring above refers to an old-style, deprecated block locator:
https://github.com/edx/opaque-keys/blob/master/opaque_keys/edx/locator.py#L625
Will the scheme described above be affected by a new-style `block-v1` tag? Will LTI use this new-style tag at all?

lti_consumer/lti_consumer.py
nedbat: full path imports are safer, in that they don't depend on this file's importer doing the right thing.  If everyone either uses the same absolute import, or uses a relative import, we're OK.  The bad thing is when there are two different absolute imports of the same module.

lti_consumer/lti.py
douglashall: The JSON submitted by the third-party app is expected to provide a standard format which includes a grade. No user identity information is expected in the JSON body. I think the risk of potentially logging some random user identity information that the third-party app might send us is worth the reward of being able to potentially help a partner debug their LTI provider.

lti_consumer/lti.py
douglashall: PUT is the only request method that would require us to parse the JSON body of the request, so this function is only called from Lti2Consumer.put_result` below. I'll try to rewrite the comment and code to make this a little more obvious.

lti_consumer/lti.py
douglashall: How about a single class called LtiConsumer and I will update the documentation to indicate that this is a limited implementation of the LTI 1.1 and 2.0 specs?

lti_consumer/lti_consumer.py
douglashall: I can't seem to get an absolute import working here when I run the unit tests with nosetests. If I change these to `from lti_consumer.exceptions import LtiError`, I get `ImportError: No module named exceptions` when running `make test`. @nedbat Can you offer any python import wisdom here?

smarnach: I'd argue that explicit relative imports are safer, since they can't ever introduce a second way of importing the module; given that this will be installed as a separate package, there will be only a single way of importing the modules anyway, so there is no risk of anything going wrong regardless of what types of import is used here. :)

That said, I can't see a reason why `from lti_consumer.exceptions import LtiError` shouldn't work.  @douglashall could you post the full traceback of the ImportError?

lti_consumer/lti.py
macdiesel: ok, this makes more sense to me now, thanks Doug.

lti_consumer/lti_consumer.py
```diff
@@ -0,0 +1,839 @@
...
+    def clear_user_module_score(self, user):
+        """
+        Clears the module user state, including grades and comments, and also scoring in db's courseware_studentmodule
+
+        Arguments:
+            user (django.contrib.auth.models.User):  Actual user whose module state is to be cleared
+
+        Returns:
+            nothing
+        """
```
doctoryes: Thanks for all these docstrings! You might consider using PyContracts in the future - just search for `@contract` in the current edx-platform codebase for examples.

lti_consumer/oauth.py
```diff
@@ -0,0 +1,153 @@
...
+        params=oauth_headers.items(),
+        signature=oauth_signature
+    )
+    mock_request_lti_2 = Mock(
+        uri=unicode(urllib.unquote(request.url)),
+        http_method=unicode(request.method),
+        params=oauth_headers.items(),
+        signature=oauth_signature
+    )
+    if oauth_body_hash != oauth_headers.get('oauth_body_hash'):
```
doctoryes: I know this code is copied over from lti_module.py. But is there an easy way to accomplish the same without using `Mock` here? Coders generally don't expect to see it in non-test code. A nit.

lti_consumer/oauth.py
```diff
@@ -0,0 +1,153 @@
...
+    params = client.get_oauth_params(request)
+    params.append((u'oauth_body_hash', oauth_body_hash))
+    mock_request = Mock(
+        uri=unicode(urllib.unquote(request.url)),
+        headers=request.headers,
+        body=u"",
+        decoded_body=u"",
+        oauth_params=params,
+        http_method=unicode(request.method),
+    )
```
doctoryes: Ditto nit.

lti_consumer/outcomes.py
```diff
@@ -0,0 +1,205 @@
...
+    Returns:
+        tuple: imsx_messageIdentifier, sourcedId, score, action
+
+    Raises:
+        LtiError
+            if submitted score is outside the permitted range
+            if the XML is missing required entities
+            if there was a problem parsing the XML body
+    """
+    lti_spec_namespace = "http://www.imsglobal.org/services/ltiv1p1/xsd/imsoms_v1p0"
```
doctoryes: Is it appropriate to hard-code this URI?

lti_consumer/outcomes.py
```diff
@@ -0,0 +1,205 @@
...
+
+    try:
+        score = root.xpath("//def:textString", namespaces=namespaces)[0].text
+    except IndexError:
+        raise LtiError('Failed to parse score textString from XML request body')
+
+    # Raise exception if score is not float or not in range 0.0-1.0 regarding spec.
+    score = float(score)
+    if not 0 <= score <= 1:
+        raise LtiError('score value outside the permitted range of 0-1.')
```
doctoryes: I'd again prefer to see `0.0` and `1.0` above.

README.rst
```diff
@@ -0,0 +1,68 @@
...
+
+Running code quality check
+--------------------------
+
+From the xblock-lti-consumer repo root, run the quality checks with the following command:
+
+.. code:: bash
+
+    $ make quality
+
```
doctoryes: IMO, the `compile_sass` instructions should be mentioned in here, along with when it's appropriate to run it (i.e. when the SCSS files change).

lti_consumer/tests/unit/test_lti_consumer.py
```diff
@@ -0,0 +1,664 @@
...
+
+        self.xblock.runtime.get_user_role.return_value = 'instructor'
+        self.assertEqual(self.xblock.role, 'Instructor')
+
+    def test_course(self):
+        """
+        Test `course` calls modulestore.get_course
+        """
+        mock_get_course = self.xblock.runtime.descriptor_runtime.modulestore.get_course
+        mock_get_course.return_value = None
```
doctoryes: How does this `get_course` method get mocked? Isn't `mock_get_course` just a regular method reference? I'm probably missing something.

smarnach: The runtime is a mock, since the unit tests are run without an actual XBlock runtime.  See https://github.com/edx/xblock-lti-consumer/pull/1/files#diff-b8f4cf5104f4cca5494e761e86fa3a0cR30.

lti_consumer/tests/unit/test_lti_consumer.py
```diff
@@ -0,0 +1,664 @@
...
+
+    @patch('lti_consumer.LtiConsumerXBlock.set_user_module_score')
+    def test_clear_user_module_score(self, mock_set_user_module_score):
+        """
+        Test that clear_user_module_score calls set_user_module_score with params set to None
+        """
+        user = Mock()
+        self.xblock.clear_user_module_score(user)
+        mock_set_user_module_score.assert_called_with(user, None, None)
+
```
doctoryes: Are there any tests anywhere that use a `max_score` that's greater than 1.0? 0.0? Less than 0.0? A `score` that's greater than `max_score`?

requirements.txt
```diff
@@ -0,0 +1,7 @@
+lxml
+bleach==1.4.2
+oauthlib==1.0.3
+Mako
+git+https://github.com/edx/XBlock.git#egg=XBlock
```
doctoryes: Is it the XBlock standard to require the XBlock depot using this unpinned form? Do other XBlocks do the same?

doctoryes:
My first pass is complete.

lti_consumer/tests/unit/test_lti_consumer.py
douglashall: Thanks @smarnach!

lti_consumer/tests/unit/test_lti_consumer.py
douglashall: @doctoryes Yes, there are. Check out the tests in [test_lti.py](https://github.com/edx/xblock-lti-consumer/pull/1/files#diff-203113196545de710258838493172076R86) and [test_outcomes.py](https://github.com/edx/xblock-lti-consumer/pull/1/files#diff-a0dbcbf039b96fab3e0e5f822c260f65R210).

requirements.txt
douglashall: There doesn't seem to be a standard convention that I can see. I think it would be great if we started creating git tags for releases of the XBlock package and pinning XBlock implementation repos to those tags. I'm also happy to defer to any other standard either existing or new.

douglashall:
@macdiesel @doctoryes Thanks for the reviews. I believe I have addressed all of your comments.

lti_consumer/outcomes.py
douglashall: This is the namespace for the XML document that the LTI outcome service endpoint expects to receive with requests to update component score. I don't see an issue with hard-coding this here.

requirements.txt
doctoryes: @nedbat Does a standard / best practice exist for pinning XBlocks to a particular version of the XBlock module?

lti_consumer/outcomes.py
doctoryes: Now that I know what it is, I agree.

lti_consumer/lti.py
```diff
@@ -0,0 +1,274 @@
...
+    except (ValueError, TypeError):
+        msg = "Supplied JSON string in request body could not be decoded: {}".format(json_str)
+        log.error("[LTI] %s", msg)
+        raise LtiError(msg)
+
+    # The JSON object must be a dict. If a non-empty list is passed in,
+    # use the first element, but only if it is a dict
+    if isinstance(json_obj, list) and len(json_obj) >= 1:
+        json_obj = json_obj[0]
+
```
macdiesel: Much clearer thank you!

doctoryes:
All remaining comments from me are nits. :+1:

lti_consumer/outcomes.py
```diff
@@ -0,0 +1,205 @@
...
+from xblockutils.resources import ResourceLoader
+
+from .exceptions import LtiError
+from .oauth import verify_oauth_body_signature
+
+
+log = logging.getLogger(__name__)
+
+
+def parse_grade_xml_body(body):
```
macdiesel: Since we are not using streaming parsing here do we want to put any limits on the max size of the XML that can be loaded to avoid buffer overflow types of attacks?

requirements.txt
macdiesel: @doctoryes @nedbat how do we version XBlock right now? Do we have anything in place where we would guarantee functionality between versions?

requirements.txt
```diff
@@ -0,0 +1,7 @@
+lxml
+bleach==1.4.2
+oauthlib==1.0.3
+Mako
```
macdiesel: @doctoryes does the version of Mako in here need to match the version we are using in edx-platform? (which btw is mako==1.0.2)

Should this be mako>=1.0,<2.0 ? perhaps  < 1.2?

lti_consumer/outcomes.py
douglashall: Would it be safe enough to assume that such attacks would be handled by the web-tier (`client_max_body_size` on nginx)? I will also add a Jira issue for changing this to parse the XML body iteratively as a stream.

https://openedx.atlassian.net/browse/LTI-7

test_requirements.txt
```diff
@@ -0,0 +1,7 @@
+-r requirements.txt
+
+coveralls
+pep8
+git+https://github.com/pmitros/django-pyfs.git@514607d78535fd80bfd23184cd292ee5799b500d#egg=djpyfs
```
macdiesel: I believe that we should be using the edx branch of this now as this is in edx-platform:
https://github.com/edx/edx-platform/blob/master/requirements/edx/github.txt#L58

@doctoryes can you confirm this?

setup.py
```diff
@@ -0,0 +1,47 @@
...
+        'lti_consumer',
+    ],
+    install_requires=[
+        'lxml',
+        'bleach',
+        'oauthlib',
+        'Mako',
+        'XBlock',
+        'xblock-utils',
+    ],
```
macdiesel: since the requirements.txt specifies versions and version ranges for these libraries, they should also be spelled out too in the setup.py.

lti_consumer/outcomes.py
macdiesel: if we are filtering on the max_body_size in nginx this is good to go.

macdiesel:
Finished my second pass on this. few nits mostly on the requiements and setup.py files, otherwise :+1: 
:shipit:

test_requirements.txt
doctoryes: Yes - it should look like the line referred to above - tag `1.0.3`.

LICENSE
nedbat: Our current plan (always up for re-assessment if there is reason) is to use AGPL for XBlocks: https://open.edx.org/open-edx-licensing

requirements.txt
douglashall: Yes. I will pin this to mako==1.0.2 as we are doing in edx-platform.

requirements.txt
nedbat: We have a tag in the XBlock repo now.  But pinning versions in a library makes things difficult as we have to change all of them in sync.

test_requirements.txt
douglashall: Thanks guys. I will make this update.

setup.py
douglashall: Yes. Will do. Thanks.

requirements.txt
doctoryes: AFAIK, there's no official policy around guaranteeing functionality between versions. Certainly, if we needed to change something that might have backwards-compatibility issues (to fix a bug or enhance functionality), we wouldn't want it to affect current installations. If we don't use tagged GitHub versions, it means we can never merge any of those types of changes back to the GitHub default branch (master). We should move to a policy of pinning to a GitHub tag or hash.

PR #2, opened 2015-12-01
Do not pin dependency versions

douglashall:
This prevents us from running into conflicts with the installed requirements in edx-platform. I think we only want to call out a version when we know we need a specific version. Specifically, oauthlib==1.0.3 was conflicting with the version (oauthlib==0.7.2) that edx-platform installs.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -68,6 +68,14 @@ and committed to the git repo using:
 Changes to style rules should be made to the Sass files, compiled to CSS,
 and committed to the git repository.
 
+Package Requirements
+--------------------
+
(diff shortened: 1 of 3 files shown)


douglashall:
@macdiesel @doctoryes @nedbat

douglashall:
@benpatterson

benpatterson:
:+1: from me

nedbat:
We need to clarify the use of requirements.txt vs setup.py, but I'm :+1: on merging this to move forward.

douglashall:
@nedbat I will add a section to the README of the repo explaining when requirements.txt is used and when setup.py is used in relation to requirements.

PR #3, opened 2015-12-17
PHX-207 Accessibility Improvements

douglashall:
* Toggle aria-hidden attribute on modal window open/close
* Set title attribute on iframe to display name of component

Sandbox URL:
https://lti-consumer.sandbox.edx.org/courses/course-v1:edX+Test101+course/courseware/039df3d8458844e68635ebe68681d9d3/a88b0dc3dbe44c4c820187763e12866b/


douglashall:
@cptvitamin Can you check out the sandbox and give this a review? Thanks.

cptvitamin:
@douglashall the behavior is now as expected.  Thanks for taking care of that.  There are two technical failures that we need to clear up, but they may be beyond the scope of this PR.  

1. The modal div has an aria-labelledby attribute that is referencing a non-existent id: lti-modal-title.  This is a technical failure.  We either need to add an element with this IDref or remove the attribute.

aria-labelledby referencing a non-existent ID

2. The IDref lean_overlay appears twice in the DOM.  This is a potential accessibility issue, especially if the IDref is being used to control focus management (which element is being referenced if there are two identical ones.)

two identical idrefs

douglashall:
Thanks @cptvitamin. I have removed the aria-labelledby attribute and made the modal overlay id attribute unique within the page.

cptvitamin:
:+1:  thanks @douglashall

PR #4, opened 2016-01-19
TNL-4046 Fix due date comparison bug

douglashall:
Use django.utils.timezone.now() when determining if the due date of a component has passed to avoid tz-aware/naive datetime comparison stacktrace.

Related PR:
https://github.com/edx/edx-platform/pull/11270

### DIFF
--- a/lti_consumer/tests/unit/test_lti_consumer.py
+++ b/lti_consumer/tests/unit/test_lti_consumer.py
@@ -4,8 +4,10 @@ Unit tests for LtiConsumerXBlock
 
 import unittest
 
-from datetime import datetime, timedelta
+from datetime import timedelta
 from mock import Mock, PropertyMock, patch
+
+from django.utils import timezone
 
 from lti_consumer.tests.unit.test_utils import FAKE_USER_ID, make_xblock, make_request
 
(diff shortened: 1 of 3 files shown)


douglashall:
@mushtaqak @adampalay

adampalay:
@douglashall , can you please add a test that would fail without the change?

douglashall:
@adampalay Sure thing. All set.

lti_consumer/tests/unit/test_lti_consumer.py
```diff
@@ -228,6 +230,19 @@ def test_is_past_due_no_graceperiod(self):
         self.xblock.due = now + timedelta(days=1)
         self.assertFalse(self.xblock.is_past_due)
 
+    @patch('lti_consumer.lti_consumer.timezone.now')
```
mushtaqak: Shouldn't we use `django.utils.timezone.now` in patch statement as described in test method description?

mushtaqak:
@douglashall Otherwise looks good to me :)

adampalay:
:+1:

lti_consumer/tests/unit/test_lti_consumer.py
douglashall: `@patch('lti_consumer.lti_consumer.timezone.now')` has the same effect, but just for the lti_consumer.lti_consumer module.

PR #6, opened 2016-03-18
asadiqbal08/WL-356: Add I18N support

asadiqbal08:
@mattdrayer your initial thoughts here ? 

###### we can ignore test failure at this point. having dependancy on PR# [335](https://github.com/edx/XBlock/pull/335)

### DIFF
--- a/lti_consumer/lti_consumer.py
+++ b/lti_consumer/lti_consumer.py
@@ -73,6 +73,7 @@ from .exceptions import LtiError
 from .oauth import log_authorization_header
 from .lti import LtiConsumer
 from .outcomes import OutcomeService
+from .utils import _
 
 log = logging.getLogger(__name__)
 
@@ -142,7 +143,7 @@ def parse_handler_suffix(suffix):
         if match_obj:
             return match_obj.group('anon_id')
     # fall-through handles all error cases
-    msg = "No valid user id found in endpoint URL"
+    msg = _("No valid user id found in endpoint URL")
     log.info("[LTI]: %s", msg)
     raise LtiError(msg)
 
@@ -243,17 +244,17 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
     """
 
     display_name = String(
-        display_name="Display Name",
-        help=(
-            "Enter the name that students see for this component.  "
+        display_name=_("Display Name"),
+        help=_(
+            "Enter the name that students see for this component. "
             "Analytics reports may also use the display name to identify this component."
         ),
         scope=Scope.settings,
-        default="LTI Consumer",
+        default=_("LTI Consumer"),
     )
     description = String(
-        display_name="LTI Application Information",
-        help=(
+        display_name=_("LTI Application Information"),
+        help=_(
             "Enter a description of the third party application. "
             "If requesting username and/or email, use this text box to inform users "
             "why their username and/or email will be forwarded to a third party application."
@@ -262,9 +263,9 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
         scope=Scope.settings
     )
     lti_id = String(
-        display_name="LTI ID",
-        help=(
-            "Enter the LTI ID for the external LTI provider.  "
+        display_name=_("LTI ID"),
+        help=_(
+            "Enter the LTI ID for the external LTI provider. "
             "This value must be the same LTI ID that you entered in the "
             "LTI Passports setting on the Advanced Settings page."
             "<br />See the {docs_anchor_open}edX LTI documentation{anchor_close} for more details on this setting."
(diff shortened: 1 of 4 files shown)


lti_consumer/utils.py
```diff
@@ -0,0 +1,11 @@
+# -*- coding: utf-8 -*-
+"""
+Make '_' a no-op so we can scrape strings
```
mattdrayer: Perhaps I am not understanding -- where is the tie-in to self.ugettext for returning the actual translated strings if the implementation references this no-op?

mattdrayer:
Okay, I am good w/ this PR after discussing w/ @asadiqbal08 offline -- :+1: 

@douglashall if you have some time today, would you mind taking a look at this update?  As mentioned above, the test failure is due to dependency on a pending change to the main XBlock project, which will hopefully be merging soon.  We will hold off on merging this PR until that one goes.

lti_consumer/utils.py
douglashall: Same question. Can you explain how this works?

lti_consumer/lti_consumer.py
```diff
@@ -72,6 +72,7 @@
 from .oauth import log_authorization_header
 from .lti import LtiConsumer
 from .outcomes import OutcomeService
+from .utils import _  # pylint: disable=unused-import
```
douglashall: Do you still need the pylint exception here?

lti_consumer/lti_consumer.py
```diff
@@ -243,32 +244,32 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
...
     display_name = String(
-        display_name="Display Name",
-        help=(
-            "Enter the name that students see for this component.  "
+        display_name=_("Display Name"),
+        help=(_(
+            "Enter the name that students see for this component. "
             "Analytics reports may also use the display name to identify this component."
-        ),
+        )),
```
douglashall: Do you need all these double parens?

lti_consumer/lti_consumer.py
```diff
@@ -717,7 +714,7 @@ def result_service_handler(self, request, suffix=''):
 
         user = self.runtime.get_real_user(anon_id)
         if not user:  # that means we can't save to database, as we do not have real user id.
-            msg = "[LTI]: Real user not found against anon_id: {}".format(anon_id)
+            msg = self.ugettext("[LTI]: Real user not found against anon_id: {}").format(anon_id)
```
douglashall: Looks like this message only goes to logs. Should we be marking this for translation?

lti_consumer/lti_consumer.py
```diff
@@ -822,17 +819,17 @@ def _get_context_for_template(self):
...
             'has_score': self.has_score,
             'weight': self.weight,
             'module_score': self.module_score,
-            'comment': sanitized_comment,
+            'comment': self.ugettext(sanitized_comment),
             'description': self.description,
             'ask_to_send_username': self.ask_to_send_username,
             'ask_to_send_email': self.ask_to_send_email,
-            'button_text': self.button_text,
+            'button_text': self.ugettext(self.button_text),
```
douglashall: Haven't you already marked these for translation up above in the field definitions?

lti_consumer/utils.py
asadiqbal08: @mattdrayer @douglashall let me explain this. 
1- The purpose of the `_(text)` method in utils.py is only to mark static strings for translations, like XBlock fields help-text where we cannot refer to `self`.
2- `_(text)` method is helpful for us to create strings for .po files (When we use [poedit](https://poedit.net/) ) then it will look for `_` and generate a .PO. 

Lastly, runtime is service is using the xBlock `self`and we are not able to get the reference of `self` for Xblock fields texts. 

Let me know If I am missing something.

lti_consumer/lti_consumer.py
asadiqbal08: If it is not required then we can eliminate.

lti_consumer/lti_consumer.py
asadiqbal08: @douglashall you are right but above, The purpose was generating a .po file. Now here runtime service is trying to translate the dynamic text if its translations are available in XBlock message catalogs.
@mattdrayer ^^

douglashall: Got it. Thanks for the clarification.

lti_consumer/lti_consumer.py
douglashall: Sounds good. Let's revert this one.

lti_consumer/utils.py
douglashall: Thanks for the clarification.

douglashall:
:+1:  Once tests are passing.

mattdrayer:
Great, thanks @douglashall!  @asadiqbal08 we'll also want to squash the commits before merging

mattdrayer:
Green build!  Merging now :)

bradenmacdonald:
@mattdrayer @douglashall @asadiqbal08 FYI, I have a PR at https://github.com/edx/xblock-utils/pull/38 which will make more of these translations appear in Studio. Currently, the field help text and names in the studio editor aren't actually working in most cases. You're welcome to review that PR if interested.

PR #7, opened 2016-03-21
PHX-255 Launch LTI provider on modal window open

douglashall:
@robrap @afzaledx

### DIFF
--- a/lti_consumer/templates/html/lti_iframe.html
+++ b/lti_consumer/templates/html/lti_iframe.html
@@ -1,8 +1,9 @@
+<%page args="initial_launch_url" expression_filter="h"/>
 <iframe
     title="${display_name}"
     class="ltiLaunchFrame"
     name="ltiFrame-${element_id}"
-    src="${form_url}"
+    src="${initial_launch_url}"
     allowfullscreen="true"
     webkitallowfullscreen="true"
     mozallowfullscreen="true"
--- a/lti_consumer/static/js/xblock_lti_consumer.js
+++ b/lti_consumer/static/js/xblock_lti_consumer.js
@@ -16,6 +16,9 @@ function LtiConsumerXBlock(runtime, element) {
                 return this.each(function () {
                     var o = options;
                     $(this).click(function (e) {
+                        var $modal = $(modal_id);
+                        // Set iframe src attribute to launch LTI provider
+                        $modal.find('iframe').attr('src', $modal.data('launch-url'));
                         $("#" + overlay_id).click(function () {
                             close_modal(modal_id)
                         });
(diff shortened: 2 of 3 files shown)


lti_consumer/templates/html/lti_iframe.html
```diff
@@ -1,8 +1,9 @@
+<%page args="launch_url"/>
```
robrap: Might as well add h filter too.

robrap:
@douglashall Let's review together. I know this is short and simple, but I'm not getting why the two ways of passing the data. I'm sure you can clear me up in no time.

douglashall:
@robrap Sounds good. Before this change, when the component was configured in "modal" launch mode, the LTI iframe was rendered and loaded on LMS page load. This essentially launched the LTI provider in a hidden iframe on the page. The modal open/close would simply show/hide the iframe. We received some user feedback that it would be better to re-launch the LTI provider whenever the modal window was opened enabling better "session" tracking on the LTI provider side. This change implements that enhancement by rendering the LTI iframe without a `src` attribute on page load. We then set/unset the src attribute of the iframe as the modal window is open/closed.

There is also the ability to configure the component to launch in "inline" mode, which means the LTI provider is displayed in an iframe inline on the LMS page. This functions as you would expect, the visible iframe is rendered and loaded on LMS page load.

The iframe HTML is shared between "modal" and "inline" mode by factoring it out into a separate template that is passed the LTI launch URL from the parent template. In the case of "modal" mode, the passed URL is an empty string because instead there is javascript that grabs the URL from a data attribute on the modal element when the modal window is opened.

Happy to also walk you through this in person tomorrow morning. I would like to get this into the RC.

Thanks for taking a look!

lti_consumer/templates/html/lti_iframe.html
```diff
@@ -1,8 +1,9 @@
+<%page args="launch_url" expression_filter="h"/>
```
robrap: @douglashall Thanks for the explanation.  It might have been just that I was tired, but my brain wasn't able to easily understand why this was needed when we had data-launch-url.

I'm not sure how much clarity this would add, but would you mind renaming this page arg to "default_launch_url" or "initial_launch_url"?  Does that make sense to you?

robrap:
:+1: @douglashall I'm not a UI expert, but it looks good to me.  I'll leave it to you to determine if you want to rename the arg, and if you want me to look at it again.  Thanks!

lti_consumer/templates/html/lti_iframe.html
douglashall: `initial_launch_url` sounds good. I have also added more detail about the modal launch to the comment in student.html.

robrap:
@douglashall Great.  Thanks.

PR #10, opened 2016-04-07
PHX-256 Added a white background to the modal window close button

douglashall:
The LTI Consumer component's modal window close button did not work well when the LTI provider that is launched within the modal window has a dark background.

cross mark is not visible

Related PR: https://github.com/edx/edx-platform/pull/12078

@robrap @mattdrayer


robrap:
:+1: looks fine, but I'm no css expert. Not sure if there is an alternative to using `important`. Mind adding an "after" screenshot?

PR #16, opened 2016-05-28
Transmit launch_presentation_locale parameter

haikuginger:
This is related to [LOC-78](https://openedx.atlassian.net/browse/LOC-78).

When using the edX LMS as a frontend for other LTI modules, this change enables us to transmit the user's preferred language setting (if present) as part of the POST request to set the iframe up.

If the user doesn't have a language set on their profile, or if they don't have an edX LMS profile set up, we won't transmit locale preferences, and the behavior will remain the same.

This PR is covered by the OpenCraft contributor agreement - thanks @antoviaque!

PR #17, opened 2016-06-04
Fixing problems with #16's use of launch_presentation_locale

haikuginger:
Fixes problems with #16 - we should now look at `language_proficiencies` to see if a single proficiency exists for us to use, rather than the defunct `language` attribute. I've also iterated the version number to 1.0.8.


douglashall:
👍 Thanks @haikuginger!

PR #18, opened 2016-06-05
Revert "Fixing problems with #16's use of launch_presentation_locale"

douglashall:
Reverts edx/xblock-lti-consumer#17

Sorry @haikuginger, I tested this after I merged and it was still not working. Turns out I was wrong about where we store the user's preferred language. The `openedx.core.djangoapps.user_api.models.UserPreference` is what we were looking for. You should be able to get the list of user preferences from the user object and filter by `key=pref-lang`. There should only ever be one UserPreference model with the key field set to pref-lang.

Please give this a manual test on your end next time.

### DIFF
--- a/lti_consumer/lti.py
+++ b/lti_consumer/lti.py
@@ -152,14 +152,7 @@ class LtiConsumer(object):
             real_user_object = self.xblock.runtime.get_real_user(self.xblock.runtime.anonymous_student_id)
             self.xblock.user_email = getattr(real_user_object, "email", "")
             self.xblock.user_username = getattr(real_user_object, "username", "")
-            user_profile = getattr(real_user_object, "profile", None)
-            if user_profile is not None:
-                language_proficiencies = user_profile.language_proficiencies.all()
-                if len(language_proficiencies) == 1:
-                    # We only want to send the user's language preference if the user has
-                    # a single preferred language - having multiple language proficiencies
-                    # isn't an API violation, but can't be set up via the UI.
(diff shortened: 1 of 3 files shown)


haikuginger:
Okay - just to make sure I'm doing everything you need, this is my understanding of the steps:

1. Install a fresh devstack VM

2. Find or implement an openly-available LTI provider example (suggestions?)

3. Implement a change to use `user.preferences.filter(key='pref-lang').value`

4. Update my devstack to use my fork of `xblock-lti-consumer` rather than the base

5. Attach the LTI provider found in step 2 to my devstack

6. Verify transmission of the `launch_presentation_locale` parameter by sniffing the in-browser POST request from the LTI consumer when accessing the course

I just want to make sure I've got a good idea of what your expectations are.

douglashall:
@haikuginger Yes, this sounds like a good plan to ensure that the feature is working the way you expect it to. Re: step 2, you don't need a real LTI provider to verify that the new LTI launch parameter is being sent as expected. You can use your browser's  debugging tools to inspect the `lti_launch` POST request that is made when attempting to launch an LTI provider and verify that `launch_presentation_locale` is included as a POST parameter in the request and that it is what you would expect it to be based on your user's language preference. Thanks again.

haikuginger:
@douglashall - sounds good! Yeah, if there's no validation process to check that the LTI provider is "valid" when being added, that makes the process much simpler. I'll have an implementation for you tomorrow.

PR #19, opened 2016-06-07
Fixing launch_presentation_locale (promise!); updating to v1.0.9

haikuginger:
@douglashall - I've updated the code and done manual testing to confirm that the `launch_presentation_locale` parameter is now being passed correctly per our conversation in #18 - sorry for the delay.

I used httpbin.org/post as the LTI target - that let me peek into the request parameters without needing to do any request sniffing.

screen shot


douglashall:
👍 Thanks @haikuginger!

PR #22, opened 2016-10-04
Add a compatibility function for get_course_lti_endpoints

xcompass:
This allows this xblock being hooked into get_course_lti_endpoints for
discovering LTI result/outcome endpoints.

**Background**: The old xmodule supports enumerates LTI endpoints so that they can be discovered and consumed by external tools. However, this functionality is missing when xmodule was refactored into xblock. Currently, only the old LTI xmodules can be discovered when /courses/COURSE_ID/lti_rest_endpoints/ is called. 

**Testing**:
- Create an LTI consumer component by following instructions here: http://edx.readthedocs.io/projects/edx-partner-course-staff/en/latest/exercises_tools/lti_component.html 
- Access htts://LMS_URL/courses/COURSE_ID/lti_rest_endpoints/

The endpoints for LTI component should be listed in JSON format.   

This PR is needed by https://github.com/edx/edx-platform/pull/13639 to enable the endpoints to query both old xmodules and new xblocks.

### DIFF
--- a/lti_consumer/lti_consumer.py
+++ b/lti_consumer/lti_consumer.py
@@ -563,6 +563,20 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
         return self.runtime.handler_url(self, "outcome_service_handler", thirdparty=True).rstrip('/?')
 
     @property
+    def result_service_url(self):
+        """
+        Return URL for results.
+
+        To test LTI on sandbox we must use http scheme.
+
+        While testing locally and on Jenkins, mock_lti_server use http.referer
+        to obtain scheme, so it is ok to have http(s) anyway.
+
+        The scheme logic is handled in lms/lib/xblock/runtime.py
+        """
+        return self.runtime.handler_url(self, "result_service_handler", thirdparty=True).rstrip('/?')
+
+    @property
     def prefixed_custom_parameters(self):
         """
         Apply prefix to configured custom LTI parameters
@@ -860,3 +874,17 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
             float: The css position offset to apply to the modal window
         """
         return (100 - viewport_percentage) / 2
+
+    def get_outcome_service_url(self, service_name="grade_handler"):
+        """
+        This function is called by get_course_lti_endpoints when using LTI result service to
+        discover the LTI result endpoints.
+        """
+
+        # using string as mapped value instead of attributes to avoid unnecessary calls as both urls
+        # are @property.
+        mapping = {
+            'grade_handler': 'outcome_service_url',
+            'lti_2_0_result_rest_handler': 'result_service_url'
+        }
+        return getattr(self, mapping[service_name])
--- a/lti_consumer/tests/unit/test_lti_consumer.py
+++ b/lti_consumer/tests/unit/test_lti_consumer.py
@@ -185,6 +185,17 @@ class TestProperties(TestLtiConsumerXBlock):
         self.xblock.runtime.handler_url.assert_called_with(self.xblock, 'outcome_service_handler', thirdparty=True)
         self.assertEqual(url, handler_url)
 
+    def test_result_service_url(self):
+        """
+        Test `result_service_url` calls `runtime.handler_url` with thirdparty kwarg
+        """
+        handler_url = 'http://localhost:8005/result_service_handler'
+        self.xblock.runtime.handler_url = Mock(return_value="{}/?".format(handler_url))
+        url = self.xblock.result_service_url
+
+        self.xblock.runtime.handler_url.assert_called_with(self.xblock, 'result_service_handler', thirdparty=True)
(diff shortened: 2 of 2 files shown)


douglashall:
👍 This looks great. Thanks for the contribution.

@robrap Would you be able to review this?

lti_consumer/lti_consumer.py
```diff
@@ -860,3 +860,15 @@ def _get_modal_position_offset(self, viewport_percentage):
...
+        """
+        This function is called by get_course_lti_endpoints when using LTI result service to
+        discover the LTI result endpoints. It is similar to outcome_service_url property but
+        with a mapping for backward compatibility.
+        """
+        mapping = {
+            'grade_handler': 'outcome_service_handler',
+            'lti_2_0_result_rest_handler': 'result_service_handler'
+        }
+        return self.runtime.handler_url(self, mapping[service_name], thirdparty=True).rstrip('/?')
```
robrap: Would it make sense to add a `result_service_url` call and instead of mapping to the service name, map to the function itself?  This would remove a little code redundancy, take advantage of the comments in the `outcome_service_url` property, and would have a little more consistency around the implementation.  Thoughts?

xcompass: Not sure if I understand correctly. The get_outcome_service_url function was missing and it was called from https://github.com/edx/edx-platform/blob/master/lms/djangoapps/courseware/views/views.py#L1075. Sorry, not sure if I understand your comment. The get_outcome_service_url was missing from lti_consumer (was throwing exception when the endpoint was called). It is called from here: https://github.com/edx/edx-platform/blob/master/lms/djangoapps/courseware/views/views.py#L1075. The service_name parameter was specific for LTI xmodule functions (grade_handler and lti_2_0_result_rest_handler). During the refactor to xblock, those function names changed. So that's why we need a mapping.

robrap: I guess I wasn't clear about my refactor request.  All looks good, so this is a minor request.  I'll explain more soon, but we have hackathon followed by a holiday on Monday.  I hope early next week will be fine.

robrap: Sorry I forgot about this.  Here is untested code to explain what I meant.  I still feel new to Python, so not sure if there are any issues with this.
```
def get_outcome_service_url(self, service_name="grade_handler"):
        """
        This function is called by get_course_lti_endpoints when using LTI result service to
        discover the LTI result endpoints.
        """
        mapping = {
            'grade_handler': outcome_service_url,
            'lti_2_0_result_rest_handler': result_service_url,  # new function needed
        }
        return mapping[service_name]()
```
I removed the docstring comment about `outcome_service_url`, because this function now just calls the appropriate property, which has its own comments.

Does this make my original comment any more clear?  Does this seem reasonable @xcompass ? @douglashall?

xcompass: Thanks. I see what you mean. Updated. How does it look now?

lti_consumer/tests/unit/test_lti_consumer.py
```diff
@@ -528,6 +539,39 @@ def test_delete_result_called(self, mock_parse_suffix, mock_delete_result):
...
+        url = self.xblock.get_outcome_service_url('grade_handler')
+
+        self.xblock.runtime.handler_url.assert_called_with(self.xblock, 'outcome_service_handler', thirdparty=True)
+        self.assertEqual(url, handler_url)
+
+    def test_get_outcome_service_url_with_service_name_lti_2_0_result_rest_handler(self):
+        """
+        Test `get_outcome_service_url` calls with service name lti_2_0_result_rest_handler
+        """
+        handler_url = 'http://localhost:8005/outcome_service_handler'
```
robrap: outcome_service_handler  => result_service_handler

robrap:
@xcompass: I had one minor typo fix for you.  Feel free to fix and amend/squash this minor change.
👍 with this change.

lti_consumer/tests/unit/test_lti_consumer.py
xcompass: Good catch. Done

gsong:
@xcompass @robrap @douglashall Should we go ahead and merge this or close it out?

xcompass:
I would be happy to see it's merged :)

robrap:
@gsong @xcompass:  This is waiting on our team.  I will try to see how we can get this merged as soon as possible @xcompass.  Thanks for your patience.

robrap:
Actually, I'm going to merge this one.  What I am going to try to make happen as soon as possible is tagging it and getting it into edx-platform along with the other PR.

robrap:
@xcompass: Someone merged another PR to xblock-lti-consumer and bumped the version to 1.1.1, but there is no release tag and it hasn't been updated in edx-platform.  

So, I need to get that resolved before being able to release this change.  

If interested, you can follow this conversation about releases here: https://github.com/edx/xblock-lti-consumer/pull/24#issuecomment-270938356

PR #25, opened 2017-01-09
added compiled translations

ziafazal:
added compiled translations.
@douglashall FYI


douglashall:
👍 Also, please work with @robrap on the versioning and rollout of these and other changes. Thanks.

PR #26, opened 2017-01-26
TNL-6400:no need to translate string coming out of database backend

ziafazal:
This PR removes translation around those strings coming out of database and can have empty values.
@douglashall could you please review? 
@asadiqbal08 FYI.


lti_consumer/lti_consumer.py
```diff
@@ -849,11 +849,11 @@ def _get_context_for_template(self):
             'has_score': self.has_score,
             'weight': self.weight,
             'module_score': self.module_score,
-            'comment': self.ugettext(sanitized_comment),
+            'comment': sanitized_comment,
```
douglashall: Should we remove the `ugettext` call from `display_name` above as well?

douglashall: @ziafazal What do you think about the above question?

ziafazal: @douglashall yeah i think we should remove it although it has default values but still it is coming from database. Removed.

douglashall:
👍

douglashall:
@ziafazal We should also bump the version here and update in edx-platform.

PR #28, opened 2017-05-02
EDUCATOR-121 – Conversion to Django App.

Qubad786:
This PR Adds basic barebones to convert LTI Consumer XBlock into a standalone Django App.


lti_consumer/fields.py
```diff
@@ -0,0 +1,132 @@
...
+    """
+    if lookup == 'in':
+        stripped_value = [_strip_object(el) for el in value]
+    else:
+        stripped_value = _strip_object(value)
+    return stripped_value
+
+
+# TODO: This code has been copies and pasted from openedx.core.djangoapps.xmodule_django
+# Need to figure out how can we access stuff from edx-platform.
```
Qubad786: @nasthagiri @jmbowman , what is the preferred way if we need modules from edx-platform – like here, I needed `CourseKey` from openedx.

lti_consumer/forms.py
```diff
@@ -0,0 +1,46 @@
...
+Defines a form for providing validation of LTI consumer course-specific configuration.
+"""
+import logging
+
+from django import forms
+from opaque_keys import InvalidKeyError
+from opaque_keys.edx.locator import CourseLocator
+
+# TODO: find a way to access the modulestore..
+from xmodule.modulestore.django import modulestore
```
Qubad786: @nasthagiri @jmbowman another similar case, I needed `modulestore` from `xmodule.modulestore.django`

muzaffaryousaf:
@cpennington can you also take a look at this PR ?

muzaffaryousaf:
FYI @nasthagiri @cpennington This is follow up work from this PR https://github.com/edx/edx-platform/pull/14766 . can you guys take a look if we're going in right direction ?

muzaffaryousaf:
@jmbowman Can you also take a look at this PR. The purpose of this is to make LTI xBlock an Django app, in order to define LTI's configuration models in this own repo. For further details please check conversation on this PR https://github.com/edx/edx-platform/pull/14766#discussion-diff-108918859R273

muzaffaryousaf:
@Qubad786 Add this branch [here](https://github.com/edx/xblock-lti-consumer/blob/master/.travis.yml#L10) to run the tests on Travis

lti_consumer/fields.py
nasthagiri: @Qubad786 We'll need to move the `CourseKeyField` functionality out of edx-platform into a common core repo.  @bmedx is making headway on this front by pulling out the `modulestore`, creating a new repo for the modulestore, and also creating a new edx-platform-core repo where some of the core/common functionality can go.

I imagine we'll want to move the entire `openedx/core/djangoapps/xmodule_django` app out of edx-platform so we can reuse this here.  Unless @clintonb or @rlucioni have already run into this in needing to store Course-Keys in SQL tables in other IDAs - and have created another library for this.

In the meantime, I'm glad you've added a `TODO` here to address this issue.  It would be good to include a JIRA ticket number with the TODO (and a back-pointer to this code) so we can keep track of this tech-debt.

lti_consumer/forms.py
nasthagiri: For the `modulestore`, as I mentioned above, @bmedx is actively working on pulling it out of edx-platform.  Once he does so, you'll be able to use it here.

In the meantime, perhaps you can disable the validation of the course-key in this code.

nasthagiri:
@Qubad786 @muzaffaryousaf Thanks for prioritizing this recent tech-debt !!  I like how this is moving forward - it's a great attempt for us to validate our future direction for introducing django dependency in xBlocks.  FYI @andy-armstrong !

lti_consumer/fields.py
clintonb: We have not extracted this. We store course keys as strings.

Qubad786:
Since this was blocking [WL-415](https://openedx.atlassian.net/browse/WL-415), we have decided to minimize our changes and we are only converting LTI to Django app as part of this PR. There'll be another PR which will incorporate the moving of LTI-related models to this repo.

@jmbowman, @cpennington, @nasthagiri please have a look.

Qubad786:
@muzaffaryousaf FYI.

.travis.yml
```diff
@@ -9,5 +9,5 @@ script:
   - make test
 branches:
     only:
-      - master
+      - mrehan/convert-lti-to-django-app
```
jmbowman: Should the branch restriction just be removed?  We usually want to run tests on all PRs.

Qubad786: Yes, this shall be removed before the merge.

bradenmacdonald:
@muzaffaryousaf @Qubad786 Is it actually necessary to make this not just a django application but a full django site (with `settings/test.py`, `manage.py`), etc.? That seems like a lot of overhead/boilerplate that I'd like to avoid, especially if other XBlocks also start using django. (It will also make django upgrades in edx-platform more work if that potentially requires changes to every xblock repo as well).


We've [been using django models in the Problem Builder XBlock](https://github.com/open-craft/problem-builder/blob/6623c55/problem_builder/models.py) for quite a while, and our approach there is just to use the existing workbench django site infrastructure for tests/CI, so that it's not necessary to include that other django boilerplate.

Qubad786:
@bradenmacdonald I've seen how the tests wouldn't need `manage.py` and `settings`  in problem builder, but how would we create migrations for our app? After this PR, we also need to move here the LTI related config models from: https://github.com/edx/edx-platform/pull/14766 which need 2 externals `xmodule.modulestore.django.modulestore` and `openedx.core.djangoapps.xmodule_django` (currently in edx-platform). 

Apart from that, I am not getting this part of read me: https://github.com/open-craft/problem-builder#workbench-installation-and-settings (i.e. the command as there isn't `manage.py` in problem builder Or is this for integrating problem builder into another Django app?) Also, I am seeing imports from `edx-platform` around https://github.com/open-craft/problem-builder/blob/master/problem_builder/tasks.py It seems this task would only run when integrated with `edx-platform`? Thanks for looking!

bradenmacdonald:
@Qubad786 An XBlock can be a django app (containing migrations and models) without being a django site (containing settings and manage.py, etc.) - just like any other django app/plugin.

Problem Builder for example has both [migrations](https://github.com/open-craft/problem-builder/tree/master/problem_builder/migrations) and [models](https://github.com/open-craft/problem-builder/blob/master/problem_builder/models.py). The trick that makes this work is that [support for `problem_builder` is hard coded into the platform via `OPTIONAL_APPS`](https://github.com/edx/edx-platform/blob/cd56cc56cdd8ccb92792d5cec3394e8ea16b3d76/lms/envs/common.py#L2730).

> need 2 externals xmodule.modulestore.django.modulestore and openedx.core.djangoapps.xmodule_django (currently in edx-platform)

Yes, if you want to use `CourseKeyField` for example, your XBlock will only work within edx-platform. You have to decide either to live with that, or to implement support for falling back to `CharField` when `CourseKeyField` is unavailable, such as in the Workbench. (Or extract `CourseKeyField` to an external repository, which makes sense).

However, you have that problem whether or not you include `manage.py` and a settings file. My feedback here is just about removing those files.

> Apart from that, I am not getting this part of read me: https://github.com/open-craft/problem-builder#workbench-installation-and-settings (i.e. the command as there isn't manage.py in problem builder Or is this for integrating problem builder into another Django app?)

The `manage.py` command there refers to the [manage.py included in xblock-sdk](https://github.com/edx/xblock-sdk/blob/master/manage.py).

> Also, I am seeing imports from edx-platform around https://github.com/open-craft/problem-builder/blob/master/problem_builder/tasks.py It seems this task would only run when integrated with edx-platform? 

Yes, most parts of Problem Builder are designed to run in either the XBlock SDK (Workbench) or in edx-platform, but the `tasks.py` only work in the platform.


Currently, XBlocks are designed to be semi-generic and to support many different runtimes, but in practice there's only one runtime environment which is really used, which is edx-platform. As XBlock developers, we often use the XBlock SDK as an alternate runtime for tests, but that's not out of choice - that's just because there's currently no easy way to get a CI build working for XBlocks using the edx-platform runtime. Eventually I'd like to see/build a way to do that, and this whole thing will become easier as we can focus on the edx-platform environment.

nasthagiri:
@bradenmacdonald This is excellent information from a team that has already been iterating on this.  Thank you for your input.  FYI @andy-armstrong @robrap 

_> currently no easy way to get a CI build working for XBlocks using the edx-platform runtime_
do you have more input on this?  is there something on our end that could help with this effort?

Qubad786:
@bradenmacdonald awesome! this helped a lot. I've simplified by keeping in mind that this XBlock will run in a runtime having Django pre-installed. Please have a look. Thanks!

Qubad786:
@bradenmacdonald can you please review this today whenever possible? thanks!

bradenmacdonald:
@Qubad786  Sorry I didn't have time today - I'll try to review for you and reply to @nasthagiri's question tomorrow.

bradenmacdonald:
@nasthagiri et al.

> > currently no easy way to get a CI build working for XBlocks using the edx-platform runtime
>
> do you have more input on this? is there something on our end that could help with this effort?

So the main problem is that for an XBlock repo like this (xblock-lti-consumer), when running tests on, say, TravisCI or CircleCI, we can't use any edx-platform functionality. This severely limits the usefulness of the tests. It would be nice if we could just include "pip install openedx" as one of the pre-test commands the TravisCI will run, and then could test the XBlock using libraries from the `openedx` namespace as well as doing Selenium tests within the LMS and Studio. However, that's presumably never going to fly due to the complexity of installing a working edx environment.

Instead, I have two suggestions - but these need review and more extensive planning, I think:

1. First and simplest: implement a plugin mechanism that allows XBlocks to register their test suite with the LMS/Studio so that the XBlock's independent test suite gets included with the platform's test suite. (This includes Selenium tests). That way, for example, every XBlock that is used on edx.org can have its test suite within its own repository (where it belongs) but the Jenkins build of each PR will ensure that no XBlocks have been broken. (e.g. it would be trivial to make a change to the Studio JS/HTML that breaks this XBlock's editing UI, and that should cause a test that lives in this repo to fail - it's currently not possible to achieve that). I think the major concern here is that doing so could significantly slow down the build for platform PRs.
1. Then (or alternately), separate XBlock projects/repos like this one need to find a way to do a CI build for their PRs that runs the test suite against edx-platform. We know it's possible to build edX platform on CircleCI (which is simpler to set up and more accessible to the XBlock/open-source community than edX's Jenkins box) because [it's in place already for the solutions fork](https://circleci.com/gh/edx-solutions/edx-platform). So I'd love to see a CircleCI template that XBlocks can use which installs the edX platform and runs the XBlock's test suite within the platform.

What do you think?

Qubad786:
@nasthagiri may you please review this too? 
Things are simplified now and this XBlock will be pluggable in a runtime having Django pre-installed. To use it as a Django app in `edx-platform`, we will add `lti_consumer` to either `OPTIONAL_APPS` or `INSTALLED_APPS` within edx-platform and I'll do that adding in platform with migration of the config models into this repo.

Thanks!

Qubad786:
@nasthagiri I've setup [sandbox](https://studio-ltidjangoapp.sandbox.edx.org/course/course-v1:UET+CS111+2018_LTI) with edx-paltform. May you please give this a review?

Qubad786:
@bradenmacdonald I am seeing odd behavior. I get 100% coverage while running nose tests directly(this has been done in master) via: `nosetests --with-coverage --cover-package=lti_consumer --verbosity=1`
```
$ nosetests --with-coverage --cover-package=lti_consumer --verbosity=1
/Users/muhammadrehan/.virtualenvs/lti-consumer-xblock/lib/python2.7/site-packages/xblock/runtime.py:576: FutureWarning: IdGenerator will be required in the future in order to support XBlockAsides
  warnings.warn("IdGenerator will be required in the future in order to support XBlockAsides", FutureWarning)
..................................................................................................
Name                           Stmts   Miss  Cover
--------------------------------------------------
lti_consumer.py                    1      0   100%
lti_consumer/exceptions.py         2      0   100%
lti_consumer/lti.py              107      0   100%
lti_consumer/lti_consumer.py     193      0   100%
lti_consumer/oauth.py             54      0   100%
lti_consumer/outcomes.py          82      0   100%
lti_consumer/utils.py              3      0   100%
--------------------------------------------------
TOTAL                            442      0   100%
----------------------------------------------------------------------
Ran 98 tests in 1.070s

OK
```
But when I ran these same tests via Django with Nose backend (in this PR). I am getting 88% coverage, although number of tests ran, are the same:
```
$ make test
./scripts/test.sh
nosetests --with-coverage --cover-package=lti_consumer --verbosity=1
Creating test database for alias 'default'...
..................................................................................................
Name                           Stmts   Miss  Cover
--------------------------------------------------
lti_consumer.py                    1      1     0%
lti_consumer/exceptions.py         2      2     0%
lti_consumer/lti.py              107     15    86%
lti_consumer/lti_consumer.py     193     79    59%
lti_consumer/oauth.py             54     12    78%
lti_consumer/outcomes.py          82     12    85%
lti_consumer/utils.py              3      2    33%
--------------------------------------------------
TOTAL                            442    123    72%
----------------------------------------------------------------------
Ran 98 tests in 0.290s

OK
Destroying test database for alias 'default'...
```
I am not getting anywhere with this, is this expected?

Qubad786:
@bradenmacdonald , @mushtaqak I've fixed the decreased coverage, I'll merge now.

bradenmacdonald:
@Qubad786 Great! Sorry I wasn't able to help you quickly on this.

PR #35, opened 2017-09-13
make system tolerant of ascii strings with unicode characters

noraiz-anwar:
### Description

In this PR, any non-unicode string is first converted to unicode so it can be encoded successfully in [parse_grade_xml_body](https://github.com/edx/xblock-lti-consumer/blob/master/lti_consumer/outcomes.py#L23).
We need to do this because encoding is failed for strings (not specified as unicode) which have unicode chars. (When not specified as unicode, python 2.7.x use ASCII for encoding, unicode characters can not be encoded using ASCII).

### Reviewers
If you've been tagged for review, please check your corresponding box once you've given the :+1:.
- [x] Style and readability review: @attiyaIshaque 
- [x] Code review: @edx/educator-all 
- [x] Code review: @awaisdar001 

### Post-review
- [x] Rebase and squash commits

### DIFF
--- a/lti_consumer/outcomes.py
+++ b/lti_consumer/outcomes.py
@@ -40,7 +40,7 @@ def parse_grade_xml_body(body):
     """
     lti_spec_namespace = "http://www.imsglobal.org/services/ltiv1p1/xsd/imsoms_v1p0"
     namespaces = {'def': lti_spec_namespace}
-    data = body.strip().encode('utf-8')
+    data = body.strip()
 
     try:
         parser = etree.XMLParser(ns_clean=True, recover=True, encoding='utf-8')  # pylint: disable=no-member
--- a/lti_consumer/tests/unit/test_outcomes.py
+++ b/lti_consumer/tests/unit/test_outcomes.py
@@ -1,3 +1,4 @@
+# -*- coding: utf-8 -*-
 """
 Unit tests for lti_consumer.outcomes module
 """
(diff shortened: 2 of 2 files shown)


lti_consumer/outcomes.py
```diff
@@ -40,7 +40,11 @@ def parse_grade_xml_body(body):
     """
     lti_spec_namespace = "http://www.imsglobal.org/services/ltiv1p1/xsd/imsoms_v1p0"
     namespaces = {'def': lti_spec_namespace}
-    data = body.strip().encode('utf-8')
+    xml_body = body.strip()
+
+    if isinstance(xml_body, str):
```
nedbat: Why aren't we sure what type the body will be? The docstring says it is str.  There seems to be only one caller of this function, and it passes `request.body` where `request` is `DjangoWebobRequest`.  Shouldn't we know what type it will be?

noraiz-anwar: `unicode(xml_body, 'utf-8')` will raise exception if first parameter i.e  `xml_body` is unicode. Intention of checking this is to make sure that conversion to unicode is safe.
I am not sure if `request.body` will always be `str`, we can remove this check if we are sure that it will be a `str` object in all cases.

awaisdar001: The docstring does say that it is `str` but some LTI responses contain unicode characters upon which `.encode` breaks out (More information in EDUCATOR-1218). 

I communicated this with @douglashall who originally contributed this code, and he mentioned `Would it be possible for us to handle both UTF-8 encode and non-UTF-8 encoded requests?`

nedbat: If we are sure that `body` is a `str`, then all of this change can be replaced with: `data = body.strip()`.  It doesn't make sense to call `.encode` on a `str`.  This can clear up some confusion: [Pragmatic Unicode](http://bit.ly/unipain).

I feel strongly that Unicode handling is something that we can and should understand. There's no need to guess, and there's no need to play encode/decode whack-a-mole.

noraiz-anwar: @nedbat I agree that we don't need to get bytes (by calling `.encode`) from `str` (bytes). 
I am trying to handle a scenario where body type is `str` but it contains unicode characters in it. I have implemented it by converting `body` to unicode (if it is `str`) and finally getting bytes by calling `.encode`.
Following are my expectations by this fix:

- If body type is unicode, we don't need to convert it to unicode. We will call `.encode` on it to get bytes.
- If body type is 'str', it may unexpectedly contain unicode characters in it ([Filed issue](https://openedx.atlassian.net/browse/EDUCATOR-1218)), convert it into unicode and call `.encode` to get bytes.
Am I missing something here? Thanks for sharing an informative piece of reading.

nedbat: If the body type is str, then this will work:  `data = body.strip()`.  data will then be a str (byte string) with utf8 bytes in it.  There's no need to convert it to unicode.  The exception in the filed issue is a failure of the `.encode` call, which is completely unnecessary.  Calling encode on a str never makes sense.  Python is trying to help you by implicitly decoding the str with the ascii codec so that you'll have a unicode string, which does make sense to .encode.  The error is of our own making.

I'm sorry, but this code is overcomplicating things.  It really can be much simpler.  Please read or watch bit.ly/unipain.

noraiz-anwar: @nedbat Thank you. Implicit operations were not clear. I think I got it now.
I will continue referring this doc for help in future. 
Thanks for your help. :)

PR #37, opened 2018-02-08
Update iframe allowances for chrome v64

fysheets:
### Description
Due to a chrome update some of our LTI blocks are broken in chrome. See documentation here: https://dev.chromium.org/Home/chromium-security/deprecating-permissions-in-cross-origin-iframes

### Reviewers
If you've been tagged for review, please check your corresponding box once you've given the :+1:.
- [x] Code review: @awaisdar001 
- [x] Code review: @efischer19 

FYI: @edx/educator-escalation 

### Post-review
- [x] Rebase and squash commits

### DIFF
--- a/lti_consumer/templates/html/lti_iframe.html
+++ b/lti_consumer/templates/html/lti_iframe.html
@@ -7,4 +7,5 @@
     allowfullscreen="true"
     webkitallowfullscreen="true"
     mozallowfullscreen="true"
+    allow="microphone *; camera *; midi *; geolocation *; encrypted-media *"
 ></iframe>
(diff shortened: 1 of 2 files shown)


lti_consumer/templates/html/lti_iframe.html
```diff
@@ -7,4 +7,5 @@
     allowfullscreen="true"
     webkitallowfullscreen="true"
     mozallowfullscreen="true"
+    allow="microphone *; camera *; midi *; geolocation *; encrypted-media *"
```
awaisdar001: Would you also like to lower the permissions here? 
`microphone *; camera *`

fysheets: No, I updated my comment on the ticket, after a brief discussion with someone on the security team, we are going to begin with allowing the same access as they had in chrome v63. I've asked to have a larger discussion with the security group and we can revisit this after their recommendations.

PR #44, opened 2019-01-06
Allow plugins for the LTI XBlock to pass extra parameters to the provider

OmarIthawi:
### Overview
This pull request makes this XBlock extensible for adding parameters to the LTI provider in addition to the existing ones.

#### Why
A customer needs `team` and `cohort` informaiton to be passed through the LTI channel to simplify integration with their LTI provider. At the same time we'd rather not fork the XBlock, so an Open-Closed model is needed.

### More Detailed Documentation
I've added a couple of sections to `README.rst` to document specifying design and configuration steps of the new LTI parameter processors. Please check it out for more information about the changes that this pull request introduces.
#### Example Usage
I've written a package that we'll need to use for our customer to pass some personal user information in addition to the cohort and team info: https://github.com/appsembler/tahoe-lti/pull/1

### TODO
 - [x] Design review by the Open edX Architects [see more on Slack.](https://openedx.slack.com/archives/C0RU5BTCP/p1541533584041800)
 - [x] Privacy and GDPR review by edX: The edX legal team have confirmed that passing the parameters optionally is OK from the edX privacy standards.
 - [x] Update the README
 - [x] Add tests
 - [x] Add a workbench scenario
 - [x] Test on devstack Studio and LMS
    * `XBLOCK_SETTINGS` are not supported in Studio. My best guess is that `SettingsService` isn't enabled in Studio.
 - [ ] **Optional**: Get the workbench updates merged: https://github.com/edx/xblock-sdk/pull/142

### Timeline
 - [x] Get the critical design and architecture review by Nov 14th, 2018.
   * The discussion has been concluded in [Open edX #architecture Slack](https://openedx.slack.com/archives/C0RU5BTCP/p1541533584041800).
 - [ ] We (Appsembler) would love to get this PR merged by Q1 2019 so we can deploy at our production environment before the  next Open edX release upgrade. The earlier the better.

### DIFF
--- a/lti_consumer/lti_consumer.py
+++ b/lti_consumer/lti_consumer.py
@@ -53,6 +53,7 @@ What is supported:
 import logging
 import bleach
 import re
+from importlib import import_module
 import json
 import urllib
 
@@ -164,6 +165,7 @@ class LaunchTarget(object):
 
 
 @XBlock.needs('i18n')
+@XBlock.wants('settings')
 @XBlock.wants('lti-configuration')
 class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
     """
@@ -246,6 +248,8 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
 
         Otherwise error message from LTI provider is generated.
     """
+
+    block_settings_key = 'lti_consumer'
 
     display_name = String(
         display_name=_("Display Name"),
@@ -421,19 +425,84 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
         default=False,
         scope=Scope.settings
     )
+    enable_processors = Boolean(
+        display_name=_("Send extra parameters"),
+        help=_("Select True to send the extra parameters, which might contain Personally Identifiable Information. "
+               "The processors are site-wide, please consult the site administrator if you have any questions."),
+        default=False,
+        scope=Scope.settings
+    )
 
     # Possible editable fields
     editable_field_names = (
         'display_name', 'description', 'lti_id', 'launch_url', 'custom_parameters',
         'launch_target', 'button_text', 'inline_height', 'modal_height', 'modal_width',
         'has_score', 'weight', 'hide_launch', 'accept_grades_past_due', 'ask_to_send_username',
-        'ask_to_send_email'
+        'ask_to_send_email', 'enable_processors',
     )
+
+    @staticmethod
+    def workbench_scenarios():
+        """
+        Gather scenarios to be displayed in the workbench
+        """
+        scenarios = [
+            ('LTI Consumer XBlock',
+             '''<sequence_demo>
+                    <lti_consumer
+                        display_name="LTI Consumer - New Window"
+                        lti_id="test"
+                        description=""
+                        ask_to_send_username="False"
+                        ask_to_send_email="False"
+                        enable_processors="True"
+                        launch_target="new_window"
+                        launch_url="https://lti.tools/saltire/tp" />
+
+                    <lti_consumer
+                        display_name="LTI Consumer - IFrame"
+                        lti_id="test"
+                        ask_to_send_username="False"
+                        ask_to_send_email="False"
+                        enable_processors="True"
+                        description=""
+                        launch_target="iframe"
+                        launch_url="https://lti.tools/saltire/tp" />
+                </sequence_demo>
+             '''),
+        ]
+        return scenarios
 
     def validate_field_data(self, validation, data):
         if not isinstance(data.custom_parameters, list):
             _ = self.runtime.service(self, "i18n").ugettext
             validation.add(ValidationMessage(ValidationMessage.ERROR, unicode(_("Custom Parameters must be a list"))))
+
+    def get_settings(self):
+        """
+        Get the XBlock settings bucket via the SettingsService.
+        """
+        settings_service = self.runtime.service(self, 'settings')
+        if settings_service:
+            return settings_service.get_settings_bucket(self)
+
+        return {}
+
+    def get_parameter_processors(self):
+        """
+        Read the parameter processor functions from the settings and return their functions.
+        """
+        if not self.enable_processors:
+            return
+
+        try:
+            for path in self.get_settings().get('parameter_processors', []):
+                module_path, func_name = path.split(':', 1)
+                module = import_module(module_path)
+                yield getattr(module, func_name)
+        except Exception:
+            log.exception('Something went wrong in reading the LTI XBlock configuration.')
+            raise
 
     @property
     def editable_fields(self):
--- a/lti_consumer/tests/unit/test_lti.py
+++ b/lti_consumer/tests/unit/test_lti.py
@@ -11,7 +11,7 @@ from six import text_type
 
 from django.utils import timezone
 
-from lti_consumer.tests.unit.test_utils import FAKE_USER_ID, make_request
+from lti_consumer.tests.unit.test_utils import make_request, patch_signed_parameters
 from lti_consumer.tests.unit.test_lti_consumer import TestLtiConsumerXBlock
 
 from lti_consumer.lti import parse_result_json, LtiConsumer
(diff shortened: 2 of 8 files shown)


edx-webhook:
Thanks for the pull request, @OmarIthawi! I've created [OSPR-2958](https://openedx.atlassian.net/browse/OSPR-2958) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

All technical communication about the code itself will still be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).

OmarIthawi:
Thanks @natabene, it's still work in progress. I've marked it as `(WIP)`.

OmarIthawi:
@natabene I've completed working on this pull request, could you please submit it for review?

Please note that this change has been reviewed and approved by both of the architecture team at edX. It was also approved by the edX legal team in terms of user privacy and GDPR perspective.

natabene:
@ormsbee Could you confirm @edx/arch-review is good with this?

natabene:
@ormsbee Thanks for confirming, moving to eng review.

OmarIthawi:
Thanks @natabene for your help here. 

@ormsbee this PR is a follow up on our design discussion in [Slack](https://openedx.slack.com/archives/C0RU5BTCP/p1541533584041800) a couple of Arch. folks were involved, mainly Nimisha.

I have written some design documentation in the README, please let me know if you have any questions.

lti_consumer/tests/unit/test_lti.py
```diff
@@ -129,30 +129,27 @@ def setUp(self):
...
-    )
-    @patch(
-        'lti_consumer.lti_consumer.LtiConsumerXBlock.prefixed_custom_parameters',
-        PropertyMock(return_value={u'custom_param_1': 'custom1', u'custom_param_2': 'custom2'})
-    )
-    @patch(
-        'lti_consumer.lti_consumer.LtiConsumerXBlock.lti_provider_key_secret',
-        PropertyMock(return_value=('t', 's'))
-    )
-    @patch('lti_consumer.lti_consumer.LtiConsumerXBlock.user_id', PropertyMock(return_value=FAKE_USER_ID))
```
ormsbee: Could we keep this patching where it was? The `patch_signed_parameters` decorator is elegant, but it obscures what's going on and moves it further away from the test itself.

lti_consumer/tests/unit/test_lti.py
```diff
@@ -199,6 +188,20 @@ def test_get_signed_lti_parameters(self):
...
+        self.xblock.enable_processors = True
+
+        with patch('lti_consumer.lti_consumer.LtiConsumerXBlock.settings', PropertyMock(return_value={
+            'parameters_processors': [
+                'lti_consumer.tests.unit.test_utils:foo_processor',
+            ],
+        })):
+            params = self.lti_consumer.get_signed_lti_parameters()
+            assert '' == params['custom_author_country']
+            assert 'author@example.com' == params['custom_author_email']
```
ormsbee: Please use the standard `self.assertEqual()` and other assertion methods from `unittest.TestCase`.

lti_consumer/tests/unit/test_lti.py
```diff
@@ -259,7 +262,8 @@ def test_verify_result_headers_verify_content_type_true(self, mock_log):
 
         with self.assertRaises(LtiError):
             self.lti_consumer.verify_result_headers(request)
-            self.assertTrue(mock_log.called)
+
+        assert mock_log.error.called
```
ormsbee: Why is `mock_log.error` always called now?

lti_consumer/tests/unit/test_lti_consumer.py
```diff
@@ -22,7 +22,7 @@
 HTML_IFRAME = '<iframe'
 
 
-class TestLtiConsumerXBlock(unittest.TestCase):
+class TestLtiConsumerXBlock(TestCase):
```
ormsbee: Was this made necessary because we're checking Django settings now?

lti_consumer/tests/unit/test_lti_consumer.py
```diff
@@ -46,6 +46,34 @@ def test_descriptor(self):
         """
         self.assertEqual(self.xblock.descriptor, self.xblock)
 
+    def test_workbench_scenarios(self):
+        """
+        Basic tests that `workbench_scenarios()` returns a well formed scenario.
+        """
+        scenarios = self.xblock.workbench_scenarios()
+        assert isinstance(scenarios, list)
```
ormsbee: Again, please replace this with `self.assertIsInstance()`. Using the TestCase methods will often give you more helpful failure messages with test runners, so you get something like "expected list, but scenarios was tuple" instead of a more generic False.

lti_consumer/lti_consumer.py
```diff
@@ -421,20 +425,83 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
...
             validation.add(ValidationMessage(ValidationMessage.ERROR, unicode(_("Custom Parameters must be a list"))))
 
+    @property
+    def settings(self):
+        """
+        Gets the XBlock settings bucket via the SettingsService.
+        """
+        settings_service = self.runtime.service(self, 'settings')
+        if settings_service:
+            return settings_service.get_settings_bucket(self)
```
ormsbee: Nit: If there isn't a strong value to having `settings` and `parameter_processors` as properties, please consider just leaving them as regular methods. People often make assumptions that object attributes are cheap, and think nothing of doing repeated dictionary access calls to them (i.e. check for things in `obj.settings` in a loop). In general, properties are great for smoothly inserting logic into what was an attribute without breaking backwards compatibility, but they obscure that something interesting, possibly expensive, and possibly error-throwing is happening.

lti_consumer/tests/unit/test_lti.py
OmarIthawi: Great question @ormsbee. The coverage reports the original `self.assertTrue(mock_log.called)` was identified as a dead statement i.e. never reached. 

That's because `self.lti_consumer.verify_result_headers(request)` throws a an `LtiError` exception, so the next statement is never reached.

So I moved it outside the `with` statement. The other options was to delete the `mock_log` assertion.

Anyway tests are supposed to be stupid (in a good way) without branching, so subsequent statements should always be unconditional and unbranched since all of them must pass.

lti_consumer/tests/unit/test_lti.py
OmarIthawi: I found that copying the `@patch` statements would be redundant, so I thought DRY is better.

lti_consumer/tests/unit/test_lti.py
OmarIthawi: Great catch. Thanks @ormsbee.

To my best understanding we're moving away from assertion methods to native `assert` statements. Could you please let me know if that's correct?

For consistency it makes sense, but as far as I can tell, `pytest` with native `assert` is far superior to what `unittest.TestCase` can offer in terms of debugging capabilities.

lti_consumer/tests/unit/test_lti_consumer.py
OmarIthawi: Thanks @ormsbee. I wouldn't mind changing this to `self.assertIsInstance()` for consistency with the other tests. But this project uses `pytest` and it offers a far superior debugging messages and clear analysis of what actually happens when native `assert` statements are used.

Unlike `nosetests` which is relatively naive, and can't debug `assert` statements, therefore only `self.assert...()` methods are useful.

What do you think? Should we give up the extra debugging `assert` provides for consistency? or we should keep using `assert` statements and sacrifice some consistency?

lti_consumer/lti_consumer.py
OmarIthawi: Thanks @ormsbee for the great suggestion. There's not much value. I can use methods, and will do.

lti_consumer/tests/unit/test_lti.py
ormsbee: Interesting. I didn't realize that pytest recommended native `assert` or supported it so well. I'll read up on it today.

lti_consumer/lti_consumer.py
```diff
@@ -421,20 +425,83 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
...
+        """
+        Read the parameter processor functions from the settings and return their functions.
+        """
+        if not self.enable_processors:
+            return
+
+        for path in self.settings.get('parameters_processors', []):
+            module_path, func_name = path.split(':', 1)
+            module = import_module(module_path)
+            yield getattr(module, func_name)
```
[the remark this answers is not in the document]
OmarIthawi: Thanks @ormsbee! Logs makes sense, I'll trap and log errors.

> Any single misconfigured or badly coded parameter processor can't blow up the other properly configured and well behaving processors.

Just to confirm, your suggesting that an error in a processor should be logged as [`log.exception()`](https://stackoverflow.com/a/5191885) and the exception should be suppressed?

This will be tricky to debug for the LTI provider since it will result in parameters being "sometimes magically" vanishing from the LTI parameter list. I find that more obscure to debug, so here's what I suggest as a better solution to address your concerns:

There are two types of errors: configuration errors and processor runtime errors, that would be best to be handled differently:

**Configuration Errors:**
Such as `XBLOCK_SETTINGS` is being set to faulty configs, incorrect module path, providing some weird yaml syntax (e.g. list instead of a dict).

In this case it's not so bad to use `log.exception` for debugging. Then re-raise the exception causing 500 error. Which is much better than failing silently due a DevOps issue, as those error don't just happen, and would probably be best addressed ASAP by platform DevOps.

The error handling would look like the following:
```python
try:
    module_path, func_name = path.split(':', 1)
    module = import_module(module_path)
    yield getattr(module, func_name)
except:
    log.exception('Something went wrong in reading the LTI XBlock configuration for the processor "%s"', path)
    raise
```

**Processor Errors:**
A lot of things can go wrong in a the processor function, i.e. an external service isn't responding, or file access error, etc.

Such errors probably should best not to break the platform. However they should also have a graceful degradation, in addition to logging the error.

So each processor should be able to define a list of expected parameters and their default values. Those parameters are read by the XBlock and be provided as a fallback in case the processor failed. If a processor doesn't specify those parameters, then no default params will be added to the LTI request when an error happens.

A processor with default parameters would look like that:
```python
def user_email_processor(xblock):
    return {
        'email': xblock.user.email,
    }

user_email_processor.lti_xblock_default_params = {
    'email': '',
}
```

Somewhere in the LTI XBlock:
```python
for path, processor in self.xblock.get_parameter_processors():
    default_params = getattr(processor, 'lti_xblock_default_params', {})
    default_params.update(default_params)

    try:
        lti_parameters.update(processor(self.xblock))
    except:
        log.exception('Error in XBlock LTI parameter processor "%s"', path)
```

What do you think?

lti_consumer/tests/unit/test_lti_consumer.py
OmarIthawi: Yes @ormsbee, that's why.

OmarIthawi:
Thanks @ormsbee for the review. I've replied to all of the comments now. I'll do the quick changes, and will wait for your reply on the larger changes. Looking forward to it!

lti_consumer/tests/unit/test_lti.py
OmarIthawi: Thanks @ormsbee!

lti_consumer/lti_consumer.py
OmarIthawi: Done. Thanks @ormsbee!

lti_consumer/tests/unit/test_lti.py
OmarIthawi: Thanks, I'll keep it as-is for now.

lti_consumer/lti_consumer.py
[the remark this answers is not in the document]
OmarIthawi: Done.

OmarIthawi:
@ormsbee As far as I know, I've addressed all of the review comments. Could you please take a look again and let me know what do you think?

cc: @natabene

OmarIthawi:
Thanks for the reminder about the version @ormsbee. Will do the push soon.

OmarIthawi:
Done @ormsbee! I did the version bump, and now the commit has enough context.

OmarIthawi:
Whooo! Thanks you @ormsbee @nasthagiri @natabene for your help here!

PR #49, opened 2019-10-04
Upgrade to pyhton 3.

waheedahmed:
Updated to run tests on both python 2 and 3.

PROD-774

### DIFF
--- a/lti_consumer/lti.py
+++ b/lti_consumer/lti.py
@@ -5,15 +5,17 @@ For more details see:
 https://www.imsglobal.org/activity/learning-tools-interoperability
 """
 
-import logging
-import urllib
-import json
+from __future__ import absolute_import, unicode_literals
 
+import json
+import logging
+
+import six.moves.urllib.error
+import six.moves.urllib.parse
 from six import text_type
 
 from .exceptions import LtiError
 from .oauth import get_oauth_request_signature, verify_oauth_body_signature
-
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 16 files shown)


waheedahmed:
@azarembok, @feanil please review this PR.

lti_consumer/lti.py
```diff
@@ -83,14 +85,14 @@ def parse_result_json(json_str):
...
-            msg = "Could not convert resultScore to float: {}".format(err.message)
+            msg = "Could not convert resultScore to float: {}".format(str(err))
             log.error("[LTI] %s", msg)
             raise LtiError(msg)
 
     return score, json_obj.get('comment', "")
 
 
-class LtiConsumer(object):
+class LtiConsumer:  # pylint: disable=bad-option-value, old-style-class
```
feanil: You shouldn't remove this just yet because the code will need to run in both python2 and python3 for a little while so it should stay compatible with both.

lti_consumer/lti_consumer.py
```diff
@@ -155,7 +155,7 @@ def parse_handler_suffix(suffix):
 LaunchTargetOption = namedtuple('LaunchTargetOption', ['display_name', 'value'])
 
 
-class LaunchTarget(object):
+class LaunchTarget:  # pylint: disable=bad-option-value, old-style-class
```
feanil: Same as above, we will need this to run in both python2 and python3 for a little bit.

lti_consumer/outcomes.py
```diff
@@ -81,7 +83,7 @@ def parse_grade_xml_body(body):
     return imsx_message_identifier, sourced_id, score, action
 
 
-class OutcomeService(object):
+class OutcomeService:  # pylint: disable=bad-option-value, old-style-class
```
feanil: Same as above, we should not remove the explicit dependence on (object) just yet because this code will need to run in both python 2 and python3.

requirements.txt
```diff
@@ -1,8 +1,9 @@
 lxml
 bleach
+django==1.11.25
 oauthlib
 mako
 lazy
 git+https://github.com/edx/XBlock.git#egg=XBlock
-git+https://github.com/edx/xblock-utils.git@v1.0.0#egg=xblock-utils==v1.0.0
+git+https://github.com/edx/xblock-utils.git#egg=xblock-utils
```
feanil: If you're not going to be setting a version, this should be prefixed with a -e for safety.  Otherwise it might get updated correctly.

waheedahmed:
@feanil Addressed your comments, kindly take another look.

PR #52, opened 2020-02-03
avoid-escape-exception

uzairr:
TypeError exception is raised by lti-consumer xblock because request body is not str type.

PROD-1119

### DIFF
--- a/lti_consumer/outcomes.py
+++ b/lti_consumer/outcomes.py
@@ -162,15 +162,16 @@ class OutcomeService(object):  # pylint: disable=bad-option-value, useless-objec
             'imsx_messageIdentifier': 'unknown',
             'response': ''
         }
+        request_body = request.body.decode('utf-8')
 
         if not self.xblock.accept_grades_past_due and self.xblock.is_past_due:
             failure_values['imsx_description'] = "Grade is past due"
             return response_xml_template.format(**failure_values)
 
         try:
-            imsx_message_identifier, sourced_id, score, action = parse_grade_xml_body(request.body)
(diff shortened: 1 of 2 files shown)


lti_consumer/outcomes.py
```diff
@@ -162,15 +162,16 @@ def handle_request(self, request):
             'imsx_messageIdentifier': 'unknown',
             'response': ''
         }
+        request_body = request.body.decode('utf-8')
```
waheedahmed: I believe we should do it in `except` block where we are escaping it, because it might break something else while parsing it.

lti_consumer/tests/unit/test_outcomes.py
```diff
@@ -370,6 +370,20 @@ def test_grade_past_due(self):
         self.assertIn('failure', response)
         self.assertIn('Grade is past due', response)
 
+    @patch('lti_consumer.outcomes.parse_grade_xml_body')
+    def test_lti_error_not_raises_type_error(self, mock_parse):
```
waheedahmed: Does this test fails without the fix with error `a bytes-like object is required`?

lti_consumer/outcomes.py
uzairr: i did it intentionally because as per the docstring of `parse_grade_xml_body` it is expecting string object while our codebase is sending it bytestring.I think it will not break anything because `request.body` is not being used anywhere else in the method.

lti_consumer/tests/unit/test_outcomes.py
uzairr: Yes! it would be failed without the fix that is going to be added!

PR #54, opened 2020-02-26
Added img to bleach safe tags

AhtishamShahid:
## Issue:
If a comment has `` in the response it will not be rendered properly because it is escaped by `bleach`

## Fix: 
Added `img` in safe tags so bleach will ignore `img`  so now  tag will renderd properly

### DIFF
--- a/lti_consumer/lti_consumer.py
+++ b/lti_consumer/lti_consumer.py
@@ -574,8 +574,8 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
             try:
                 lti_id, key, secret = [i.strip() for i in lti_passport.split(':')]
             except ValueError:
-                msg = self.ugettext('Could not parse LTI passport: {lti_passport}. Should be "id:key:secret" string.').\
-                    format(lti_passport='{0!r}'.format(lti_passport))
+                msg = 'Could not parse LTI passport: {lti_passport!r}. Should be "id:key:secret" string.'
+                msg = self.ugettext(msg).format(lti_passport=lti_passport)
                 raise LtiError(msg)
 
             if lti_id == self.lti_id.strip():
@@ -694,9 +694,8 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
                     param_name, param_value = [p.strip() for p in custom_parameter.split('=', 1)]
                 except ValueError:
                     _ = self.runtime.service(self, "i18n").ugettext
-                    # pylint: disable=line-too-long
-                    msg = self.ugettext('Could not parse custom parameter: {custom_parameter}. Should be "x=y" string.').\
-                        format(custom_parameter="{0!r}".format(custom_parameter))
+                    msg = 'Could not parse custom parameter: {custom_parameter!r}. Should be "x=y" string.'
+                    msg = self.ugettext(msg).format(custom_parameter=custom_parameter)
                     raise LtiError(msg)
 
                 # LTI specs: 'custom_' should be prepended before each custom parameter, as pointed in link above.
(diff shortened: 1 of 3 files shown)


DawoudSheraz:
Please add a unit test. The changes are looking good otherwise.

awaisdar001:
@AhtishamShahid Looks good, but I have same concern as above.

lti_consumer/tests/unit/test_lti_consumer.py
```diff
@@ -825,6 +827,17 @@ def test_context_keys(self):
         for key in context_keys:
             self.assertIn(key, context)
 
+    @ddt.data('a', 'abbr', 'acronym', 'b', 'blockquote', 'code', 'em', 'i', 'li', 'ol', 'strong', 'ul', 'img')
```
DawoudSheraz: We can use `allowed_tags` from the lti_consumer/lti_consumer.py file instead of writing all the tags again. Also, since we made the changes for img tag(and the allowed attributes), so the test should be specific to those changes as well.

AhtishamShahid: `allowed_tags` cannot be reuses because it is scoped to `function` only and I have added test for `img` src.

lti_consumer/lti_consumer.py
```diff
@@ -925,15 +926,37 @@ def _get_context_for_template(self):
...
+            'ol',
+            'strong',
+            'ul',
+            'img'
+        ]
+        allowed_attributes = {
+            'a': ['href', 'title'],
+            'abbr': ['title'],
+            'acronym': ['title'],
+            'img': ['src']
```
DawoudSheraz: We won't be able to use `alt` attribute. Not sure of the overall usage of this xblock but it can introduce a11y problems.

lti_consumer/tests/unit/test_lti_consumer.py
```diff
@@ -825,6 +827,27 @@ def test_context_keys(self):
...
+    def test_comment_retains_image_src(self):
+        """
+        Test that image tag has src and other attrs are sanitized
+        """
+        comment = u'<img src="example.com/image.jpeg" onerror="myFunction()">'
+        self.xblock.set_user_module_score(Mock(), 0.92, 1.0, comment)
+        context = self.xblock._get_context_for_template()  # pylint: disable=protected-access
+
+        self.assertIn(u'<img src="example.com/image.jpeg">', context['comment'])
+
```
DawoudSheraz: Add another attribute that isn't allowed and test it as well.

lti_consumer/lti_consumer.py
```diff
@@ -695,7 +695,8 @@ def prefixed_custom_parameters(self):
                 except ValueError:
                     _ = self.runtime.service(self, "i18n").ugettext
                     # pylint: disable=line-too-long
-                    msg = self.ugettext('Could not parse custom parameter: {custom_parameter}. Should be "x=y" string.').\
+                    msg = self.ugettext(
+                        'Could not parse custom parameter: {custom_parameter}. Should be "x=y" string.'). \
```
awaisdar001: Please check your linebreaks. 
```
msg = self.ugettext(
    'Could not parse custom parameter: {custom_parameter}. Should be "x=y" string.')
    .format(custom_parameter="{0!r}".format(custom_parameter)
)
```

lti_consumer/lti_consumer.py
```diff
@@ -925,15 +926,37 @@ def _get_context_for_template(self):
...
         # ALLOWED_ATTRIBUTES are
         #     'a': ['href', 'title'],
         #     'abbr': ['title'],
         #     'acronym': ['title'],
-        #
+        #     'img': ['src']
+
         # This lets all plaintext through.
-        sanitized_comment = bleach.clean(self.score_comment)
+        allowed_tags = [
```
awaisdar001: Do we have a product opinion on this one? that we are adding this limitation on allowed tags?

AhtishamShahid: Only new `` tag is added in `allowed_tags` with other tags which were by default allowed by `bleach` and no tag is removed from allowed so it should work as it was previously working + `img`

lti_consumer/lti_consumer.py
```diff
@@ -574,8 +574,8 @@ def lti_provider_key_secret(self):
             try:
                 lti_id, key, secret = [i.strip() for i in lti_passport.split(':')]
             except ValueError:
-                msg = self.ugettext('Could not parse LTI passport: {lti_passport}. Should be "id:key:secret" string.').\
-                    format(lti_passport='{0!r}'.format(lti_passport))
+                msg = 'Could not parse LTI passport: {lti_passport}. Should be "id:key:secret" string.'
+                msg = self.ugettext(msg).format(lti_passport='{0!r}'.format(lti_passport))
```
awaisdar001: Why can't the following could help? 2 formats doesn't look a right way to do this. 
```
msg = 'Could not parse LTI passport: {lti_passport!r}. Should be "id:key:secret" string'.format(lti_passport=lti_passport)
```

lti_consumer/lti_consumer.py
```diff
@@ -925,15 +924,18 @@ def _get_context_for_template(self):
 
         # use bleach defaults. see https://github.com/jsocol/bleach/blob/master/bleach/__init__.py
         # ALLOWED_TAGS are
-        # ['a', 'abbr', 'acronym', 'b', 'blockquote', 'code', 'em', 'i', 'li', 'ol',  'strong', 'ul']
```
awaisdar001: Comments like this are hard to maintain, Instead, The URL above is enough and will provide updated information. 

I would suggest instead of hardcoding these tags and attributes, just have a direct URL that points to documentation or in GitHub ( whichever makes more sense).

PR #58, opened 2020-04-13
[BD-2] BB-2250: LTI 1.3 base implementation

giovannicimolin:
This PR implements a proof-of-concept implementation of the LTI 1.3 standard to the current LTI Block without breaking LTI 1.1 support.

**JIRA tickets**: [TNL-7133 (epic)](https://openedx.atlassian.net/browse/TNL-7133), [TNL-7183](https://openedx.atlassian.net/browse/TNL-7183).

**Sandbox:** TBD (issues provisioning master sandbox).

**Discussions**: 
- Slack: https://openedx.slack.com/archives/CVCD5LACC

**Setup instructions:**
1. Clone this branch on your devstack's `src` folder 
2. Install in LMS and Studio 
```
$ lms-shell
> pip install -e /edx/src/xblock-lti-consumer
```
3. Set up a local tunnel (using `ngrok` or a similar tool) 
4. Set up a tool in the IMS Global reference implementation (https://lti-ri.imsglobal.org/lti/tools/).

**Testing instructions:**
1. Create a new course, and add the `lti_consumer` block to the advanced modules list.
2. In the course, create a new unit and add the LTI block.
3. In studio, you'll see a few parameters being displayed in the preview:
* **Client**: 335dde47-0bc3-4dc8-8faa-c48b38b1539e
* **Deployment ID**: 1
* **Keyset URL**: /api/lti_consumer/v1/public_keysets/block-v1:OpenCraft+LTI101+2020_T2+type@lti_consumer+block@996c72b16070434098bc598bd7d6dbde
* **OIDC Callback URL**: /api/lti_consumer/v1/launch/
4. Add the tunnel url to the keyset url as it'll need to be accessed by the tool.
```
# This is <LMS_URL>/api/lti_consumer/v1/launch/<BLOCK_LOCATION>
https://647dd2e1.ngrok.io/api/lti_consumer/v1/public_keysets/block-v1:OpenCraft+LTI101+2020_T2+type@lti_consumer+block@996c72b16070434098bc598bd7d6dbde
```
5. Append the callback url to the LMS address.
```
# The block information comes from the OIDC login hint now :)
http://localhost:18000/api/lti_consumer/v1/launch/
```
6. Edit the tool you previously created and add the parameters above.
6. Publish block, go into LMS and launch LTI following the instructions on the XBlock.
8. Check that the LTI launch was successful.

**Notes:**
You can use this test tool configuration as a reference: https://lti-ri.imsglobal.org/lti/tools/857.

**Reviewers:**
- [ ] @viadanna
- [ ] @davidjoy 
- [ ] @davestgermain

### DIFF
--- a/lti_consumer/lti_1p3/consumer.py
+++ b/lti_consumer/lti_1p3/consumer.py
@@ -1,0 +1,275 @@
+"""
+LTI 1.3 Consumer implementation
+"""
+import json
+import time
+
+# Quality checks failing due to know pylint bug
+# pylint: disable=relative-import
+from six.moves.urllib.parse import urlencode
+
+from Crypto.PublicKey import RSA
+from jwkest.jwk import RSAKey
+from jwkest.jws import JWS
+from jwkest import jwk
+
+from .constants import LTI_1P3_ROLE_MAP, LTI_BASE_MESSAGE
+
+
+class LtiConsumer1p3(object):
+    """
+    LTI 1.3 Consumer Implementation
+    """
+    def __init__(
+            self,
+            iss,
+            lti_oidc_url,
+            lti_launch_url,
+            client_id,
+            deployment_id,
+            rsa_key,
+            rsa_key_id
+    ):
+        """
+        Initialize LTI 1.3 Consumer class
+        """
+        self.iss = iss
+        self.oidc_url = lti_oidc_url
+        self.launch_url = lti_launch_url
+        self.client_id = client_id
+        self.deployment_id = deployment_id
+
+        # Generate JWK from RSA key
+        self.jwk = RSAKey(
+            # Using the same key ID as client id
+            # This way we can easily serve multiple public
+            # keys on teh same endpoint and keep all
+            # LTI 1.3 blocks working
+            kid=rsa_key_id,
+            key=RSA.import_key(rsa_key)
+        )
+
+        # IMS LTI Claim data
+        self.lti_claim_user_data = None
+        self.lti_claim_launch_presentation = None
+        self.lti_claim_custom_parameters = None
+
+    def _encode_and_sign(self, message):
+        """
+        Encode and sign JSON with RSA key
+        """
+        # The class instance that sets up the signing operation
+        # An RS 256 key is required for LTI 1.3
+        _jws = JWS(message, alg="RS256", cty="JWT")
+
+        # Encode and sign LTI message
+        return _jws.sign_compact([self.jwk])
+
+    @staticmethod
+    def _get_user_roles(role):
+        """
+        Converts platform role into LTI compliant roles
+
+        Used in roles claim: should return array of URI values
+        for roles that the user has within the message's context.
+
+        Supported roles:
+        * Core - Administrator
+        * Institution - Instructor (non-core role)
+        * Institution - Student
+
+        Reference: http://www.imsglobal.org/spec/lti/v1p3/#roles-claim
+        Role vocabularies: http://www.imsglobal.org/spec/lti/v1p3/#role-vocabularies
+        """
+        lti_user_roles = set()
+
+        if role:
+            # Raise value error if value doesn't exist in map
+            if role not in LTI_1P3_ROLE_MAP:
+                raise ValueError("Invalid role list provided.")
+
+            # Add roles to list
+            lti_user_roles.update(LTI_1P3_ROLE_MAP[role])
+
+        return list(lti_user_roles)
+
+    def prepare_preflight_url(
+            self,
+            callback_url,
+            hint="hint",
+            lti_hint="lti_hint"
+    ):
+        """
+        Generates OIDC url with parameters
+        """
+        oidc_url = self.oidc_url + "?"
+        parameters = {
+            "iss": self.iss,
+            "client_id": self.client_id,
+            "lti_deployment_id": self.deployment_id,
+            "target_link_uri": callback_url,
+            "login_hint": hint,
+            "lti_message_hint": lti_hint
+        }
+
+        return {
+            "oidc_url": oidc_url + urlencode(parameters),
+        }
+
+    def set_user_data(
+            self,
+            user_id,
+            role,
+            full_name=None,
+            email_address=None
+    ):
+        """
+        Set user data/roles and convert to IMS Specification
+
+        User Claim doc: http://www.imsglobal.org/spec/lti/v1p3/#user-identity-claims
+        Roles Claim doc: http://www.imsglobal.org/spec/lti/v1p3/#roles-claim
+        """
+        self.lti_claim_user_data = {
+            # User identity claims
+            # sub: locally stable identifier for user that initiated the launch
+            "sub": user_id,
+
+            # Roles claim
+            # Array of URI values for roles that the user has within the message's context
+            "https://purl.imsglobal.org/spec/lti/claim/roles": self._get_user_roles(role)
+        }
+
+        # Additonal user identity claims
+        # Optional user data that can be sent to the tool, if the block is configured to do so
+        if full_name:
+            self.lti_claim_user_data.update({
+                "name": full_name,
+            })
+
+        if email_address:
+            self.lti_claim_user_data.update({
+                "email": email_address,
+            })
+
+    def set_launch_presentation_claim(
+            self,
+            document_target="iframe"
+    ):
+        """
+        Optional: Set launch presentation claims
+
+        http://www.imsglobal.org/spec/lti/v1p3/#launch-presentation-claim
+        """
+        if document_target not in ['iframe', 'frame', 'window']:
+            raise ValueError("Invalid launch presentation format.")
+
+        self.lti_claim_launch_presentation = {
+            # Launch presentation claim
+            "https://purl.imsglobal.org/spec/lti/claim/launch_presentation": {
+                # Can be one of: iframe, frame, window
+                "document_target": document_target,
+                # TODO: Add support for `return_url` handler to allow the tool
+                # to return error messages back to the lms.
+                # See the spec referenced above for more information.
+            },
+        }
+
+    def set_custom_parameters(
+            self,
+            custom_parameters
+    ):
+        """
+        Stores custom parameters configured for LTI launch
+        """
+        if not isinstance(custom_parameters, dict):
+            raise ValueError("Custom parameters must be a key/value dictionary.")
+
+        self.lti_claim_custom_parameters = {
+            "https://purl.imsglobal.org/spec/lti/claim/custom": custom_parameters
+        }
+
+    def generate_launch_request(
+            self,
+            preflight_response,
+            resource_link
+    ):
+        """
+        Build LTI message from class parameters
+
+        This will add all required parameters from the LTI 1.3 spec and any additional ones set in
+        the configuration and JTW encode the message using the provided key.
+        """
+        # Start from base message
+        lti_message = LTI_BASE_MESSAGE.copy()
+
+        # TODO: Validate preflight response
+        # Add base parameters
+        lti_message.update({
+            # Issuer
+            "iss": self.iss,
+
+            # Nonce from OIDC preflight launch request
+            "nonce": preflight_response.get("nonce"),
+
+            # JWT aud and azp
+            "aud": [
+                self.client_id
+            ],
+            "azp": self.client_id,
+
+            # LTI Deployment ID Claim:
+            # String that identifies the platform-tool integration governing the message
+            # http://www.imsglobal.org/spec/lti/v1p3/#lti-deployment-id-claim
+            "https://purl.imsglobal.org/spec/lti/claim/deployment_id": self.deployment_id,
+
+            # Target Link URI: actual endpoint for the LTI resource to display
+            # MUST be the same value as the target_link_uri passed by the platform in the OIDC login request
+            # http://www.imsglobal.org/spec/lti/v1p3/#target-link-uri
+            "https://purl.imsglobal.org/spec/lti/claim/target_link_uri": self.launch_url,
+
+            # Resource link: stable and unique to each deployment_id
+            # This value MUST change if the link is copied or exported from one system or
+            # context and imported into another system or context
+            # http://www.imsglobal.org/spec/lti/v1p3/#resource-link-claim
+            "https://purl.imsglobal.org/spec/lti/claim/resource_link": {
+                "id": resource_link,
+                # Optional claims
+                # "title": "Introduction Assignment"
+                # "description": "Assignment to introduce who you are",
+            },
+        })
+
+        # Check if user data is set, then append it to lti message
+        # Raise if isn't set, since some user data is required for the launch
+        if self.lti_claim_user_data:
+            lti_message.update(self.lti_claim_user_data)
+        else:
+            raise ValueError("Required user data isn't set.")
+
+        # Set optional claims
+        # Launch presentation claim
+        if self.lti_claim_launch_presentation:
+            lti_message.update(self.lti_claim_launch_presentation)
+
+        # Custom variables claim
+        if self.lti_claim_custom_parameters:
+            lti_message.update(self.lti_claim_custom_parameters)
+
+        # Add `exp` and `iat` JWT attributes
+        lti_message.update({
+            "iat": int(round(time.time())),
+            "exp": int(round(time.time()) + 3600)
+        })
+
+        return {
+            "state": preflight_response.get("state"),
+            "id_token": self._encode_and_sign(lti_message)
+        }
+
+    def get_public_keyset(self):
+        """
+        Export Public JWK
+        """
+        public_keys = jwk.KEYS()
+        public_keys.append(self.jwk)
+        return json.loads(public_keys.dump_jwks())
--- a/lti_consumer/lti_consumer.py
+++ b/lti_consumer/lti_consumer.py
@@ -54,6 +54,7 @@ from __future__ import absolute_import, unicode_literals
 
 import logging
 import re
+import uuid
 from collections import namedtuple
 from importlib import import_module
 
@@ -61,6 +62,7 @@ import six.moves.urllib.error
 import six.moves.urllib.parse
 import six
 import bleach
+from Crypto.PublicKey import RSA
 from django.utils import timezone
 from webob import Response
 from xblock.core import List, Scope, String, XBlock
@@ -72,9 +74,16 @@ from xblockutils.studio_editable import StudioEditableXBlockMixin
 
 from .exceptions import LtiError
 from .lti import LtiConsumer
+from .lti_1p3.consumer import LtiConsumer1p3
 from .oauth import log_authorization_header
 from .outcomes import OutcomeService
-from .utils import _
+from .utils import (
+    _,
+    get_lms_base,
+    get_lms_lti_keyset_link,
+    get_lms_lti_launch_link,
+)
+
 
 log = logging.getLogger(__name__)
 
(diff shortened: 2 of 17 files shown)


edx-webhook:
Thanks for the pull request, @giovannicimolin! I've created [OSPR-4380](https://openedx.atlassian.net/browse/OSPR-4380) to keep track of it in JIRA. JIRA is a place for product owners to prioritize feature reviews by the engineering development teams.

- supporting documentation
- edx-code email threads

All technical communication about the code itself will still be done via the GitHub pull request interface. As a reminder, [our process documentation is here](http://edx-developer-guide.readthedocs.org/en/latest/process/overview.html).

lti_consumer/utils.py
```diff
@@ -1,11 +1,48 @@
...
+def get_lms_base():
+    """
+    Returns LMS base url to be used as issuer on OAuth2 flows
+
+    TODO: This needs to be improved and account for Open edX sites and
+    organizations.
+    One possible improvement is to use `contentstore.get_lms_link_for_item`
+    and strip the base domain name.
+    """
+    return settings.LMS_ROOT_URL
```
giovannicimolin, line 25: This might not be the case for urls fetched from `SiteConfiguration`, but I don't think it works from this context.

natabene:
@giovannicimolin Thank you for your contribution. Please let me know once it is ready for our review.

giovannicimolin:
@natabene This is from a blended development project (BD-2).
I've linked the relevant TNL tickets and edX reviewers in the PR's description.

natabene:
@giovannicimolin Got it, thanks. Can you please include [BD-2] in title of all PRs you will be contributing for this?

marcotuts:
Thanks Giovanni for this! I have updated the prefix to be "[BD-2]" to help us quickly identify these OSPR tickets in JIRA. Thanks!

lti_consumer/lti_consumer.py
```diff
@@ -270,6 +280,48 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
...
+        scope=Scope.settings
+    )
+    # Client ID and block key
+    lti_1p3_client_id = String(
+        display_name=_("LTI 1.3 Block Client ID"),
+        default=str(uuid.uuid4()),
+        scope=Scope.settings
+    )
+    # This key isn't editable, and should be regenerated
+    # for every block created (and not be carried over)
```
viadanna, line 316: You mean for each user or each block instance in the course?

Maybe `user_info` would fit here, otherwise `settings` seems fine. [ref](https://edx.readthedocs.io/projects/xblock-tutorial/en/latest/concepts/fields.html#predefined-scopes)

lti_consumer/lti_consumer.py
```diff
@@ -558,6 +617,23 @@ def role(self):
         """
         return ROLE_MAP.get(self.runtime.get_user_role(), u'Student')
 
+    @property
+    def user_roles(self):
+        """
+        Get system user role and convert it to LTI 1.3 user roles.
```
viadanna, line 623: Seems the convert step is being performed somewhere else.

lti_consumer/lti_consumer.py
```diff
@@ -558,6 +617,23 @@ def role(self):
...
+        """
+        roles = ['student']
+
+        if self.runtime.get_user_role() == 'instructor':
+            # Instructor have both instructor and administrator
+            # permissions on LTI tool.
```
viadanna, line 629: Is this still the case? I thought we were aiming for a 1:1 mapping of roles.

lti_consumer/lti_consumer.py
```diff
@@ -740,6 +830,85 @@ def student_view(self, context):
...
+        if self.lti_version != "lti_1p3":
+            return self.student_view(context)
+
+        fragment = Fragment()
+        loader = ResourceLoader(__name__)
+        context = {
+            "client": self.lti_1p3_client_id,
+            "deployment_id": "1",
+            "keyset_url": self.keyset_url,
+            "oidc_callback": self.consumer_launch_url,
```
viadanna, line 858: Why not use `get_lms_lti_launch_link` directly?

setup.py
```diff
@@ -44,6 +44,9 @@ def package_data(pkg, roots):
     entry_points={
         'xblock.v1': [
             'lti_consumer = lti_consumer:LtiConsumerXBlock',
+        ],
```
viadanna, line 47: Just a reminder to bump the version here 🙂

lti_consumer/lti_1p3/consumer.py
```diff
@@ -0,0 +1,275 @@
...
+        # IMS LTI Claim data
+        self.lti_claim_user_data = None
+        self.lti_claim_launch_presentation = None
+        self.lti_claim_custom_parameters = None
+
+    def _encode_and_sign(self, message):
+        """
+        Encode and sign JSON with RSA key
+        """
+        msg = message.copy()
```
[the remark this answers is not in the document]
giovannicimolin: Not really, since the message won't be used until later, but I should probably move setting `iat` and `exp` to inside the LTI message generation.

lti_consumer/lti_1p3/consumer.py
```diff
@@ -0,0 +1,275 @@
...
+        """
+        Build LTI message from class parameters
+
+        This will add all required parameters from the LTI 1.3 spec and any additional ones set in
+        the configuration and JTW encode the message using the provided key.
+        """
+        # Start from base message
+        lti_message = LTI_BASE_MESSAGE.copy()
+
+        # TODO: Validate preflight response
```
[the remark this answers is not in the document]
giovannicimolin: Here I mean to validate if the preflight request response has all the required parameters to make the LTI launch.
It works like this:
1. XBlock generates and makes a request to preflight url with parameters.
2. The tool will check the request and make a GET request witn the preflight response back to the LMS.
3. We take that preflight response and craft the `lti_message`, but we need to check if we have all the required keys (`nonce` and `state`).

lti_consumer/lti_consumer.py
giovannicimolin: When using `settings` as scope, if you duplicate the block, the key is also copied over.
I'm looking for some way to change this when copying the XBlock.

lti_consumer/lti_consumer.py
giovannicimolin: Yeah, this can definitively be improved. I think I'll just pass the Django user role into the library.

lti_consumer/lti_consumer.py
giovannicimolin: I don't think a 1:1 role mapping is a good solution here.

If we do so, course creators (`staff`) won't have admin access to a LTI Tool and won't be able to manage users/moderate content inside LTI. This is the same reasoning why the LTI 1.1 implementation does the mapping like this.

lti_consumer/lti_1p3/consumer.py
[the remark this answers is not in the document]
viadanna: Oh, ok, I thought you meant the HTTP response.

lti_consumer/lti_consumer.py
viadanna: `Scope.settings` has `Block usage` scope which I understand as independent for each instance in a course. Am I missing anything here?

viadanna:
@giovannicimolin Took me long enough but I tested this working correctly 👍

giovannicimolin:
@viadanna Thanks for the review and testing! I've addressed most comments and improved test coverage.

@davidjoy @davestgermain This is ready for review. CC @natabene

setup.py
giovannicimolin: I went for a big version bump since we're changing a lot of the internals and adding support for a new spec.

lti_consumer/lti_consumer.py
```diff
@@ -756,14 +894,71 @@ def lti_launch_handler(self, request, suffix=''):  # pylint: disable=unused-argu
...
-        context = self._get_context_for_template()
-        context.update({'lti_parameters': lti_parameters})
-        template = loader.render_mako_template('/templates/html/lti_launch.html', context)
+        context = {}
+
+        lti_consumer = self._get_lti1p3_consumer()
+
+        # Pass user data
+        lti_consumer.set_user_data(
+            user_id=self.runtime.user_id,
```
giovannicimolin, line 922: @davidjoy This is waiting on https://github.com/edx/edx-platform/pull/23665.

lti_consumer/lti_consumer.py
```diff
@@ -740,6 +799,85 @@ def student_view(self, context):
...
+            xblock.fragment.Fragment: XBlock HTML fragment
+        """
+        if self.lti_version != "lti_1p3":
+            return self.student_view(context)
+
+        fragment = Fragment()
+        loader = ResourceLoader(__name__)
+        context = {
+            "client": self.lti_1p3_client_id,
+            "deployment_id": "1",
```
davidjoy, line 825: Does this deployment_id need to be replaced with a "real" value instead of `"1"`?

davidjoy:
@giovannicimolin I've taken a look through the PR and nothing jumped out at me except that one comment above.  I'd like to spend some time early next week testing it/seeing how it all works.  Will ignore the studio experience for the moment, as you suggested!

lti_consumer/lti_consumer.py
giovannicimolin: @davidjoy The current implementation uses `Single tenant` use case described in the core specification here: http://www.imsglobal.org/spec/lti/v1p3/#single-tenant-tool-registered-and-deployed-once.

deployments_onetoone_bw

So the deployment ID is one for all deployments, but this would be different if the XBlocks shared the same key and client ID, which shouldn't happen.

giovannicimolin:
@davidjoy Thanks for taking the time to review this. I've replied to your comment.
Let me know if you want me to change anything else.

lti_consumer/lti_consumer.py
```diff
@@ -740,6 +799,85 @@ def student_view(self, context):
...
+        configuration of the instance of this XBlock
+
+        Arguments:
+            context (dict): XBlock context
+
+        Returns:
+            xblock.fragment.Fragment: XBlock HTML fragment
+        """
+        if self.lti_version != "lti_1p3":
+            return self.student_view(context)
```
davidjoy, line 819: Just trying to recall from our conversations - why is it we need a new author_view but didn't need one for v1.1?  Is it because we're displaying a preview of the tool in studio, which we didn't do before?

lti_consumer/lti_consumer.py
```diff
@@ -270,6 +280,48 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
...
+
+    # LTI 1.3 fields
+    lti_version = String(
+        display_name=_("LTI Version"),
+        scope=Scope.settings,
+        values=[
+            {"display_name": "LTI 1.1/1.2", "value": "lti_1p1"},
+            {"display_name": "LTI 1.3", "value": "lti_1p3"},
+        ],
+        default="lti_1p1",
```
davidjoy, line 292: So just to be totally pedantic here - when this xblock's code is updated underneath an existing LTI 1.1 XBlock, this lti_version field will not have a value set, and will therefore default to `lti_1p1` for any/all existing LTI XBlocks in the system?  I.e., just sanity checking that there's a clean, seamless upgrade path for existing LTI 1.1 content.

lti_consumer/lti_consumer.py
```diff
@@ -740,6 +799,85 @@ def student_view(self, context):
...
+        fragment.initialize_js('LtiConsumerXBlock')
+        return fragment
+
+    def _get_lti1p3_consumer(self):
+        """
+        Returns LTI 1.3 Consumer class
+
+        Instanced from parameters from XBlock
+        """
+        return LtiConsumer1p3(
```
davidjoy, line 842: Does it matter that this is creating a new instance of `LtiConsumer1p3` each time `_get_lti1p3_consumer` is called?  Naively, I might have expected it was returning an instance variable.  I don't know that it gets called more than once for a request, but if so, are there any consequences for having separate instances?  LtiConsumer1p3 seems to have instance variables, but I'm not sure if they ever change from what it's instantiated with, so it may (technically) be fine.  Feels slightly odd, though.

lti_consumer/lti_consumer.py
```diff
@@ -740,6 +799,85 @@ def student_view(self, context):
...
+            "oidc_callback": get_lms_lti_launch_link(),
+            "launch_url": self.lti_1p3_launch_url,
+        }
+        fragment.add_content(loader.render_mako_template('/templates/html/lti_1p3_studio.html', context))
+        fragment.add_css(loader.load_unicode('static/css/student.css'))
+        fragment.add_javascript(loader.load_unicode('static/js/xblock_lti_consumer.js'))
+        fragment.initialize_js('LtiConsumerXBlock')
+        return fragment
+
+    def _get_lti1p3_consumer(self):
```
davidjoy, line 836: So from an architectural point of view, I'm thinking about the roles and responsibilities of this `LtiConsumerXBlock` class vs. `LtiConsumer1p3`.  What code belongs in `LtiConsumerXBlock` vs being pushed into `LtiConsumer1p3`?  And likewise, if we need a delegate for handling 1.3 functionality, should we also have an `LtiConsumer1p1` where we put some of the existing code?

Purely as an example, `lti_launch_callback` appears to be solely for 1.3 - should `LtiConsumerXBlock` delegate to `LtiConsumer1p3` for fulfillment of the launch callback handling?  Meaning that `lti_launch_callback` would be a very short method that either returns a 404 or delegates to `LtiConsumer1p3` and returns its response?

Similarly, should `_launch_lti_1p3` and `_launch_lti_1p1` live in `LtiConsumer1p3` and `LtiConsumer1p1` respectively?

Just throwing out examples to try to illustrate what I'm thinking, I'm not trying to make pronouncements on where to put things, so please don't start refactoring unless I'm making sense!

Curious what you think/how you made decisions on where to put some of these things.  I imagine it may have to do with whether or not you need particular instance variables in `LtiConsumerXBlock`...

lti_consumer/lti_consumer.py
giovannicimolin: No, you're right!
But, there's 1 issue that I'm hitting: sometimes the keys from Studio and LMS differ. I'll investigate this on a follow up ticket.

lti_consumer/lti_consumer.py
giovannicimolin: So the `author_view` here is being used to display a few variables that the user need to set up in the LTI Tool in order for the launch to work.
Furthermore, launches can't be performed from the Studio since the tool will need to access LMS URLs.

This is the Studio view with this change when using LTI 1.3: 
image

The LTI 1.1 works as usual and displays the launch preview.

lti_consumer/lti_consumer.py
davestgermain: Since you're saying `default=str(uuid.uuid4())`, and this is on a class variable, the default value will be fixed _when the code is imported_. If you want a unique default for every instance, this will have to be a function call. I don't know if XBlock fields support that feature. 
The same applies to your block key below.

lti_consumer/lti_consumer.py
giovannicimolin: I've split this function in order to avoid code duplication in the multiple places that it needs to be used. This LTI consumer class is stateless, and doesn't store any request/launch data. It's main function is to receive LTI related variables and generate an output from them.

Do you have an idea of how we can improve this? 
Do you think a property will do better here?

>I don't know that it gets called more than once for a request, but if so, are there any consequences for having separate instances?

We shouldn't see any consequences of this as long as the classes are instantiated in the same way.
I have a small worry about instancing and storing this class since the user might change some parameter in the settings and the saved class might end up with old information (this issue will only happen in the Studio, so it shouldn't cause any trouble, but it's worth mentioning).

lti_consumer/lti_consumer.py
giovannicimolin: @davestgermain You're right! That's precisely the issue.
I'll check if the block field API is compatible with this.

lti_consumer/lti_consumer.py
```diff
@@ -756,14 +894,71 @@ def lti_launch_handler(self, request, suffix=''):  # pylint: disable=unused-argu
         Returns:
             webob.response: HTML LTI launch form
         """
-        lti_consumer = LtiConsumer(self)
-        lti_parameters = lti_consumer.get_signed_lti_parameters()
+        if self.lti_version == "lti_1p3":
```
davestgermain, line 897: I'm trying to follow where this handler will be called for 1.3 requests. Is it possible to have a different handler for 1.3, like `lti_launch_handler_1p3`, so that you don't have to do this switch, and keep `lti_launch_handler` for 1.1 requests? It's only a minor criticism.

lti_consumer/lti_consumer.py
davidjoy: If it's stateless, I'm with you. 👍

giovannicimolin: @davidjoy As we discussed in the meeting today: 
It isn't exactly a stateless class, since it has a few methods to set user and configuration variables, but it should be used as one on most cases (by instancing a new class, setting variables and running desired LTI method).

I'll add some comments on the code to clarify how this class should be used.

I ended up using some internal variables to reduce the amount of back and forth and amount of variables that needed to be passed on each function call.

lti_consumer/lti_consumer.py
giovannicimolin: @davidjoy Yes, that's exactly what happens.
I've tested it locally like this:
1. Installed master version of LTI XBlock
2. Created a new block using this test tool:  `https://lti.tools/saltire/tp` .
3. Checked LMS and Studio that the launch worked.
4. Installed the new block version, checked that it was set to use the 1.1 version on the previously created block.
5. Checked that the launch worked on both LMS & Studio.

lti_consumer/lti_consumer.py
giovannicimolin: Sure, I can do this. I've just use this switch to avoid changing the student view, but it makes much more sense to change the student view than this one. 
I'll adjust the PR and push a fix for this.

giovannicimolin: @davestgermain Reverted to the old method and created a new handler for the LTI 1.3 launch. 
Does it look better this way?

lti_consumer/lti_consumer.py
giovannicimolin: I've mentioned in the meeting, but for the record:
We've decided to make the LTI 1.3 implementation using different patterns than the ones used by the old implementation.

I've implemented the LTI 1.3 consumer to only handle LTI related stuff, so the endpoint implementation and HTTP  handling are not in it's scope. This will make the module to be easily reusable in other parts of the LMS/Studio, as it is independent of the actual request handlers (example: XBlocks use `WebOb` objects, django uses `HttpReponse` objects, and so on, but the LTI module doesn't need to know that).

giovannicimolin:
@davidjoy Can you merge the PR? I don't have merge permissions for this repo.
CC @natabene

edx-webhook:
@giovannicimolin 🎉 Your pull request was merged!

Please take a moment to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=edx/xblock-lti-consumer&entry.867055334=https://github.com/edx/xblock-lti-consumer/pull/58&entry.1484655318=https://github.com/giovannicimolin&entry.752974735=2020-04-13+19:46&entry.1917517419=2020-04-30+14:35&entry.2133058324=Yes) so we can improve your experience in the future.

PR #85, opened 2020-07-01
[BD-2] [BB-2612]: Add ADR to define LTI Plugin architecture implementation

giovannicimolin:
This PR adds an ADR to define the software architecture of the LTI consumer implementation in the platform.
The proposed approach, if implemented, will allow configuring and performing LTI launches anywhere in the platform, and have a consistent configuration storage for LTI related settings.

I've tried to follow the conventions from [this file](https://github.com/edx/frontend-app-learning/blob/master/docs/decisions/0001-record-architecture-decisions.rst) and linked documents. 

Feel free to leave comments on content and format of the ADR.

**Jira tickets:**
 - https://openedx.atlassian.net/browse/TNL-7315

**Reviewers:**
- [x] @xitij2000
- [x] @davidjoy 
- [ ] @davestgermain 
- [ ] @nasthagiri

### DIFF
--- a/docs/decisions/0001-lti-extensions-plugin.rst
+++ b/docs/decisions/0001-lti-extensions-plugin.rst
@@ -1,0 +1,115 @@
+Platform wide LTI support
+----------------------------
+
+Status
+======
+
+Accepted
+
+Context
+=======
+
+The LTI Consumer implementation in the Open edX platform is currently limited to LTI launches inside XBlocks.
+
+The goal of this ADR is to:
+
+#. Define an architecture to enable LTI launch and configuration for multiple contexts (from block level to platform level).
+#. Make LTI configuration reusable throughout a context.
+#. Support both LTI 1.1 and LTI 1.3, as well as LTI Advantage.
+
+Decisions
+=========
+
+Develop LTI related code to be reusable in the platform context, and allow LTI to be configured for different contexts in the platform.
+This requires adding a separation layer between the framework presenting the content and the LTI logic, configuration and storage layer.
+
+The first steps were already taken to move towards this implementation with these PRs:
+
+`Merge LTI 1.3 support to master`_: Implemented support for LTI 1.3 by creating a framework agnostic python module that handles the
+LTI specific logic (such as launch message creation, token verification and OIDC flows) and passing responses as JSON objects to be
+interpreted and presented by the XBlock LTI consumer.
+The module is easily reusable in contexts outside the XBlock and all the code in `lti_consumer/lti_1p3` don't rely on XBlock or
+Django specific features.
+The PR uses the XBlock's fields to store LTI configuration.
+
+`Add support for LTI embeds in course tabs and elsewhere`_: This is a partial implementation of logic separation for the LTI 1.1
+consumer to allow LTI launches from outside the context of an XBlock, following the guidelines set by the LTI 1.3 implementation.
+While this is a smaller step than the implementation above, it allows LTI 1.1 launches on any LMS/Studio view, but doesn't implement
+the configuration storage.
+
+.. _`Merge LTI 1.3 support to master`: https://github.com/edx/xblock-lti-consumer/pull/82
+.. _`Add support for LTI embeds in course tabs and elsewhere`: https://github.com/edx/xblock-lti-consumer/pull/77
+
+
+A few actions that need to be taken to achieve the desired architecture are outlined below:
+
+Centralized configuration for LTI launches
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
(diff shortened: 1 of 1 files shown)


edx-webhook:
Thanks for the pull request, @giovannicimolin! I've created [BLENDED-320](https://openedx.atlassian.net/browse/BLENDED-320) to keep track of it in Jira.  More details are on the [BD-02](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1544913089/BD-02+LTI+v1.3+Consumer+XBlock+Support) project page.

docs/decisions/0001-lti-extensions-plugin.rst
```diff
@@ -0,0 +1,116 @@
...
+across all platform's contexts.
+
+Change folder structure and rename repository
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To better reflect the new repo usage, rename the repository to "openedx-lti-consumer" and
+change the folder structure.
+
+Here is the new proposed folder structure:
+```
```
bradenmacdonald, line 74: There is a formatting error here - this is the markdown syntax for a code block, not rST syntax, so it's not working - see the GitHub preview.

docs/decisions/0001-lti-extensions-plugin.rst
```diff
@@ -0,0 +1,116 @@
...
+
+The transition phase can allow configuring LTI integration both locally (on the XBlock) and using the
+generic configuration.
+
+
+Consequences
+============
+
+* LTI configuration will be independent from courses, but hooks can be used to
+  import/export data to the LTI configuration store.
```
bradenmacdonald, line 112: Is this saying that when one imports/exports a course, the LTI configuration will import/export with it? I'm not sure if that's a good thing or a bad thing; for sharing courses within an org/instance, it makes sense, but for sharing courses across orgs it'd be better to omit the LTI credentials.

giovannicimolin: Yeah, that makes more sense from a security perspective.
I think we can omit the LTI credentials and display validation errors in the blocks and pages using it.

In a course context, we can display errors in the blocks and also on the LTI configuration UI (something like: `your course is using **configuration-123** but that isn't set up`).

docs/decisions/0001-lti-extensions-plugin.rst
```diff
@@ -0,0 +1,115 @@
...
+.. _`Merge LTI 1.3 support to master`: https://github.com/edx/xblock-lti-consumer/pull/82
+.. _`Add support for LTI embeds in course tabs and elsewhere`: https://github.com/edx/xblock-lti-consumer/pull/77
+
+
+A few actions that need to be taken to achieve the desired architecture are outlined below:
+
+Centralized configuration for LTI launches
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Use Django models to store LTI related configuration and launch data for all LTI integrations.
```
davidjoy, line 49: I'm assuming so, but would this be used for existing LTI XBlocks as well?  i.e., when you say _all_ do you really mean _all_? 😄

davidjoy: Ahh - I see this is covered below in "Tech Debt" 👍

davidjoy:
Thanks for writing this, @giovannicimolin.  Makes a lot of sense to me, and seems very clearly broken out.  Very nice.  If you're good with it, I think we can feel free to merge now.  Ready!?

docs/decisions/0001-lti-extensions-plugin.rst
giovannicimolin: >I'm assuming so, but would this be used for existing LTI XBlocks as well? 

For new XBlocks, we can give the user the option to opt for the new storage or keep using the old one. And then, over the next few releases, deprecate and disable the old configuration method.

giovannicimolin:
@davidjoy Thanks for the review! :)

I'm good with the current state of the ADR, should we change the status from `provisional` to `Accepted` before we merge?

docs/decisions/0001-lti-extensions-plugin.rst
xitij2000: I think the old method can be marked as deprecated already, with clear links to documentation or the new UI to add these settings. If we already know it's on the way out, marking it as deprecated sooner rather than later is better right? If need be it can just be kept around for longer.

giovannicimolin: Yup, we should mark it as deprecated as soon as we have a configuration model working with full capabilities (models + UI + working launches).

>If we already know it's on the way out, marking it as deprecated sooner rather than later is better right?

:+1: Yes, definitively.

PR #89, opened 2020-07-11
[BD-2][TNL-7315] BB-2612: Add ADR for LTI 1.3 variable definition

giovannicimolin:
This PR add's an ADR to list the LTI 1.3 integration variable mapping and the reasoning behind each value.

**Reviewers:**
- [ ] @xitij2000
- [ ] @davidjoy 
- [ ] @davestgermain 
- [ ] @nasthagiri

### DIFF
--- a/docs/decisions/0002-lti-1p3-variables.rst
+++ b/docs/decisions/0002-lti-1p3-variables.rst
@@ -133,7 +133,7 @@ This claim is optional and includes information about the context where LTI laun
    * - id
      - course id if using CourseOffering
      - This claim requires a unique identifier per deployment,
-       and Course ID is already unique platform wide to represent a course.
+       and Course ID is already unique platform-wide to represent a course.
    * - type
      - http://purl.imsglobal.org/vocab/lis/v2/course#CourseOffering
      - LTI launches will mostly occur in the context of courses.
--- a/docs/decisions/0002-lti-1p3-variables.rst
+++ b/docs/decisions/0002-lti-1p3-variables.rst
@@ -114,7 +114,7 @@ The user data claims are identifiers passed through the LTI message to send user
      - User full name as defined in the spec, gated by a feature flag to avoid leaking PII to external tools.
    * - email
      - User email address
-     - User email, gated by a feature flag to avoid leaking PII to external tools.
+     - User email, gated by a feature flag to avoid leaking PII to external tools (this flag should be off by default).
 
 Reference: http://www.imsglobal.org/spec/lti/v1p3/#user-identity-claims-0
 
--- a/docs/decisions/0002-lti-1p3-variables.rst
+++ b/docs/decisions/0002-lti-1p3-variables.rst
@@ -17,7 +17,7 @@ variables required by LTI 1.3 launches.
 Decisions
 =========
 
-The following section go over required and optional variables in each supported claim in the LTI 1.3 implementation.
+The following sections go over required and optional variables in each supported claim in the LTI 1.3 implementation.
 This ADR should be updated when new claims are added.
 
 Deployment ID
--- a/docs/decisions/0002-lti-1p3-variables.rst
+++ b/docs/decisions/0002-lti-1p3-variables.rst
@@ -42,7 +42,7 @@ Roles
 The roles claim takes in a list of LTI 1.3 compliant roles from http://www.imsglobal.org/spec/lti/v1p3/#role-vocabularies.
 The claim name is: https://purl.imsglobal.org/spec/lti/claim/roles.
 
-The mappings from edX platform are shown in the table below:
+The mappings from Open edX roles are shown in the table below:
 
 .. list-table::
    :widths: auto
--- a/docs/decisions/0002-lti-1p3-variables.rst
+++ b/docs/decisions/0002-lti-1p3-variables.rst
@@ -157,5 +157,5 @@ Tech Debt
 Consequences
 ============
 
-* Changing the resource id to the value defined above will make all context on currently deployed XBlocks to be lost (since a different resource means the LTI is being launched
+* Changing the resource id to the value defined above will lose contexts on all currently deployed XBlocks (since a different resource means the LTI is being launched
   in a different location from the Tool's POV).
--- a/docs/decisions/0002-lti-1p3-variables.rst
+++ b/docs/decisions/0002-lti-1p3-variables.rst
@@ -53,7 +53,7 @@ The mappings from edX platform are shown in the table below:
      - Reasoning
    * - guest
      - Empty
-     - Guests are users not logged in, they shoudn't be able to access LTI content.
+     - Guests users are not logged in, they shouldn't be able to access LTI content.
    * - student
      - 'http://purl.imsglobal.org/vocab/lis/v2/institution/person#Student'
      - Students only have permission to view and interact with the LTI content.
--- a/docs/decisions/0002-lti-1p3-variables.rst
+++ b/docs/decisions/0002-lti-1p3-variables.rst
@@ -152,7 +152,7 @@ Reference: http://www.imsglobal.org/spec/lti/v1p3/#context-type-vocabulary
 Tech Debt
 =========
 
-* The current LTI 1.3 implementation merged on `master` uses the internally generate resource id from the LTI 1.1.1 implementation.
+* The current LTI 1.3 implementation merged on `master` uses the internally generated resource id from the LTI 1.1.1 implementation.
 
 Consequences
 ============
--- a/docs/decisions/0002-lti-1p3-variables.rst
+++ b/docs/decisions/0002-lti-1p3-variables.rst
@@ -1,0 +1,161 @@
+LTI 1.3 variable definition
+---------------------------
+
+Status
+======
+
+Provisional
+
+Context
+=======
+
+The LTI 1.3 specification defines multiple variables related to course, context, problem, student id and PII.
+
+The goal of this ADR is to provide well defined mappings from variables available in the platform and
+variables required by LTI 1.3 launches.
+
+Decisions
+=========
+
+The following section go over required and optional variables in each supported claim in the LTI 1.3 implementation.
+This ADR should be updated when new claims are added.
+
+Deployment ID
+~~~~~~~~~~~~~
+
+.. list-table::
+   :widths: auto
+   :header-rows: 1
+
+   * - LTI 1.3 variable name
+     - Open edX
+     - Reasoning
+   * - https://purl.imsglobal.org/spec/lti/claim/deployment_id
+     - "1" (fixed value)
+     - We're using single-tenant deployments, were each key is used for a single XBlock. Even when moving up to the course level,
+       we'll can keep this setup since multi-tenancy will only be required for instance wide configuration.
+
+See http://www.imsglobal.org/spec/lti/v1p3/#single-tenant-tool-registered-and-deployed-once for more details.
+
+Roles
+~~~~~
+The roles claim takes in a list of LTI 1.3 compliant roles from http://www.imsglobal.org/spec/lti/v1p3/#role-vocabularies.
+The claim name is: https://purl.imsglobal.org/spec/lti/claim/roles.
+
+The mappings from edX platform are shown in the table below:
+
+.. list-table::
+   :widths: auto
+   :header-rows: 1
+
+   * - Open edX role
+     - LTI 1.3 Roles included
+     - Reasoning
+   * - guest
+     - Empty
+     - Guests are users not logged in, they shoudn't be able to access LTI content.
+   * - student
+     - 'http://purl.imsglobal.org/vocab/lis/v2/institution/person#Student'
+     - Students only have permission to view and interact with the LTI content.
+   * - instructor
+     - 'http://purl.imsglobal.org/vocab/lis/v2/institution/person#Instructor'
+       'http://purl.imsglobal.org/vocab/lis/v2/institution/person#Student'
+     - Instructor have both instructor and student access to a tool. No admin permissions.
+   * - staff
+     - 'http://purl.imsglobal.org/vocab/lis/v2/system/person#Administrator'
+       'http://purl.imsglobal.org/vocab/lis/v2/institution/person#Instructor'
+       'http://purl.imsglobal.org/vocab/lis/v2/institution/person#Student'
+     - A staff user should be able to modify the tool settings and have full access to the content and settings.
+
+Resource link
+~~~~~~~~~~~~~
+
+This claim has properties for the resource link from which the launch message occurs. It represents a unique place in the course where the LTI launch happens.
+
+.. list-table::
+   :widths: auto
+   :header-rows: 1
+
+   * - Claim variable
+     - Block location (on course context).
+     - Reasoning
+   * - id
+     - Block location (on course context).
+     - The resource link id needs to be a identifier unique to the deployment.
+       In the Open edX platform, we already have an analog that is unique to the entire instance: block location.
+   * - title
+     - Display name (either from block or parent block)
+       Empty if not available/applicable.
+     - We use the `Display name` attribute to provide a descriptive title to the LTI tool.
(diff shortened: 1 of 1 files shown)


edx-webhook:
Thanks for the pull request, @giovannicimolin! I've created [BLENDED-430](https://openedx.atlassian.net/browse/BLENDED-430) to keep track of it in Jira.  More details are on the [BD-02](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1544913089/BD-02+LTI+v1.3+Consumer+XBlock+Support) project page.

docs/decisions/0002-lti-1p3-variables.rst
```diff
@@ -0,0 +1,123 @@
...
++---------------------------------------------------------+------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
+| https://purl.imsglobal.org/spec/lti/claim/deployment_id | "1" (fixed value)| We're using single-tenant deployments, were each key is used for a single XBlock. Even when moving up to the course level, we'll can keep this setup since multi-tenancy will only be required for instance wide configuration. |
++---------------------------------------------------------+------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```
xitij2000, line 29: I'm not sure I understand this. Do you mean that instead of one global client_id with multiple different deployments within that, we have multiple client_ids each with a single deployment?

docs/decisions/0002-lti-1p3-variables.rst
```diff
@@ -0,0 +1,123 @@
...
+~~~~~
+The roles claim takes in a list of LTI 1.3 compliant roles from http://www.imsglobal.org/spec/lti/v1p3/#role-vocabularies.
+The claim name is: https://purl.imsglobal.org/spec/lti/claim/roles.
+
+The mappings from edX platform are shown in the table below:
+
++---------------+------------------------------------------------------------------------+-----------------------------------------------------------------------------------------------+
+| Open edX role | LTI 1.3 Roles included                                                 | Reasoning                                                                                     |
++---------------+------------------------------------------------------------------------+-----------------------------------------------------------------------------------------------+
+| guest         | 'http://purl.imsglobal.org/vocab/lis/v2/institution/person#Student'    | Guests and students only have permission to view and interact with the LTI content.           |
```
xitij2000, line 43: I'm afraid I don't know what you mean by guest in the Open edX context? Is it an anonymous user?

docs/decisions/0002-lti-1p3-variables.rst
```diff
@@ -0,0 +1,123 @@
...
+Tech Debt
+=========
+
+* The current LTI 1.3 implementation merged on `master` uses the internally generate resource id from the LTI 1.1.1 implementation.
+
+Consequences
+============
+
+* Changing the resource id to the value defined above will make all context on currently deployed XBlocks to be lost (since a different resource means the LTI is being launched
+  in a different location from the Tool's POV).
```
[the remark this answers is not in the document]
giovannicimolin: >Is there some way to mitigate this? I guess we can temporary have a flag to make it use the old id, but is it possible to detect this and use the old context id for existing content?

The context ID is injected when doing the launch request here (https://github.com/edx/xblock-lti-consumer/blob/master/lti_consumer/lti_xblock.py#L1091).

We'd need to add additional logic there to migrate the values, but given that LTI 1.3 isn't widely used and and isn't enabled on edX.org, if we push a PR to make the changes soon we don't need to worry about migrating configurations.

docs/decisions/0002-lti-1p3-variables.rst
giovannicimolin: On LTI, Client ID > Deployment > Resource link (those are 3 separate things).

The client ID is coupled with a single key pair in the platform, and each Client ID can have multiple deployments inside the LMS. I've opted out of that approach for the LTI block implementation because it introduces a level of complexity not needed for the XBlock level implementation.

See http://www.imsglobal.org/spec/lti/v1p3/#single-tenant-tool-registered-and-deployed-once for more details on how LTI defines what a deployment is.

docs/decisions/0002-lti-1p3-variables.rst
giovannicimolin: I think that when I did the implementation I wanted to cover all use cases (values that could come up), but I think it makes sense to not pass any roles to a `guest` user since it's anonymous.
We should probably return an empty array in that case, does that seems better?

docs/decisions/0002-lti-1p3-variables.rst
xitij2000: Sure, so what I understand is, we aren't using the concept of deployments at all because there isn't really any use for it yet in the platform.

docs/decisions/0002-lti-1p3-variables.rst
xitij2000: I think my confusion stemmed from "guest" being an Open edX role, since I haven't encountered that terminology before. 

I don't know how common this use case would be, but perhaps for public courses LTI can use a student role but otherwise it can be an empty array? Does that make sense? 

I'm just outlining the approach here not suggesting we develop this right now.

docs/decisions/0002-lti-1p3-variables.rst
giovannicimolin: Yes, we're not using it, but we can easily support it. 
`Deployment` in LTI is basically an abstraction that allows us to use the same client id/key pair to deploy isolated tools (for different courses, for example).

docs/decisions/0002-lti-1p3-variables.rst
giovannicimolin: I think we should go with the empty array if the user is a `guest` (or any other type of user not specified).

docs/decisions/0002-lti-1p3-variables.rst
```diff
@@ -0,0 +1,123 @@
...
+
+Decisions
+=========
+
+The following section go over required and optional variables in each supported claim in the LTI 1.3 implementation.
+This ADR should be updated when new claims are added.
+
+Deployment ID
+~~~~~~~~~~~~~
++---------------------------------------------------------+------------------+---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
```
nedbat, line 25: BTW, .rst has a table style called list-table: https://docutils.sourceforge.io/docs/ref/rst/directives.html#list-table

It lets you author the file as a list, but be rendered as a table.

docs/decisions/0002-lti-1p3-variables.rst
nedbat: I am certain I don't understand this yet, because the LTI spec is kind of impenetrable.  But [§3.1.3 Tool Deployment](http://www.imsglobal.org/spec/lti/v1p3/#tool-deployment) says:
> A platform MUST generate a unique deployment_id for each tool it integrates with.
Doesn't that mean that `"1"` isn't enough?

nedbat: Hmm, never mind, the diagram a few paragraphs below explicitly shows three tools all using the same deployment id.

docs/decisions/0002-lti-1p3-variables.rst
giovannicimolin: @nedbat Much better, thanks for the suggestion!

giovannicimolin:

docs/decisions/0002-lti-1p3-variables.rst
```diff
@@ -0,0 +1,161 @@
...
+
+The LTI 1.3 specification defines multiple variables related to course, context, problem, student id and PII.
+
+The goal of this ADR is to provide well defined mappings from variables available in the platform and
+variables required by LTI 1.3 launches.
+
+Decisions
+=========
+
+The following section go over required and optional variables in each supported claim in the LTI 1.3 implementation.
```
nedbat, line 20:
```suggestion
The following sections go over required and optional variables in each supported claim in the LTI 1.3 implementation.
```

docs/decisions/0002-lti-1p3-variables.rst
```diff
@@ -0,0 +1,161 @@
...
+   :widths: auto
+   :header-rows: 1
+
+   * - LTI 1.3 variable name
+     - Open edX
+     - Reasoning
+   * - https://purl.imsglobal.org/spec/lti/claim/deployment_id
+     - "1" (fixed value)
+     - We're using single-tenant deployments, were each key is used for a single XBlock. Even when moving up to the course level,
+       we'll can keep this setup since multi-tenancy will only be required for instance wide configuration.
```
nedbat, line 36:
```suggestion
       we can keep this since multi-tenancy will only be required for instance wide configuration.
```

docs/decisions/0002-lti-1p3-variables.rst
```diff
@@ -0,0 +1,161 @@
...
+
+See http://www.imsglobal.org/spec/lti/v1p3/#single-tenant-tool-registered-and-deployed-once for more details.
+
+Roles
+~~~~~
+The roles claim takes in a list of LTI 1.3 compliant roles from http://www.imsglobal.org/spec/lti/v1p3/#role-vocabularies.
+The claim name is: https://purl.imsglobal.org/spec/lti/claim/roles.
+
+The mappings from edX platform are shown in the table below:
```
nedbat, line 45:
```suggestion
The mappings from Open edX roles are shown in the table below:
```

docs/decisions/0002-lti-1p3-variables.rst
```diff
@@ -0,0 +1,161 @@
...
+.. list-table::
+   :widths: auto
+   :header-rows: 1
+
+   * - Open edX role
+     - LTI 1.3 Roles included
+     - Reasoning
+   * - guest
+     - Empty
+     - Guests are users not logged in, they shoudn't be able to access LTI content.
```
nedbat, line 56:
```suggestion
     - Guests users are not logged in, they shouldn't be able to access LTI content.
```

docs/decisions/0002-lti-1p3-variables.rst
```diff
@@ -0,0 +1,161 @@
...
+     - Students only have permission to view and interact with the LTI content.
+   * - instructor
+     - 'http://purl.imsglobal.org/vocab/lis/v2/institution/person#Instructor'
+       'http://purl.imsglobal.org/vocab/lis/v2/institution/person#Student'
+     - Instructor have both instructor and student access to a tool. No admin permissions.
+   * - staff
+     - 'http://purl.imsglobal.org/vocab/lis/v2/system/person#Administrator'
+       'http://purl.imsglobal.org/vocab/lis/v2/institution/person#Instructor'
+       'http://purl.imsglobal.org/vocab/lis/v2/institution/person#Student'
+     - A staff user should be able to modify the tool settings and have full access to the content and settings.
```
nedbat, line 68: We should say something here about why we are choosing `http://purl.imsglobal.org/vocab/lis/v2/system/person#Administrator` instead of `http://purl.imsglobal.org/vocab/lis/v2/institution/person#Administrator`

docs/decisions/0002-lti-1p3-variables.rst
```diff
@@ -0,0 +1,161 @@
...
+     - LTI 1.3 Roles included
+     - Reasoning
+   * - guest
+     - Empty
+     - Guests are users not logged in, they shoudn't be able to access LTI content.
+   * - student
+     - 'http://purl.imsglobal.org/vocab/lis/v2/institution/person#Student'
+     - Students only have permission to view and interact with the LTI content.
+   * - instructor
+     - 'http://purl.imsglobal.org/vocab/lis/v2/institution/person#Instructor'
```
nedbat, line 61: Why Instructor instead of Staff?

docs/decisions/0002-lti-1p3-variables.rst
```diff
@@ -0,0 +1,161 @@
...
+   * - Claim variable
+     - Open edX
+     - Reasoning
+   * - sub
+     - external user id (with the `lti` type)
+     - Using the id provided by the external_user_id from openedx core.
+       See https://github.com/edx/edx-platform/blob/master/openedx/core/djangoapps/external_user_ids/docs/decisions/0001-externalid.rst
+   * - name
+     - User full name
+     - User full name as defined in the spec, gated by a feature flag to avoid leaking PII to external tools.
```
nedbat, line 114: The feature flag defaults on or off?

docs/decisions/0002-lti-1p3-variables.rst
```diff
@@ -0,0 +1,161 @@
...
+   :widths: auto
+   :header-rows: 1
+
+   * - Claim variable
+     - Open edX
+     - Reasoning
+   * - id
+     - course id if using CourseOffering
+     - This claim requires a unique identifier per deployment,
+       and Course ID is already unique platform wide to represent a course.
```
nedbat, line 136:
```suggestion
       and Course ID is already unique platform-wide to represent a course.
```

docs/decisions/0002-lti-1p3-variables.rst
```diff
@@ -0,0 +1,161 @@
...
+     - Using a readable identifier of the organization and course name to the tool.
+
+Reference: http://www.imsglobal.org/spec/lti/v1p3/#context-type-vocabulary
+
+
+
+Tech Debt
+=========
+
+* The current LTI 1.3 implementation merged on `master` uses the internally generate resource id from the LTI 1.1.1 implementation.
```
nedbat, line 155:
```suggestion
* The current LTI 1.3 implementation merged on `master` uses the internally generated resource id from the LTI 1.1.1 implementation.
```

docs/decisions/0002-lti-1p3-variables.rst
```diff
@@ -0,0 +1,161 @@
...
+
+Consequences
+============
+
+* Changing the resource id to the value defined above will make all context on currently deployed XBlocks to be lost (since a different resource means the LTI is being launched
```
nedbat, line 160:
```suggestion
* Changing the resource id to the value defined above will lose contexts on all currently deployed XBlocks (since a different resource means the LTI is being launched
```

nedbat (changes requested):
Looks good. I made minor copy-edits, and asked for some clarifications.

docs/decisions/0002-lti-1p3-variables.rst
giovannicimolin: If should probably default to `off`.
This is currently not implemented through a feature flag, and relies on the `lti-configuration` service from the LTI 1.1 implementation. We can turn it into a flag, since there's not that much documentation on the LTI Configuration service.

docs/decisions/0002-lti-1p3-variables.rst
giovannicimolin: @nedbat I've picked the most generic one to give the maximum amount of permissions to a course creator. We should probably reevaluate these roles once we have tested the platform with a few LTI tools. 

I've just noticed that the spec mentions:
```
The role vocabularies are derived from the LIS 2.0 specification [LIS-20]. LTI divides them into core and non-core roles. Core roles are those which are most likely to be relevant within LTI and hence vendors should support them by best practice. Vendors may also use the non-core rules, but they may not be widely used.
```
But the cores role are just `Administrator` and `None`: http://www.imsglobal.org/spec/lti/v1p3/#lis-vocabulary-for-system-roles

docs/decisions/0002-lti-1p3-variables.rst
giovannicimolin: We don't want to give instructors staff permissions, no?
In my understanding, `instructors` don't have the same level of access that `staff` (course creators) do.

Is this assumption wrong?

nedbat: We need to remove Student from the instructor and staff lists.

giovannicimolin: @nedbat Done.

Agrendalath:
Sorry for the noise here, I was testing the labels.

PR #93, opened 2020-07-23
[BD-24] [TNL-7318] BB-2355: Move LTI configuration to models.

giovannicimolin:
This PR moves the LTI configuration handling to a django model. All the settings are still stored in the XBlock, but the retrieving and instancing of LTI config is done through a Python API now.

**Testing instructions:**
1. Checkout and install this branch.
2. Set up a LTI 1.1 and a LTI 1.3 blocks in a course (follow the guide on https://github.com/edx/xblock-lti-consumer#testing-against-an-lti-provider).
3. Test that LTI 1.1 works from the LMS and Studio.
4. Test that LTI 1.3 launch works from LMS.
5. Check on the LTI Configuration on Django admin.

**Notes:**
There's no new "features" here, and this is just laying the groundwork to support LTI Advantage extensions.
This is the first implementation step of https://github.com/edx/xblock-lti-consumer/pull/92.

**Reviewers:**
- [x] @viadanna 
- [x] @davidjoy 
- [x] @davestgermain


edx-webhook:
Thanks for the pull request, @giovannicimolin! I've created [BLENDED-504](https://openedx.atlassian.net/browse/BLENDED-504) to keep track of it in Jira.   More details are on the [BD-24](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1545076784/BD-24+LTI+v1.3+Improvements+Advantage+Support) project page.

viadanna:
@giovannicimolin I'm trying to test this PR on a master devstack but the XBlocks fails to load correctly after running migrations:
```
edx.devstack-master.studio |   File "/edx/app/edxapp/edx-platform/xblock-lti-consumer/lti_consumer/models.py", line 92, in _get_lti_1p1_consumer
edx.devstack-master.studio |     key, secret = block.lti_provider_key_secret
edx.devstack-master.studio |   File "/edx/app/edxapp/edx-platform/xblock-lti-consumer/lti_consumer/lti_xblock.py", line 657, in lti_provider_key_secret
edx.devstack-master.studio |     for lti_passport in self.course.lti_passports:
edx.devstack-master.studio |   File "/edx/app/edxapp/edx-platform/xblock-lti-consumer/lti_consumer/lti_xblock.py", line 650, in course
edx.devstack-master.studio |     return self.runtime.descriptor_runtime.modulestore.get_course(self.course_id)  # pylint: disable=no-member
edx.devstack-master.studio |   File "/edx/app/edxapp/edx-platform/common/lib/xmodule/xmodule/x_module.py", line 2025, in __getattr__
edx.devstack-master.studio |     return getattr(self._descriptor_system, name)
edx.devstack-master.studio | AttributeError: 'CachingDescriptorSystem' object has no attribute 'descriptor_runtime'
```

Any ideas what I'm doing wrong?

giovannicimolin:
@viadanna I've hit that issue too at first, then I reinstalled the block and cleaned up all pyc files and it started working again.

lti_consumer/api.py
```diff
@@ -0,0 +1,44 @@
...
+
+    Returns LTI configuration ID.
+    """
+    lti_config, _ = LtiConfiguration.objects.get_or_create(
+        location=block_location,
+        config_store=LtiConfiguration.CONFIG_ON_XBLOCK,
+    )
+
+    if lti_config.version != lti_version:
+        lti_config.version = lti_version
```
viadanna, line 26: This means configs are overwritten when changing versions, right? Would it make sense to keep both?

giovannicimolin: @viadanna Yes, but this will only happen for configuration stored in the block.
Also, since the configuration is still stored in block variables, you a LTI 1.1 block is configured and then switched to LTI 1.3, the settings will be kept.

This overriding just happen for configurations stored with the `CONFIG_ON_XBLOCK` type.

giovannicimolin:
@davestgermain @davidjoy I've rebased the PR bumped the version and squashed the commits. This is good to do.

Note that this will require running migrations on stage/production, so I'm not sure how we should roll-out this feature.

lti_consumer/models.py
```diff
@@ -0,0 +1,143 @@
...
+    # Configuration storage
+    # Initally, this only supports the configuration
+    # stored on the block, but should be expanded to
+    # enable storing LTI configuration in this model.
+    CONFIG_ON_XBLOCK = 'CONFIG_ON_XBLOCK'
+    CONFIG_STORE_CHOICES = [
+        (CONFIG_ON_XBLOCK, 'Configuration Stored on XBlock fields'),
+    ]
+    config_store = models.CharField(
+        max_length=25,
```
davestgermain, line 48: As far as I know, there isn't any space advantage to making this field small.

lti_consumer/models.py
```diff
@@ -0,0 +1,143 @@
...
+                "Block location missing or invalid. The block was possibly deleted."
+            )
+
+        return modulestore().get_item(usage_key)
+
+    def _get_lti_1p1_consumer(self):
+        """
+        Return a class of LTI 1.1 consumer.
+
+        Not implemented yet.
```
davestgermain, line 87: It is implemented

lti_consumer/lti_xblock.py
```diff
@@ -830,41 +828,37 @@ def is_past_due(self):
...
-            tool_keyset_url=None,
+        # Runtime import since this will only run in the
+        # Open edX LMS/Studio environments.
+        # pylint: disable=import-outside-toplevel
+        from lti_consumer.api import (
+            get_or_create_local_lti_config,
+            get_lti_consumer
+        )
+
+        lti_config_id = get_or_create_local_lti_config(
```
davestgermain, line 855: Instead of having to get/create the `lti_config_id` and then immediately looking up the model by id, how about having something like `get_lti_consumer(config_id=None, block=self)`, which will get_or_create the config and return the consumer? And since configuration is only on the block for now, you can avoid having to load the block again (in `_load_block`), since it's already being passed in.

lti_consumer/models.py
```diff
@@ -0,0 +1,143 @@
...
+    # stored in this model in a JSON field.
+    # This field stores a UsageKey.
+    location = models.CharField(
+        max_length=255,
+        db_index=True,
+        null=True,
+        blank=True,
+    )
+
+    def _load_block(self):  # pragma: no cover
```
davestgermain, line 64: How about this instead? This way, you can set the block on the config if it's passed in to `get_lti_consumer`, otherwise it'll load the block.

```python
@property
def block(self):
    block = getattr(self, '_block', None)
    if block is None:
        from xmodule.modulestore.django import modulestore
        block = self._block = modulestore().get_item(self.location)
    return block

@block.setter
def block(self, block):
    self._block = block
```

lti_consumer/models.py
```diff
@@ -0,0 +1,143 @@
...
+        max_length=25,
+        choices=CONFIG_STORE_CHOICES,
+        default=CONFIG_ON_XBLOCK,
+    )
+
+    # Block location where the configuration is stored.
+    # In the future, the LTI configuration will be
+    # stored in this model in a JSON field.
+    # This field stores a UsageKey.
+    location = models.CharField(
```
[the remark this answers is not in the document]
giovannicimolin: Much better! I though that this was part of the core, but it seems that I was wrong. I'll implement it.

giovannicimolin:
@davestgermain I've implemented your review suggestions and added a few more tests, can you take a look again?

davestgermain:
@giovannicimolin looks good to me! Squash and merge 👍

giovannicimolin:
@davestgermain Done!

PR #96, opened 2020-08-07
Convert is_past_due to a method

mikix:
This is just to keep it in line with other xblocks that define an is_past_due attribute. Nicer if they all have the same access pattern.


mikix:
> Did you do a cursory check as well to see if we have any other xblocks that should be updated?

Yeah, I didn't see any others in github.

PR #99, opened 2020-09-02
Correct the license copies

nedbat:
We said Apache, and it is Apache, so why have a copy of AGPL?

### DIFF
--- a/NOTICE
+++ b/NOTICE
@@ -1,4 +1,6 @@
-Copyright (C) 2015  edX
+Open edX LTI Consumer Library
+
+Copyright (C) 2015-2020, edX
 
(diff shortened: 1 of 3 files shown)


nedbat:
@giovannicimolin A simple fix.

giovannicimolin:
:+1: 

I've checked the original discussion at #1 (where the LICENSE file was introduced) and also https://open.edx.org/trademark-licensing-details/.

LGTM.

nedbat:
OK, I've switched to AGPL.

giovannicimolin (approved):
:+1:

- [x] I read through the changes
- [x] Includes documentation

PR #102, opened 2020-09-04
This code isn't used?

nedbat:
I turned on coverage measurement of the tests (use "make covreport" to see the results). It showed that this code wasn't used.  Is it OK to remove it?

### DIFF
--- a/lti_consumer/tests/unit/test_utils.py
+++ b/lti_consumer/tests/unit/test_utils.py
@@ -53,34 +53,6 @@ def make_request(body, method='POST'):
     return request
 
 
-def patch_signed_parameters(func):
-    """
-    Prepare the patches for the get_signed_lti_parameters function for tests.
-    """
-    func = patch(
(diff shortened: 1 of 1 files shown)


nedbat:
@pcockwell @giovannicimolin

giovannicimolin:
@nedbat I'm guessing that if it isn't used we can remove it.
@pcockwell What's the original intent of this code? Can it be removed? It doesn't seem to be used.

pcockwell:
@nedbat @giovannicimolin 

Based on what I can figure out using Git Blame, this was originally written by Douglas Hall in Nov 2015 (e633f55b), and rewritten by Omar Al-Ithawi in Jan 2019 (c857e5d0). I have no idea what the purpose or previous usage of `faulty_processor` was, but `patch_signed_parameters` was used as a utility decorator on test functions that made LTI request calls.

I've been unable to find anywhere they are used, or a reason to keep them, so I would venture a guess that they can be removed.

PR #107, opened 2020-09-24
Staff and instructors shouldn't have the Student role.

nedbat:
With this change, the IMS validation suite passes:
screencapture-ltiadvantagevalidator-imsglobal-org-ltiplatform-no-back-2020-09-24-16_52_22


nedbat:
@giovannicimolin A simple change.
I'm not sure admins should still have the instructor role, but we can figure that out later?

giovannicimolin (approved):
:+1:

- [x] I tested this:
1. CI checks are green.
2. LTI certification suite passes.
- [x] I read through the code: simple changes to role scopes.

@nedbat Good to go.

PR #108, opened 2020-09-30
[BD-24] Implement LTI AGS Score Publish Service and Results

pcockwell:
This PR implements the `/scores` endpoint for publishing new `LtiAgsScore`s for `LtiAgsLineItem`s, and the `/results` endpoint for retrieving scores. This PR builds directly upon #92 

**JIRA tickets**: 
[BD-24]

**Dependencies**: #92 

**Testing instructions**:

1. Checkout this branch for the `xblock-lti-consumer` repository located in your devstack `src` folder.
2. Follow the [ReadMe Instructions](https://github.com/edx/xblock-lti-consumer/#installing-in-docker-devstack) to install on both `lms` and `studio`
3. Run migrations with `make dev.migrate.lms dev.migrate.studio`
4. Edit `studio.yml` and enable the following feature flag:
```
FEATURES:
    LTI_1P3_ENABLED: true
```
5. Restart Studio (`make studio-restart`) and log in.
6. Complete steps 2-4 from the [LTI 1.3 ReadMe Setup Instructions](https://github.com/edx/xblock-lti-consumer/#lti-13). Make note of the `OAuth Token URL` for use in step 9
7. Go to https://lti-ri.imsglobal.org/keygen/index and copy the public key, paste it into the LTI 1.3 Tool Public Key field. Save the private key for later.
8. Generate a client token
    * Go to www.jwt.io
    * select RS-256
    * Fill in both the private and public keys section in the bottom right
    * Set the header and payload as follows
```
# HEADER:
{
  "alg": "RS256",
  "typ": "JWT",
  "kid": "123"
}
```
```
# PAYLOAD:
{
  "sub": "1234567890",
  "name": "John Doe",
  "admin": true,
  "iat": 1516239022
}
```

9. Retrieve your LTI Authentication Keys. Copy the OAuth Token URL and make the following requests (don't forget to replace the JWT from the previous step). This will yield your OAuth Access Token to use the endpoints. 
```
curl --request POST \
  --url '<OAUTH_TOKEN_URL>' \
  --header 'content-type: application/x-www-form-urlencoded' \
  --data grant_type=client_credentials \
  --data 'scope=https://purl.imsglobal.org/spec/lti-ags/scope/lineitem https://purl.imsglobal.org/spec/lti-ags/scope/score https://purl.imsglobal.org/spec/lti-ags/scope/result.readonly \
  --data client_assertion_type=urn:ietf:params:oauth:client-assertion-type:jwt-bearer \
  --data client_assertion=<JWT_TOKEN>
```
10. Go into Django admin, and copy the LTI Configuration ID (http://localhost:18000/admin/lti_consumer/lticonfiguration/).
11. Make a request to the LTI AGS List endpoint to create a line item
```
curl --request POST \
  --url http://localhost:18000/api/lti_consumer/v1/lti/<LTI_CONFIG_ID>/lti-ags/ \
  --header 'authorization: Bearer <OAUTH_ACCESS_TOKEN>' \
  --header 'content-type: application/vnd.ims.lis.v2.lineitem+json' \
  --data '{"scoreMaximum": 100, "label": "Score", "tag": "score", "resourceId": "some-random-string"}'
```
12. Make a request to the LTI AGS List endpoint to create a score
```
curl --request POST \
  --url http://localhost:18000/api/lti_consumer/v1/lti/<LTI_CONFIG_ID>/lti-ags/1/scores \
  --header 'authorization: Bearer <OAUTH_ACCESS_TOKEN>' \
  --header 'content-type: application/vnd.ims.lis.v2.lineitem+json' \
  --data '{"timestamp": "2020-09-27T18:54:36.736+00:00","scoreGiven" : 86,"scoreMaximum" : 100,"comment" : "This is exceptional work.","activityProgress" : "Completed","gradingProgress": "FullyGraded","userId" : "123"}'
```
13. Make a request to the LTI AGS List endpoint to create multiple more scores (at least 2). Please change the `timestamp`, and on all of them, and modify the `userId` for at least 1 other score.
```
curl --request POST \
  --url http://localhost:18000/api/lti_consumer/v1/lti/<LTI_CONFIG_ID>/lti-ags/1/scores \
  --header 'authorization: Bearer <OAUTH_ACCESS_TOKEN>' \
  --header 'content-type: application/vnd.ims.lis.v2.lineitem+json' \
  --data '{"timestamp": "2020-09-27T18:54:36.736+00:00","scoreGiven" : 86,"scoreMaximum" : 100,"comment" : "This is exceptional work.","activityProgress" : "Completed","gradingProgress": "FullyGraded","userId" : "123"}'
```
14. Make a request to the LTI AGS List endpoint to retrieve results. This should return all results
```
curl --request GET \
  --url http://localhost:18000/api/lti_consumer/v1/lti/<LTI_CONFIG_ID>/lti-ags/1/results \
  --header 'authorization: Bearer <OAUTH_ACCESS_TOKEN>'
```
15. Make a request to the LTI AGS List endpoint to retrieve results and filter them using `user_id` and `limit`
```
curl --request GET \
  --url http://localhost:18000/api/lti_consumer/v1/lti/<LTI_CONFIG_ID>/lti-ags/1/results?user_id=123&limit=2 \
  --header 'authorization: Bearer <OAUTH_ACCESS_TOKEN>'
```

**Reviewers**
- [ ] @giovannicimolin 
- [ ] edX reviewer[s] TBD


edx-webhook:
Thanks for the pull request, @pcockwell! I've created [BLENDED-610](https://openedx.atlassian.net/browse/BLENDED-610) to keep track of it in Jira.   More details are on the [BD-24](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1545076784/BD-24+LTI+v1.3+Improvements+Advantage+Support) project page.


lti_consumer/views.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+    serializer_class = LtiAgsLineItemSerializer
+    pagination_class = None
+
+    # Custom permission classes for LTI APIs
+    authentication_classes = [Lti1p3ApiAuthentication]
+    permission_classes = [LtiAgsPermissions]
+
+    # Renderer/parser classes to accept LTI AGS content types
+    renderer_classes = [
```
giovannicimolin, line 36: @pcockwell Both the results and scores service return different content types:
See https://www.imsglobal.org/spec/lti-ags/v2p0#media-type-and-schema and https://www.imsglobal.org/spec/lti-ags/v2p0#media-type-and-schema.

You can either build new renderer classes, or modify the existing ones to return the proper `Content-Types` depending on the request.

giovannicimolin (commented):
@pcockwell Haven't done a full review yet, but great work on the implementation so far! :rocket: 
I've left a comment about the renderer classes.

lti_consumer/plugin/views.py
```diff
@@ -130,3 +144,79 @@ def get_queryset(self):
...
+          * An copy of the saved record, formatted by LtiAgsScoreSerializer
+                and returned with the media-type for LineItemScoreRenderer
+        """
+        line_item = self.get_object()
+        serializer = LtiAgsScoreSerializer(
+            data=request.data,
+            context={'request': self.request},
+        )
+        serializer.is_valid(raise_exception=True)
+        serializer.save(line_item=line_item)
```
giovannicimolin, line 216: To avoid needing to handle multiple `LtiAgsScore` entries for each user, we can use a single item and modify when new data gets pushed through the score service. 

LTI AGS doesn't require the platform to keep a history of changes to the students score, and just needs to make sure that scores are not updated out of order (looking at the `timestamp` field):
>The platform MUST NOT update a result if the last timestamp on record is later than the incoming score update. It may just ignore the incoming score update, or log it if it maintains any kind of history or for traceability.

That will simplify the implementation in a sense that we won't need to implement signals and clear up old values when a empty score or comment value is sent through the tool.

We can use a `unique_together` relation between the User and the LineItem.

giovannicimolin (changes requested):
@pcockwell I found a few things you need to implement so we can meet the spec requirements and a few nits.

This is only my `spec` compliance review.

lti_consumer/models.py
```diff
@@ -204,3 +204,75 @@ def __str__(self):
             self.resource_link_id,
             self.label,
         )
+
+
+class LtiAgsScore(models.Model):
```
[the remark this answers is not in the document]
giovannicimolin: @pcockwell The `unique_together` might not be necessary if we implement what I've commented here:
https://github.com/edx/xblock-lti-consumer/pull/108#discussion_r500535603

lti_consumer/models.py
```diff
@@ -204,3 +204,75 @@ def __str__(self):
...
+    # This links the score to a specific line item
+    line_item = models.ForeignKey(
+        LtiAgsLineItem,
+        on_delete=models.CASCADE,
+        related_name='scores',
+    )
+
+    timestamp = models.DateTimeField()
+    score_given = models.FloatField()
+    score_maximum = models.FloatField()
```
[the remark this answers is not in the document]
pcockwell: I've decided to do a bunch of the validation/checking on the serializer itself, so there shouldn't ever be a record in the DB with a null `score_given` or `score_maximum`. All record manipulation (timestamp validation and deletion of existing records when no score is provided) happens as a result of the serializer.

lti_consumer/plugin/views.py
```diff
@@ -130,3 +144,84 @@ def get_queryset(self):
...
+          * A JSON object capable of being serialized by LtiAgsScoreSerializer
+
+        Returns:
+          * An copy of the saved record, formatted by LtiAgsScoreSerializer
+                and returned with the media-type for LineItemScoreRenderer
+        """
+        line_item = self.get_object()
+
+        user_id = request.data.get('userId')
+        existing_score = line_item.scores.filter(user_id=user_id).first()
```
giovannicimolin, line 213: @pcockwell You can use `line_item.scores.get(user_id=user_id)`, given the unique constraint.

lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
```diff
@@ -89,3 +89,137 @@ class Meta:
             'startDateTime',
             'endDateTime',
         )
+
+
+class LtiAgsScoreSerializer(serializers.ModelSerializer):
```
giovannicimolin, line 94: >I've decided to do a bunch of the validation/checking on the serializer itself, so there shouldn't ever be a record in the DB with a null score_given or score_maximum. All record manipulation (timestamp validation and deletion of existing records when no score is provided) happens as a result of the serializer.

The new implementation is more complex and only handle cases through the LTI-AGS API, and doesn't put any safeguards in the model (we'll have internal python APIs later).

> so there shouldn't ever be a record in the DB with a null score_given or score_maximum.

Adding this artificial requirement only makes it harder to implement the actual working endpoint. Please remove the validators from the serializer and allow `score_given` and `score_maximum` to store null values.

This needs to be a valid request:
```
{
  "timestamp": "2017-04-16T18:54:36.736+00:00",
  "activityProgress" : "Completed",
  "gradingProgress" : "PendingManual",
  "userId" : "5323497"
}
```
And it won't work in the current implementation.

lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
```diff
@@ -89,3 +89,137 @@ class Meta:
...
+    scoreMaximum = serializers.FloatField(source='score_maximum', required=False, allow_null=True, default=None)
+    comment = serializers.CharField(required=False, allow_null=True)
+    activityProgress = serializers.CharField(source='activity_progress')
+    gradingProgress = serializers.CharField(source='grading_progress')
+    userId = serializers.CharField(source='user_id')
+
+    def validate_scoreGiven(self, value):
+        if not self.instance and value is None:
+            raise serializers.ValidationError('Cannot delete score. No score currently exists for this user')
+        return value
```
giovannicimolin, line 129: This does not comply to the spec. 
This is a valid score request:
```
{
  "timestamp": "2017-04-16T18:54:36.736+00:00",
  "activityProgress" : "Completed",
  "gradingProgress" : "PendingManual",
  "userId" : "5323497"
}
```
If this is the first request, it'll fail. Remove this validator and make `score_given` allow null values in the model.

lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
```diff
@@ -89,3 +89,137 @@ class Meta:
...
+
+    def validate_scoreMaximum(self, value):
```
giovannicimolin, line 131: :+1:

lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
```diff
@@ -89,3 +89,137 @@ class Meta:
...
+
+    def update(self, instance, validated_data):
+        # When 'scoreGiven' is not present, this indicates there is presently no score for that user,
+        # and the platform should clear any previous score value it may have previously received from
+        # the tool and stored for that user and line item
+        if instance and validated_data.get('score_given', None) is None:
+            instance.delete()
+            return instance
+
+        return super().update(instance, validated_data)
```
giovannicimolin, line 152: No, this is not the desired behavior.
Again, with the same request as before. If a user already has a grade, and this is received:
```
{
  "timestamp": "2017-04-16T18:54:36.736+00:00",
  "activityProgress" : "Completed",
  "gradingProgress" : "PendingManual",
  "userId" : "5323497"
}
```
You should just store `None` in score, but leave the other attributes.

lti_consumer/plugin/views.py
```diff
@@ -130,3 +144,84 @@ def get_queryset(self):
...
+                one record per unique `user_id` will be returned
+          * limit (integer): The maximum number of records to return. Records are
+                sorted with most recent timestamp first
+
+        Returns:
+          * An array of Result records, formatted by LtiAgsResultSerializer
+                and returned with the media-type for LineItemResultsRenderer
+        """
+        line_item = self.get_object()
+        scores = line_item.scores.all().order_by('-timestamp')
```
giovannicimolin, line 168: Haven't done a full test yet, but I think this can be replaced with a simple queryset (given the database constraints implemented in `models.py`).

giovannicimolin (changes requested):
@pcockwell Another round of review. 

The new changes break spec rules and didn't went in the direction I expected. I've left a few more comments.

lti_consumer/plugin/views.py
pcockwell: I'll add a comment to make it explicit, but I want `existing_score` to be set to `None` if there are no scores. Using `filter` and `first` does that, whereas `get` raises a `DoesNotExist` error.

lti_consumer/plugin/views.py
pcockwell: @giovannicimolin I'm not sure what you mean by this. Could you explain?

lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
pcockwell: @giovannicimolin 
In a situation where a user has a score, the score is 'erased' using a record with a blank `scoreGiven` (thus saving `null` values to the database for the blanks), and then a request is made to the results endpoint, should the endpoint return 1 record with a blank score, or no records?

giovannicimolin: The LTI spec mentions that the Results service can ignore any score entry that doesn't have a final grade:
https://www.imsglobal.org/spec/lti-ags/v2p0/#results-url-may-skip-empty-results

So this simplifies a lot our implementation, and don't need to return any result.

giovannicimolin (commented):
Nice work! I'll do some testing later on today and comment if I find anything else, thanks for addressing all comments.

lti_consumer/plugin/views.py
giovannicimolin: @pcockwell I'm a bit confused on what the for loop below is doing. I don't think it's necessary since only user only has a single Score for each LineItem.

pcockwell: This is an outdated view. Here is the new code

https://github.com/open-craft/xblock-lti-consumer/blob/0a98de3921aa65ad507620c253957509ef084d4b/lti_consumer/plugin/views.py#L167-L182

giovannicimolin (changes requested):
I see a server error (500), when trying to send a Score with `scoreGiven` but without `scoreMaximum`.

This should fail gracefully and return a proper error message.
CURL:
```

curl --request POST \
  --url http://localhost:18000/api/lti_consumer/v1/lti/1/lti-ags/1/scores \
  --header 'authorization: Bearer TOKEN' \
  --header 'content-type: application/vnd.ims.lis.v1.score+json' \
  --data '{"timestamp": "2020-09-27T18:54:36.740+00:00","scoreGiven" : 86,"comment" : "This is exceptional work.","activityProgress" : "Completed","gradingProgress": "FullyGraded","userId" : "3334"}'
```
---
All my other testing was successful (some testing instructions had the wrong content type, which made the content negotiation fail, but easy to fix). 

Also, I've found a bug on https://github.com/edx/xblock-lti-consumer/blob/master/lti_consumer/lti_1p3/extensions/rest_framework/permissions.py#L42: the correct is `destroy`, not `delete`. No need to fix this here. 

I've tested:
- [x] LineItem management
- [x] Token creation with multiple scopes
- [x] Permissions
- [x] Creating and modifying score (with the only failure detailed above)
- [x] Results endpoint, using both filters

You should:
1. Fix the issue above
2. Add tests to the views. I see a lot of code not covered by any tests (create a new test file based on https://github.com/edx/xblock-lti-consumer/blob/master/lti_consumer/tests/unit/plugin/test_views_lti_ags.py to test the views).
3. Fix quality issues.

lti_consumer/models.py
```diff
@@ -204,3 +206,91 @@ def __str__(self):
...
+
+    def save(self, *args, **kwargs):  # pylint: disable=arguments-differ
+        self.full_clean()
+        super().save(*args, **kwargs)
+
+    def __str__(self):
+        return "{} Score ({})".format(
+            self.line_item,
+            self.score_given,
+        )
```
giovannicimolin, line 293: Two small nits here:

1. Add a test for this. 
2. Remove the `ForeignKey` dependency from this (it will lead to a query explosion on the Django Admin page). The final string can be `LineItem {self.line_item.id}: score {self.score_given} out of {self.score_maximum} - {self.grading_progress}` (use format, I just used this format to illustrate).

lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
```diff
@@ -89,3 +90,159 @@ class Meta:
...
+                'pk': obj.line_item.pk
+            },
+            request=request,
+        )
+
+    def get_resultMaximum(self, obj):
+        if obj.score_maximum <= 0:
+            return 1
+
+        return obj.score_maximum
```
giovannicimolin, line 237: @pcockwell We don't need this for now. You can remove this and the serializer method field above and just return the final value.
```suggestion

```

lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
```diff
@@ -89,3 +90,159 @@ class Meta:
...
+            if errors:
+                # Unset any existing validated data because the model failed to validate
+                self._validated_data = {}  # pylint: disable=attribute-defined-outside-init
+
+                if raise_exception:
+                    raise serializers.ValidationError(errors)
+
+                return False
+
+        return True
```
giovannicimolin, line 167: @pcockwell I think we can replace this by simply duplicating the validation in the serializer.
The main issue with this implementation is that it does a 2-step validation process, where if I send a request like this:
```
{"timestamp": "2020-09-27T18:54:36.740+00:00","scoreGiven":123, "comment" : "This is exceptional work.","activityProgress" : "Completed","gradingProgress": "FullyGraded"}
```
It'll first fail with `userId` not present, and after fixing it, it'll throw `maximumGrade` not present.

```suggestion
    def validate_scoreGiven(self, value):
        """
        Ensure that if an existing record is being updated, that the timestamp is in the after the existing one
        """
        if 'scoreMaximum' not in self.initial_data:
            raise serializers.ValidationError('ScoreMaximum is a required field when setting a scoreGiven value.')

        return value
```

I prefer simpler code, even if that means some duplication. It's easier for maintainability.

lti_consumer/lti_1p3/extensions/rest_framework/permissions.py
```diff
@@ -23,25 +23,32 @@ def has_permission(self, request, view):
         """
         Check if LTI AGS permissions are set in auth token.
```
giovannicimolin, line 24: You are not testing the new actions on `test_permissions.py`.

lti_consumer/tests/unit/plugin/test_views_lti_ags.py
```diff
@@ -274,3 +301,556 @@ def test_create_lineitem_invalid_resource_link_id(self):
         )
 
         self.assertEqual(response.status_code, 400)
+
+
+class LtiAgsViewSetScoresTests(LtiAgsLineItemViewSetTestCase):
```
giovannicimolin, line 306: Nicely written tests! :)

giovannicimolin (changes requested):
@pcockwell A few extra nits to simplify the code. It's working well for me, nice work!

@nedbat Can you run the LTI Certification tool here? It should work and pass all tests on this branch.

lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
pcockwell: @giovannicimolin I'm not quite sure I understand... Why are we removing this? It matches the spec right? The results will only ever pull scores with non-null values for `score_given`, and `score_maximum` is required when `score_given` is non-null. What do you mean this is not needed right now?

giovannicimolin: @pcockwell You can leave this as is. I've misread the spec here, thanks for the heads up.

giovannicimolin (approved):
:+1:

- [x] I tested this:
1. Creating and modifying LineItem 
2. Token creation with new scopes
3. Endpoint permissions
4. Creating and modifying score 
5. Trying to create a score with invalid values (multiple tries, specifically trying to follow the spec).
5. Tested the results endpoint, using both filters (`limit` and `user_id`)
- [x] I read through the code
- [ ] Includes documentation: No :x:

@pcockwell Good to go, nice work! 

@nedbat This is ready for your review and certification testing.

nedbat:
This PR passes the IMS certification suite. I'm also reading the code....

lti_consumer/tests/unit/test_models.py
```diff
@@ -133,3 +133,39 @@ def test_repr(self):
...
             "block-v1:course+test+2020+type@problem+block@test - this-is-a-test"
         )
+
+
+class TestLtiAgsScoreModel(TestCase):
+    """
+    Unit tests for LtiAgsScore model methods.
+    """
+    def setUp(self):
+        super(TestLtiAgsScoreModel, self).setUp()
```
nedbat, line 143: nit: since we only run on Python 3 now, we can use `super().setUp()`

lti_consumer/tests/unit/plugin/test_views_lti_ags.py
```diff
@@ -274,3 +301,556 @@ def test_create_lineitem_invalid_resource_link_id(self):
...
+        )
+
+        # Request results with userId
+        response = self.client.get(results_user_endpoint, data={"userId": self.secondary_user_id})
+
+        self.assertEqual(response.status_code, 200)
+
+        # There should be 1 result for that user
+        self.assertEqual(len(response.data), 1)
+        self.assertEqual(response.data[0]['userId'], self.secondary_user_id)
```
nedbat, line 841: Don't we want to make an assertion about the result we got?

lti_consumer/tests/unit/plugin/test_views_lti_ags.py
```diff
@@ -274,3 +301,556 @@ def test_create_lineitem_invalid_resource_link_id(self):
...
+        self._set_lti_token('https://purl.imsglobal.org/spec/lti-ags/scope/result.readonly')
+
+        # Request results with limit
+        response = self.client.get(self.results_endpoint, data={"limit": 1})
+
+        self.assertEqual(response.status_code, 200)
+
+        # There should be 1 results, and it should be the one with the latest timestamp
+        self.assertEqual(len(response.data), 1)
+        self.assertEqual(response.data[0]['userId'], self.primary_user_id)
```
nedbat, line 856: This doesn't check everything the comment mentions: "the one with the latest timestamp".

nedbat (changes requested):
Two tests could use more asserts (I think?) and a few nits/

lti_consumer/tests/unit/plugin/test_views_lti_ags.py
pcockwell: @nedbat 

I'm not quite sure I understand what you mean. That line (841) makes an assertion that the result we received in the response, matches the user id requested. Combined with the check that there is only 1 result, that confirms that it filtered results with user ids that didn't match. There is already a test above that verifies the formatting of the individual results, and based on the records created during the `setUp()` for this class (see [here](https://github.com/edx/xblock-lti-consumer/blob/82af9de9e2509b9740cd45bfa376a3c6a837e616/lti_consumer/tests/unit/plugin/test_views_lti_ags.py#L711-L746)), there should only be 3 records, each with a unique user id. Thus, if the user id from the response matches the corresponding user id for the desired record (in this case it's `self.secondary_user_id`), then it's a sufficient match.

Should I add more assertions about the result and/or add comments to it?

lti_consumer/tests/unit/plugin/test_views_lti_ags.py
pcockwell: @nedbat 

Unfortunately, the results service does not return any timestamp information. As such, I can't make any assertions of that kind. I will add comments clarifying this though.

nedbat (approved):

PR #113, opened 2020-10-22
[BD-24] [TNL-7607] BB-3072: Move key and client management to model level.

giovannicimolin:
This PR removes the LTI 1.3 private key generation and storage from the XBlock settings.
The two main benefits are:
-  The public endpoint doesn't need to load the `modulestore` to retrieve the public keyset.
-  Keys are not stored in block settings anymore (and are automatically generated when missing, making key rotations easier).

This is the first step to allow LTI launches from Studio.

**Testing instructions:**
1. Checkout the `master` branch.
2. Set up and test a LTI Launch (make a note of the `client_id`).
3. Check out this branch on your devstack.
4. Apply migrations & restart LMS.
5. Go to Django admin and check that the `lti_1p3_client_id` and `_lti_1p3_platform_private_key_id` are filled with the `client_if` from the XBlock and `_lti_1p3_platform_private_key` is filled.
5. Run a LTI launch without changing any setting.

**Reviewers:**
- [ ] @shimulch or @pcockwell 
- [ ] @nedbat


edx-webhook:
Thanks for the pull request, @giovannicimolin! I've created [BLENDED-639](https://openedx.atlassian.net/browse/BLENDED-639) to keep track of it in Jira.   More details are on the [BD-24](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1545076784/BD-24+LTI+v1.3+Improvements+Advantage+Support) project page.


lti_consumer/plugin/views.py
```diff
@@ -44,14 +46,21 @@ def public_keyset_endpoint(request, usage_id=None):
...
-            request=request,
-            course_id=str(usage_key.course_key),
-            usage_id=str(usage_key),
-            handler='public_keyset_endpoint'
+        lti_config = LtiConfiguration.objects.get(
+            location=UsageKey.from_string(usage_id)
         )
+
+        if lti_config.version != lti_config.LTI_1P3:
+            raise Exception(
```
shimulch, line 54: @giovannicimolin  Shouldn't we use ``LtiError`` type here? https://github.com/edx/xblock-lti-consumer/blob/master/lti_consumer/exceptions.py#L6

lti_consumer/migrations/0004_keyset_mgmt_to_model.py
```diff
@@ -0,0 +1,33 @@
...
+        ),
+        migrations.AddField(
+            model_name='lticonfiguration',
+            name='_lti_1p3_platform_public_jwk',
+            field=models.TextField(blank=True, help_text="Platform's generated JWK keyset."),
+        ),
+        migrations.AddField(
+            model_name='lticonfiguration',
+            name='lti_1p3_client_id',
+            field=models.CharField(default='7bbf5517-07e5-43e7-af58-b800cd40dd40', help_text='Key ID used by LTI tool', max_length=255),
```
pcockwell, line 31: Why is a default being set here?

lti_consumer/migrations/0005_migrate_keyset_to_model.py
```diff
@@ -0,0 +1,62 @@
...
+            block = _load_block(lti_config.location)
+
+            # If client_id exists, move it to model
+            if block.lti_1p3_client_id:
+                lti_config.lti_1p3_client_id = block.lti_1p3_client_id
+
+            # If key exists, move it to model. Set kid as client_id, respecting
+            # old implementation
+            if block.lti_1p3_block_key:
+                lti_config._lti_1p3_platform_private_key_id = block.lti_1p3_client_id
```
pcockwell, line 35: Please confirm that you want to use `block.lti_1p3_client_id` as the value for this. You are checking above whether the block has the `lti_1p3_client_id` before setting it on the config, which means you expect it to be possible that it isn't present. You subsequently use it for the `_lti_1p3_platform_private_key_id` without verifying it is present.

lti_consumer/models.py
```diff
@@ -64,6 +68,29 @@ class LtiConfiguration(models.Model):
...
+    )
+
+    _lti_1p3_platform_public_jwk = models.TextField(
+        blank=True,
+        help_text="Platform's generated JWK keyset.",
+    )
+
+    lti_1p3_client_id = models.CharField(
+        max_length=255,
+        default=str(uuid.uuid4()),
```
pcockwell, line 90: I think you might want to use [`UUIDField`](https://docs.djangoproject.com/en/3.1/ref/models/fields/#uuidfield) instead here

lti_consumer/models.py
```diff
@@ -64,6 +68,29 @@ class LtiConfiguration(models.Model):
...
+        help_text="Key ID used by LTI tool",
```
pcockwell, line 91: Do you mean `Client ID`?

lti_consumer/plugin/views.py
pcockwell: I will second this

pcockwell (changes requested):

lti_consumer/migrations/0004_keyset_mgmt_to_model.py
giovannicimolin: My mistake, I'll remove this.

lti_consumer/migrations/0005_migrate_keyset_to_model.py
giovannicimolin: In the current `master` implementation, we're `client_id` and `key_id` are the same variable. I want to create a separate ID for each, but keep the old values for backwards compatibility. 

Does that address your concern?

>You subsequently use it for the _lti_1p3_platform_private_key_id without verifying it is present.

It was supposed to be a nested if. Will fix it :)

lti_consumer/models.py
giovannicimolin: @pcockwell I want to use a `string` here, like on the old implementation. I'm only using `uuid4` since it's an easy way to generate a random string.

giovannicimolin:
@pcockwell @shimulch I've addressed your review comments.

shimulch (approved):
LGTM :+1:

- [x] I tested this. Tested locally with existing and new LTI block.
- [x] I read through the code
- [x] Includes documentation

nedbat:
Where are we on this pull request? Is it waiting on me?

giovannicimolin:
@nedbat Looks like I need to rebase this. I'll do that and it'll be ready for review.

giovannicimolin:
@nedbat This is ready for your review now.

lti_consumer/api.py
```diff
@@ -52,3 +57,35 @@ def get_lti_consumer(config_id=None, block=None):
...
+def get_lti_1p3_launch_info(config_id=None, block=None):
+    """
+    Retrieves the Client ID, Keyset URL and other urls used to configure a LTI tool.
+    """
+    if not config_id and not block:
+        raise Exception("You need to pass either a config_id or a block.")
+
+    # Empty reponse
+    launch_info = {}
+
```
nedbat, line 71: There's no reason to have these lines: the only return comes after we create `launch_info` as a full dict.

lti_consumer/lti_xblock.py
```diff
@@ -304,26 +300,15 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
             "Basic Outcomes requests.</b>"
         ),
     )
-    # Client ID and block key
+    # DEPRECATED - These variables were moved to the LtiConfiguration Model
     lti_1p3_client_id = String(
-        display_name=_("LTI 1.3 Block Client ID"),
+        display_name=_("LTI 1.3 Block Client ID - DEPRECATED"),
```
nedbat, line 305: Can we fully delete these eventually?

lti_consumer/models.py
```diff
@@ -65,6 +76,29 @@ class LtiConfiguration(models.Model):
         blank=True,
     )
 
+    # LTI 1.3 Related variables
+    _lti_1p3_platform_private_key = models.TextField(
```
nedbat, line 80: I see why there's a leading underscore, because there's a property for accessing it.  But this will make the db column have a leading underscore... ick.

lti_consumer/plugin/views.py
```diff
@@ -44,15 +49,22 @@ def public_keyset_endpoint(request, usage_id=None):
...
+
+        if lti_config.version != lti_config.LTI_1P3:
+            raise LtiError(
+                "LTI Error: LTI 1.1 blocks do not have a public keyset endpoint."
+            )
+
+        # Retrieve block's Public JWK
+        # The underlying method will generate a new Private-Public Pair if one does
+        # not exist, and retrieve the values.
+        response = JsonResponse(json.loads(lti_config.lti_1p3_platform_public_jwk))
```
nedbat, line 64: Decoding and then re-encoding the JSON seems unfortunate, but I guess it's easier to use JsonResponse like this...

lti_consumer/tests/unit/plugin/test_views.py
```diff
@@ -34,11 +37,16 @@ def test_public_keyset_endpoint(self):
...
-        self.assertEqual(kwargs['handler'], 'public_keyset_endpoint')
+        self.assertEqual(response['Content-type'], 'application/json')
+        self.assertEqual(response['Content-Disposition'], 'attachment; filename=keyset.json')
+
+        # Check public keyset
+        self.lti_config.refresh_from_db()
+        retrieved_public_keyset = json.loads(response.content.decode('utf-8'))
+        self.assertEqual(
+            json.loads(self.lti_config.lti_1p3_platform_public_jwk),
+            retrieved_public_keyset
```
nedbat, line 48: Nit: Since you have to use json.loads on both values, it would be better to make them both inline calls here, just for readability.

nedbat (approved):
Only minor comments, thanks

lti_consumer/lti_xblock.py
giovannicimolin: @nedbat We can delete these fields after applying migration 0005 included in this PR.

lti_consumer/plugin/views.py
giovannicimolin: I've changed it a bit, now the model returns a proper python object that can be passed directly to `JsonResponse`

giovannicimolin:
@nedbat I've addressed your review comments. This is ready for review again.

nedbat:
Great, thanks!

PR #114, opened 2020-10-23
[BD-24][TNL-7647] BB-3128: Score linking ADR

giovannicimolin:
This PR introduces the Score linking decisions related to grading when using LTI 1.3 tools.

**Reviewers:**
- [ ] @shimulch
- [ ] @nedbat

### DIFF
--- a/docs/decisions/0003-lti-1p3-score-linking.rst
+++ b/docs/decisions/0003-lti-1p3-score-linking.rst
@@ -80,11 +80,11 @@ delete LineItems using the LineItem endpoint.
 A *post_save* Django signal in the *LtiAgsScore* should be responsible for loading the XBlock from the modulestore,
 bind the user to the session, and set the score (after doing the proper scaling using the `scoreMaximum` attribute).
 
-If a tool creates and links multiple LineItems to the same problem, the platform will ??? the results.
-
-???: Not sure what should be the behavior here:
-1. Link just the latest grade submitted by the tool.
-2. Average all items and submit average to block.
+If a tool creates and links multiple LineItems to the same problem, the platform will merge the results as follows:
+1. When a *Score* is pushed, the post save signal checks if the grade submitted is the final grade (by checking that *gradingProgress* is *FullyGraded*).
+2. The signal queries for other *LineItems* **linked to the problem** (with *LtiResourceLinkId* set) and checks if there's any fully graded *Score* for the user.
+3. With the just created *Score* plus the others previously submitted retrieved, the platform will normalize the values (using *scoreMaximum*) and average them.
+4. The signal will launch the XBlock handler, bind to the user and submit his grade.
 
 If a tool doesn't send any grades back or doesn't link any *resourceLinkId's* to a LineItem, the block will stay ungraded.
 
--- a/docs/decisions/0003-lti-1p3-score-linking.rst
+++ b/docs/decisions/0003-lti-1p3-score-linking.rst
@@ -78,7 +78,7 @@ No LineItems are created when the LTI configuration model is created, but the to
 delete LineItems using the LineItem endpoint.
 
 A *post_save* Django signal in the *LtiAgsScore* should be responsible for loading the XBlock from the modulestore,
-bind the user to the session, and set the score (after doing the proper scaling using the `scoreMaximum` attribute).
+bind the user to the session, and set the score (after doing the proper scaling using the *scoreMaximum* attribute).
 
 If a tool creates and links multiple LineItems to the same problem, the platform will merge the results as follows:
 1. When a *Score* is pushed, the post save signal checks if the grade submitted is the final grade (by checking that *gradingProgress* is *FullyGraded*).
--- a/docs/decisions/0003-lti-1p3-score-linking.rst
+++ b/docs/decisions/0003-lti-1p3-score-linking.rst
@@ -15,8 +15,7 @@ which don't map 1:1 with the grading and gradebook structure present in Open edX
 There's two models of interation to pushing grades to the platform in the LTI AGS services:
 
 1. Declarative: the platform creates a LineItem (equivalent of a gradebook line/grade) and tools can only push results to that item.
-2. Programmatic: the tool uses the AGS endpoints to manage it's own line items and grades. The tool is responsible for linking
-each line item to the resourceLinks, which means that a tool might not link a grade to it's respective problem.
+2. Programmatic: the tool uses the AGS endpoints to manage it's own line items and grades. The tool is responsible for linking each line item to the resourceLinks, which means that a tool might not link a grade to it's respective problem.
 
 See a more detailed description in the `Coupled vs decoupled line items`_ section of the spec.
 
--- a/docs/decisions/0003-lti-1p3-score-linking.rst
+++ b/docs/decisions/0003-lti-1p3-score-linking.rst
@@ -12,7 +12,7 @@ Context
 LTI Advantage provides new ways for LTI tools to push grades back into the platform through the `Assignment and Grades Services (AGS)`_,
 which don't map 1:1 with the grading and gradebook structure present in Open edX.
 
-There's two models of interation to pushing grades to the platform in the LTI AGS services:
+There's two models of interaction to pushing grades to the platform in the LTI AGS services:
 
 1. Declarative: the platform creates a LineItem (equivalent of a gradebook line/grade) and tools can only push results to that item.
 2. Programmatic: the tool uses the AGS endpoints to manage it's own line items and grades. The tool is responsible for linking each line item to the resourceLinks, which means that a tool might not link a grade to it's respective problem.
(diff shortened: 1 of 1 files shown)


edx-webhook:
Thanks for the pull request, @giovannicimolin! I've created [BLENDED-643](https://openedx.atlassian.net/browse/BLENDED-643) to keep track of it in Jira.   More details are on the [BD-24](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1545076784/BD-24+LTI+v1.3+Improvements+Advantage+Support) project page.


docs/decisions/0003-lti-1p3-score-linking.rst
```diff
@@ -0,0 +1,94 @@
...
+delete LineItems using the LineItem endpoint.
+
+A *post_save* Django signal in the *LtiAgsScore* should be responsible for loading the XBlock from the modulestore,
+bind the user to the session, and set the score (after doing the proper scaling using the `scoreMaximum` attribute).
+
+If a tool creates and links multiple problems to the same grade, the platform will ??? the results.
+
+???: Not sure what should be the behavior here:
+1. Link just the latest grade submitted by the tool.
+2. Average all items and submit average to block.
```
giovannicimolin, line 85: @nedbat This is my main concern regarding grade submission in the `programmatic` approach. Not sure what's the best option here. (1) is simpler, (2) doesn't leave any grading out of the final grade.

docs/decisions/0003-lti-1p3-score-linking.rst
```diff
@@ -0,0 +1,94 @@
...
+
+Programmatic grade handling
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+No LineItems are created when the LTI configuration model is created, but the tools have permissions to create, update and
+delete LineItems using the LineItem endpoint.
+
+A *post_save* Django signal in the *LtiAgsScore* should be responsible for loading the XBlock from the modulestore,
+bind the user to the session, and set the score (after doing the proper scaling using the `scoreMaximum` attribute).
+
+If a tool creates and links multiple problems to the same grade, the platform will ??? the results.
```
nedbat, line 81: You mean "multiple grades to the same problem", right?

docs/decisions/0003-lti-1p3-score-linking.rst
nedbat: IIUC, we are talking about multiple line items for the same problem, so conceptually, two different grades that are both "current" or "latest".  So I don't think #1 is an option: time-based "latest" isn't a way to choose between two grades that conceptually are side-by-side.  Or have I misunderstood?

nedbat (changes requested):

docs/decisions/0003-lti-1p3-score-linking.rst
giovannicimolin: @nedbat Yes! Fix:
```suggestion
If a tool creates and links multiple LineItems to the same problem, the platform will ??? the results.
```

docs/decisions/0003-lti-1p3-score-linking.rst
giovannicimolin: @nedbat I've assumed that most tools would just send a single grade, and in that case it won't be an issue.
I'll go with option 2 then, average all linked LineItems and push them the problem's grade.

nedbat: @marcotuts suggests that we only use "the primary grade."  Is there one that is clearly the primary?

pcockwell: If I read the [`Score Service` spec](https://www.imsglobal.org/spec/lti-ags/v2p0#timestamp) correctly, assuming a score is submitted for a specific `LineItem` with a `user_id` field that already has a score recorded for it, the new score should overwrite the existing one (up to and including setting the score to blank and/or deleting it) as long as the timestamp is greater than the one already recorded.

docs/decisions/0003-lti-1p3-score-linking.rst
```diff
@@ -0,0 +1,94 @@
...
+2. Average all items and submit average to block.
+
+If a tool doesn't send any grades back or doesn't link any *resourceLinkId's* to a LineItem, the block will stay ungraded.
+
+Consequences
+============
+
+1. This will make the platform LTI compliant and allow simpler grading workflows (if supported by tools).
+2. When using the programmatic approach, tools might not send a grade back, leaving students ungraded.
+3. Also when usint the programmatic approach, tools might send/link more than one grade for a given problem, and the criteria we're using to handle that is purely technical.
```
pcockwell, line 94: `usint` -> `using`

pcockwell (changes requested):

docs/decisions/0003-lti-1p3-score-linking.rst
giovannicimolin: @pcockwell The issue is when there are multiple `LineItems` for a single block. In `edx-platform` a block only accepts a single grade.

We're discussing averaging the values retrieved by the tool. 

I think that the main issue here is that we don't have a good enough sample of LTI 1.3 to decide on an approach.

giovannicimolin:
@pcockwell @nedbat I've fixed all the nits, and I'm just waiting on the outcome of the open discussion to push the final version.

docs/decisions/0003-lti-1p3-score-linking.rst
```diff
@@ -0,0 +1,96 @@
...
+
+Programmatic grade handling
+~~~~~~~~~~~~~~~~~~~~~~~~~~~
+No LineItems are created when the LTI configuration model is created, but the tools have permissions to create, update and
+delete LineItems using the LineItem endpoint.
+
+A *post_save* Django signal in the *LtiAgsScore* should be responsible for loading the XBlock from the modulestore,
+bind the user to the session, and set the score (after doing the proper scaling using the `scoreMaximum` attribute).
+
+If a tool creates and links multiple LineItems to the same problem, the platform will merge the results as follows:
```
giovannicimolin, line 83: @nedbat Final version here, averaging multiple `LineItems` if they exist.

docs/decisions/0003-lti-1p3-score-linking.rst
```diff
@@ -0,0 +1,96 @@
...
+No LineItems are created when the LTI configuration model is created, but the tools have permissions to create, update and
+delete LineItems using the LineItem endpoint.
+
+A *post_save* Django signal in the *LtiAgsScore* should be responsible for loading the XBlock from the modulestore,
+bind the user to the session, and set the score (after doing the proper scaling using the *scoreMaximum* attribute).
+
+If a tool creates and links multiple LineItems to the same problem, the platform will merge the results as follows:
+1. When a *Score* is pushed, the post save signal checks if the grade submitted is the final grade (by checking that *gradingProgress* is *FullyGraded*).
+2. The signal queries for other *LineItems* **linked to the problem** (with *LtiResourceLinkId* set) and checks if there's any fully graded *Score* for the user.
+3. With the just created *Score* plus the others previously submitted retrieved, the platform will normalize the values (using *scoreMaximum*) and average them.
```
nedbat, line 86: In step 1, does this mean nothing happens if the grade is not FullyGraded? That is, steps 2 and 3 only happen if the grade is FullyGraded.
In step 3, only FullyGraded values are averaged?

docs/decisions/0003-lti-1p3-score-linking.rst
```diff
@@ -0,0 +1,96 @@
...
+.. list-table::
+   :widths: auto
+   :header-rows: 1
+
+   * - Option name
+     - Behavior
+   * - Only allow tools to push a single grade per student (Default)
+     - This is the "declarative" approach, where the platform will create a single LineItem, and only allow read-only access to it.
+       The tool is only able to read LineItems, push scores and retrieve grades.
+   * - Allow tools to manage their own grades
```
nedbat, line 43: I'm wondering how to help the course author make this decision.   The spec says,
> [programmatic access] requires an elevated level of trust from the platform which might be cautious to which tools it grants POST on 'LineItems.url'.

I imagine tools will indicate that they need this access?  So the instructor instructions could be, "Only use this option if the tool requires it."  Are we worried about risk?  "elevated level of trust" implies that something could go wrong.  I suppose a malicious or badly written tool could send thousands of line items.  But it's very unlikely.

docs/decisions/0003-lti-1p3-score-linking.rst
```diff
@@ -0,0 +1,96 @@
...
+     - This is the "programmatic" approach, where tools have full permissions to create, edit and delete LineItems, as well as
+       pushing and retrieving grades. Note that there might be cases where a tool doesn't set a `resouceLinkId` leaving the grade
+       unlinked to a problem. Also, the other edge case is when a tool pushes multiple grades for a single problem, in which case
```
nedbat, line 46: "unlinked to a problem": is that a problem? What will happen?

docs/decisions/0003-lti-1p3-score-linking.rst
```diff
@@ -0,0 +1,96 @@
...
+       the implementation needs to decide on a criteria to merge the grades or pick one.
```
nedbat, line 47: You've decided on how to handle this case.

docs/decisions/0003-lti-1p3-score-linking.rst
```diff
@@ -0,0 +1,96 @@
...
+3. With the just created *Score* plus the others previously submitted retrieved, the platform will normalize the values (using *scoreMaximum*) and average them.
+4. The signal will launch the XBlock handler, bind to the user and submit his grade.
+
+If a tool doesn't send any grades back or doesn't link any *resourceLinkId's* to a LineItem, the block will stay ungraded.
+
+Consequences
+============
+
+1. This will make the platform LTI compliant and allow simpler grading workflows (if supported by tools).
+2. When using the programmatic approach, tools might not send a grade back, leaving students ungraded.
```
nedbat, line 95: Isn't it also true when using the declarative approach?

docs/decisions/0003-lti-1p3-score-linking.rst
nedbat: We should also note here that this is a choice we can revisit later.  In the fullness of time, based on demand, we could (hypothetically) provide the instructor with a choice of how to handle multiple grades.

giovannicimolin:
@nedbat As per our discussion from last week, I've removed the programmatic approach from the ADR.

nedbat (approved):

PR #117, opened 2020-11-05
[BD-24] [TNL-7661] [BB-3172] LTI Improvements - Load XBlock properly

giovannicimolin:
This PR fixes the block loading, previously we only loaded the descriptor, but to be able to bind back to a user to set grades, we need to load the block.

**Testing instructions:**
1. Checkout this branch.
2. Do a test launch, make sure it works.
3. Test rebinding a the block to a user. Run `make lms-shell`, then `./manage.py lms shell` and then run this code:
```
lti_config = LtiConfiguration.objects.get(pk=5)  # Adjust the PK here
user = User.objects.get(username='edx')

# This will retry the module and bind it to AnonymousUser
block = lti_config.block  

# This should work
block.runtime.rebind_noauth_module_to_user(block, user)
```

**Reviewers:**
- [ ] @nedbat 
- [ ] @shimulch

### DIFF
--- a/lti_consumer/models.py
+++ b/lti_consumer/models.py
@@ -11,6 +11,7 @@ from opaque_keys.edx.django.models import UsageKeyField
 from lti_consumer.lti_1p1.consumer import LtiConsumer1p1
 # LTI 1.3
 from lti_consumer.lti_1p3.consumer import LtiAdvantageConsumer
+from lti_consumer.plugin import compat
 from lti_consumer.utils import get_lms_base, get_lti_ags_lineitems_url
 
 
@@ -77,11 +78,7 @@ class LtiConfiguration(models.Model):
         if block is None:
             if self.location is None:
                 raise ValueError("Block location not set, it's not possible to retrieve the block.")
-
-            # Import on runtime only
-            # pylint: disable=import-outside-toplevel,import-error
-            from xmodule.modulestore.django import modulestore
-            block = self._block = modulestore().get_item(self.location)
+            block = self._block = compat.load_block_as_anonymous_user(self.location)
         return block
 
     @block.setter
--- a/lti_consumer/plugin/compat.py
+++ b/lti_consumer/plugin/compat.py
@@ -19,3 +19,33 @@ def run_xblock_handler_noauth(*args, **kwargs):
     # pylint: disable=import-error,import-outside-toplevel
     from lms.djangoapps.courseware.module_render import handle_xblock_callback_noauth
     return handle_xblock_callback_noauth(*args, **kwargs)
+
+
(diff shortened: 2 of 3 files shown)


edx-webhook:
Thanks for the pull request, @giovannicimolin! I've created [BLENDED-657](https://openedx.atlassian.net/browse/BLENDED-657) to keep track of it in Jira.   More details are on the [BD-24](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1545076784/BD-24+LTI+v1.3+Improvements+Advantage+Support) project page.


giovannicimolin:
@nedbat This is ready for review, and it's required to be able to set grades back into the XBlock. 
Related PR here, where @shimulch bumped into the issue: https://github.com/edx/xblock-lti-consumer/pull/116

shimulch (approved):
LGTM :+1:

- [x] I tested this on local devstack via manage.py shell
- [x] I read through the code
- [x] Includes documentation

ormsbee:
I haven't been paying close attention to the LTI extraction efforts, so please bear with me if the following question is off-base. If we've extracted a lot of the LTI logic out of the XBlock itself and into a Django app, is it possible to implement the grade passback feature without ever actually instantiating the XBlock/XBlock runtime at all? What do we really need from it?

I mostly ask because the user rebinding thing is a bit of deeper modulestore magic that I would love to just get rid of in this case if it is practical to do so.

giovannicimolin:
@ormsbee We haven't fully extracted the all the parameters from the LTI launches out of the XBlocks, **but** we only need the descriptors to retrieve all LTI-related settings.

There's a task in the epic to remove all dependencies of the LTI Launch from the XBlock, but it adds some complexity (showing fields on Studio that are not present in the XBlock requires a bit of work with templates, like it's done in the ORA repo).
 
>is it possible to implement the grade passback feature without ever actually instantiating the XBlock/XBlock runtime at all? What do we really need from it?

If there's a gradebook API available that would allow me to set the grade cleanly, I'd prefer using it. 

I've had to dig deep in the `modulestore` to figure out how to instantiate a module properly, but I agree that this is not a good use case of it's APIs. I went with this approach mostly because I didn't want to miss any sinals/methods that normally run when a grade is set through the XBlock and mess up the data somewhere along the way. 

- Do you have any recommendations when setting a grade? 
- Is there a gradebook API that allows me to set the grade _as a student_ and run all the signals/handlers attached to it? (and without instancing the block)

giovannicimolin:
@ormsbee We'll be using the `SCORE_PUBLISHED` signal and avoid rebinding, though I'd still like to keep this change and have a proper instance of the module returned to be used for LTI configuration).

PR #124, opened 2020-12-06
[BD-24] [TNL-7330] [BB-2726] LTI Improvements - Implement Context Membership Service

shimulch:
This PR Implements the Context Membership Service specified on LTI Names and Role Provisioning Services [spec](https://www.imsglobal.org/spec/lti-nrps/v2p0#context-membership).

The ADR that this implementation follows can be found at - https://github.com/edx/xblock-lti-consumer/blob/master/docs/decisions/0004-lti-advantage-nrps.rst

This PR depends on -
- https://github.com/edx/edx-platform/pull/25843

**JIRA tickets**: 
- https://tasks.opencraft.com/browse/BB-2726
- https://tasks.opencraft.com/browse/BB-4203
- https://openedx.atlassian.net/browse/TNL-7330

**Discussions**: N/A

**Dependencies**: None

**Screenshots**: 
Screenshot from 2020-12-06 19-59-24

**Sandbox URL**: None

**Merge deadline**: None

**Testing instructions**:

1. Pull this PR on local devstack. Pull https://github.com/edx/edx-platform/pull/25843 for edx-platform repository.
2. Make sure LTI 1.3 enabled by setting the feature flag on ``studio.yml``
```
FEATURES:
    LTI_1P3_ENABLED: true
    LTI_NRPS_ENABLED: true
```
3. Create an LTI 1.3 block. Follow the procedure from [here](https://github.com/edx/xblock-lti-consumer#lti-13).
4. Enable LTI NRPS from xblock settings as per the screenshot above.
5. Check that LTI Launches successfully and LTI NRPS claim is being shown.
6. Generate a JWT token. Go to www.jwt.io, select RS-256, fill in both the private and public keys section in the bottom right
Set the header and payload as follows
```
# HEADER:
{
  "alg": "RS256",
  "typ": "JWT",
  "kid": "123"
}
```
```
# PAYLOAD:
{
  "sub": "1234567890",
  "name": "John Doe",
  "admin": true,
  "iat": 1516239022
}
```
7. Get access token
```
curl --request POST \
  --url 'http://localhost:18000/api/lti_consumer/v1/token/<USAGE_KEY>' \
  --header 'Content-Type: application/x-www-form-urlencoded' \
  --data grant_type=client_credentials \
  --data scope=https://purl.imsglobal.org/spec/lti-nrps/scope/contextmembership.readonly \
  --data client_assertion_type=urn:ietf:params:oauth:client-assertion-type:jwt-bearer \
  --data client_assertion=<JWT_TOKEN>
```
8. Use the returned access token to perform the ``GET`` request to the Context Membership URL.
```
curl --request GET \
  --url http://localhost:18000/api/lti_consumer/v1/lti/<LTI_CONFIG_ID>/lti-nrps/memberships \
  --header 'Authorization: Bearer <ACCESS_TOKEN>'
```
9. You should see a response with the same schema defined it https://www.imsglobal.org/spec/lti-nrps/v2p0#context-membership
15. To enable exposing PII information, add a course waffle flag from ``http://localhost:18000/admin/waffle_utils/waffleflagcourseoverridemodel/``. 
16. Set waffle flag to - `lti_consumer.lti_nrps_transmit_pii` and set course key to your course key. ex: `course-v1:edX+DemoX+Demo_Course`.
17. Make sure you've chosen Override Choice - `Force On` and enabled it.
18. Now perform the request from step 8. You should see the name, email, and profile picture in the result now. 
19. To test if API correctly gates when the course has a large number of enrollments, set the `LTI_NRPS_ACTIVE_ENROLLMENT_LIMIT` Django setting to a lower value than the number of enrollments of your test course.
20. Then perform the request from step 8, you should see API now throws 404.

N/A

**Reviewers**
- [ ] @giovannicimolin 
- [ ] edX reviewer[s] TBD


edx-webhook:
Thanks for the pull request, @shimulch! I've created [BLENDED-677](https://openedx.atlassian.net/browse/BLENDED-677) to keep track of it in Jira.   More details are on the [BD-24](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1545076784/BD-24+LTI+v1.3+Improvements+Advantage+Support) project page.


lti_consumer/plugin/compat.py
```diff
@@ -94,3 +94,43 @@ def publish_grade(block, user, score, possible, only_if_higher=False, score_dele
...
+def get_user_profiles(users):
+    """
+    Given a list of user, returns corresponding user profiles
+    """
+    # pylint: disable=import-error,import-outside-toplevel
+    from student.models import UserProfile
+    return UserProfile.objects.filter(user__in=users)
+
+
+def get_external_ids(users):
```
giovannicimolin, line 118: @shimulch I'd prefer moving this to `edx-platform` and using `compat.py` merely as a mock isolation layer from the platform.

Do you think we can push this to the core platform (create a bulk get_or_create external ID method)?

lti_consumer/plugin/views.py
```diff
@@ -231,3 +233,68 @@ def scores(self, request, *args, **kwargs):
...
+
+        # get all course enrollments
+        enrollments = compat.get_user_enrollments(course_key)
+        enrolled_users = [enrol.user for enrol in enrollments]
+
+        # get user profiles in bulk
+        user_profiles = compat.get_user_profiles(enrolled_users)
+
+        # get external ids in bulk
+        user_eids = compat.get_external_ids(enrolled_users)
```
giovannicimolin, line 272: Instead of fetching everything all at once, I would only fetch what's requested by the API call.

Does the spec has any filtering per user_id/role/limit? If so, we should implement it, otherwise each NRPS API call is going to be costly.

giovannicimolin (changes requested):
@shimulch Left a few comments, I'll do a more thorough review tomorrow.

lti_consumer/plugin/views.py
shimulch: @giovannicimolin, Yes, there is three different thing needs to be implemented here. 
- Filtering via [role](https://www.imsglobal.org/spec/lti-nrps/v2p0#role-query-parameter)
- Pagination via [Next URL](https://www.imsglobal.org/spec/lti-nrps/v2p0#limit-query-parameter). It seems how we do pagination is up to us, we can use ID or some other param here.
- Filtering via [difference](https://www.imsglobal.org/spec/lti-nrps/v2p0#membership-differences)? 

My current implementation accommodates only Students. But the Spec also speaks about other roles. ``CourseEnrollment`` doesn't seem to have Teacher / Staff user linked to it. Is there any better Model? Otherwise, this pagination implementation will be tricky.

lti_consumer/plugin/compat.py
shimulch: I think we can push a class method to ``ExternalId`` that does this bulk get_or_create. Should I open a PR for that?

shimulch:
@giovannicimolin, I think this is ready for a proper review. :-)

lti_consumer/lti_1p3/constants.py
```diff
@@ -34,6 +34,32 @@
     ],
 }
 
+# Context membership roles
```
giovannicimolin, line 37: @shimulch We already have a role map (right above, `LTI_1P3_ROLE_MAP`).

For the context membership service, we don't need the "simple" version (it's just used in the `Resource Link Membership Service`). 

Can you remove this and simplify the current implementation?

lti_consumer/plugin/views.py
```diff
@@ -231,3 +240,153 @@ def scores(self, request, *args, **kwargs):
...
+        Get Page size for pagination
+        """
+        return settings.REST_FRAMEWORK.get('PAGE_SIZE', 10)
+
+    def build_queryset(self, filter_by=None):
+        """
+        Given a course key return a list of enrollments.
+        """
+        # pylint: disable=import-error,import-outside-toplevel,no-name-in-module
+        from common.djangoapps.student.models import CourseAccessRole
```
giovannicimolin, line 271: This is a platform import, it needs to be in `compat.py`.

It seems like we're using too many internal platform methods. It might be better to create a python API in the core that retrieves user data from a course (something like `get_enrolled_users(course_key, roles=None)`).

Do you know a good place to put this API in? I haven't looked at the platform code yet.

lti_consumer/plugin/views.py
```diff
@@ -231,3 +240,153 @@ def scores(self, request, *args, **kwargs):
...
+
+        course_key = self.request.lti_configuration.location.course_key
+
+        User = get_user_model()
```
giovannicimolin, line 275: I don't think we should use this method here. I prefer converting most of this core into a core platform API.

lti_consumer/plugin/views.py
```diff
@@ -231,3 +240,153 @@ def scores(self, request, *args, **kwargs):
...
+    def memberships(self, request, *args, **kwargs):
+        """
+        LTI NRPS Context Membership service endpoint.
+
+        See full API spec on -
+        https://www.imsglobal.org/spec/lti-nrps/v2p0#context-membership
+        """
+        role_filter_param = self.request.query_params.get('role', None)
+
+        # get current page number, make sure it's an integer
```
giovannicimolin, line 324: Instead of building the paginator mechanism like this, can we use a custom paginator class? 
Ref: https://www.django-rest-framework.org/api-guide/pagination/#custom-pagination-styles

giovannicimolin (changes requested):
@shimulch First review round is on.
The authentication is working perfectly, but I have a few concerns to querying user data from the LMS. We'll probably need to move a few things to an API in the core.

Feel free to discuss and reply to any comments. I haven't done a deep dive yet, so I might've missed a few things you've bumped on while implementing the API.

lti_consumer/models.py
```diff
@@ -247,6 +248,9 @@ def _get_lti_1p3_consumer(self):
                     lineitem_url=get_lti_ags_lineitems_url(self.id, lineitem.id),
                 )
 
+                if self.block.lti_1p3_enable_nrps:
+                    consumer.enable_nrps(get_lti_nrps_context_membership_url(self.id))
```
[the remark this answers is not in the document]
shimulch: Updated.

lti_consumer/lti_1p3/constants.py
shimulch: @giovannicimolin Are ``institution roles`` can be used interchangeably with ``context roles``? IMS seems to have separate vocabulary for both of them.

lti_consumer/plugin/views.py
shimulch: @giovannicimolin, Does ``openedx.core.djangoapps.enrollments.data`` seems a good place to move this function?

lti_consumer/plugin/compat.py
shimulch: @giovannicimolin, https://github.com/edx/edx-platform/pull/25844 has been opened for upstreaming this.

lti_consumer/plugin/views.py
shimulch: @giovannicimolin, The API now supports filtering and pagination.

lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
```diff
@@ -227,3 +229,70 @@ class Meta:
             'resultMaximum',
             'comment',
         )
+
+
+def create_lti_nrps_member_serializer():
+    """
+    Extend UserReadOnlySerializer and return extended Serializer class
+    """
+    UserReadOnlySerializer = compat.get_user_readonly_serializer()
```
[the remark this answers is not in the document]
shimulch: @giovannicimolin, This has been fixed. Now the response has the correct output.

lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
```diff
@@ -227,3 +229,70 @@ class Meta:
...
+
+    class ContextMemberSerializer(UserReadOnlySerializer):
```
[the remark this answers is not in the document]
shimulch: @giovannicimolin This serializer has been removed.

lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
```diff
@@ -227,3 +229,70 @@ class Meta:
...
+            externalid, _ = compat.get_or_create_externalid(user)
+            result['user_id'] = serializers.UUIDField().to_representation(externalid.external_user_id)
+
+            return result
+
+    return ContextMemberSerializer
+
+
+# pylint: disable=abstract-method
+class LtiNrpsContextSerializer(serializers.Serializer):
```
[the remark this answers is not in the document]
shimulch: Renamed

lti_consumer/plugin/compat.py
```diff
@@ -94,3 +94,21 @@ def publish_grade(block, user, score, possible, only_if_higher=False, score_dele
         score_deleted=score_deleted,
         grader_response=comment,
     )
+
+
+def get_user_readonly_serializer():
```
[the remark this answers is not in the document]
shimulch: Removed. There are two separate serializers now for PII and non-PII response.

lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
```diff
@@ -227,3 +229,70 @@ class Meta:
...
+                roles += [role['role'] for role in LTI_1P3_CONTEXT_ROLE_MAP['student']]
+
+            for accessrole in user.courseaccessrole_set.all():
+                if LTI_1P3_CONTEXT_ROLE_MAP.get(accessrole.role):
+                    roles += [role['role'] for role in LTI_1P3_CONTEXT_ROLE_MAP[accessrole.role]]
+
+            result['roles'] = set(roles)
+
+            # set user_id to external_user_id
+            externalid, _ = compat.get_or_create_externalid(user)
```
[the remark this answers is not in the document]
shimulch: Even though it will be an atomic operation, the number of queries will be large. I've opened a https://github.com/edx/edx-platform/pull/25844 for batch operation. If that gets approved we can use that here.

lti_consumer/plugin/views.py
shimulch: https://github.com/edx/edx-platform/pull/25843 has been opened to remove these.

lti_consumer/plugin/views.py
shimulch: https://github.com/edx/edx-platform/pull/25843 will remove this.

lti_consumer/plugin/views.py
shimulch: Yeah, added a custom paginator class and removed these portion.

lti_consumer/lti_1p3/constants.py
giovannicimolin: @shimulch I've missed that these are context roles as opposed of institution roles. You can keep this dict as is.

lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
```diff
@@ -227,3 +230,90 @@ class Meta:
...
+    """
+    status = serializers.SerializerMethodField()
+    user_id = serializers.SerializerMethodField()
+    roles = serializers.SerializerMethodField()
+
+    def get_user_id(self, user):
+        """
+        Get external user id for the user.
+        """
+        externalid, _ = compat.get_or_create_externalid(user)
```
giovannicimolin, line 255: Calling this here will not be optimal.

Ideally, the external ID should be prefetched/annotated in the queryset, and mapped using the `origin` parameter. I haven't tested this out to see if it works though.

lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
```diff
@@ -227,3 +230,90 @@ class Meta:
...
+        return serializers.UUIDField().to_representation(externalid.external_user_id)
+
+    def get_status(self, user):
```
giovannicimolin, line 258: The `status` field in the NRPS service means if the membership is active. Currently, we only fetch users that have active enrollments in the course, so the answer here should always be `True`.
We should not return inactive users through this endpoint.

lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
```diff
@@ -227,3 +230,90 @@ class Meta:
...
+    name = serializers.CharField(source='profile.name', required=False)
+    picture = serializers.SerializerMethodField()
+    email = serializers.EmailField(required=False)
+
+    def get_picture(self, user):
+        """
+        Get absolute url of user's profile picture.
+        """
+        request = self.context['request']
+        picture_url = compat.get_user_profile_image(user)['medium']
```
giovannicimolin, line 292: Again, this can cause a query "explosion", and needs to be prefetched in the queryset.
If there's no easy way, you can leave this parameter out for now.

lti_consumer/plugin/views.py
```diff
@@ -231,3 +235,80 @@ def scores(self, request, *args, **kwargs):
...
+        filter_by = self.parse_role_filter()
+
+        if filter_by is None:
+            # if no role filter given, we are interested in instructor & staff. other access
+            # roles like beta tester etc doesn't have equivalent role in LTI Context Role vocab.
+            kwargs['access_roles'] = ['instructor', 'staff']
+        elif filter_by != 'student':
+            kwargs['access_roles'] = [filter_by]
+            # exclude students if filtered by any access role
+            kwargs['include_students'] = False
```
giovannicimolin, line 297: Is there any use case where we want to exclude students? If so, can't we do it in the same way as staff or instructor?

giovannicimolin (changes requested):
Left a few comments in the PR

lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
shimulch: I will be using the batch external id to create PR. But for now, as we have opened two seperate PR on ``edx-platform``, this PR depends on one of them. I will remove this to use the batch method when ``edx-platform`` PR related to this gets updated.

lti_consumer/plugin/views.py
shimulch: AFAIK, in ``edx-platform``, ``student`` is not a ``CourseAccessRole``. As this API is part of ``edx-platform``, I thought this might make more sense to treat ``access_roles`` as ``CourseAccessRole`` and students as ``CourseEnrollment``. And the consumer of the API (lti_consumer xblock in this case) would build the required parameters themselves.

lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
shimulch: @giovannicimolin, I don't think it will cause a query "explosion" as the ``openedx.core.djangoapps.user_api.accounts.image_helpers.get_profile_image_urls_for_user`` doesn't do any query. It will use the image storage to retrieve image URLs and generate names using a hash function.

ormsbee:
@nedbat mentioned this PR in standup today. I realize that this is part of the LTI 1.3 NRPS spec and all, but I have serious concerns around how it would handle large courses with millions of enrolled students. At that point, even things like the pagination done here break down, since even things like `count` and `limit ... offset` get prohibitively slow and expensive. Even if that's made more performant with cursor pagination, we're still left with an API pattern that would be actively encouraging clients out there to make hundreds of thousands of successive requests, as quickly as they can, in order to crawl our enrollment data.

Stepping back, what is the use case we're trying to solve for in these cases? Is it still relevant to courses at MOOC scale? Will callers choke if they get back a million results? Can we gate these to specific types of users that those systems will care about more (e.g. you only get back verified enrollments, with no visibility into audit learners). Can we gate them so that we just return nothing at all when enrollments grow too large?

My apologies if this was already hashed over. If there's a forum post or PR I should read for more background, please let me know. Thank you.

ormsbee:
Another question: Can we live with a list of enrollments that's out of date? So something like S3 files that are generated nightly by celery tasks and just dumbly redirected to by this endpoint?

giovannicimolin:
@ormsbee @nedbat Those are valid concerns. I have some time scheduled to write an ADR this week, so I can try to explore a few solutions (such as course gating or limiting the feature to smaller courses - less than 1k active enrollments).

lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
```diff
@@ -374,3 +376,89 @@ class LtiDlImageSerializer(serializers.Serializer):
...
+        Get external user id for the user.
+        """
+        externalid, _ = compat.get_or_create_externalid(user)
+        return serializers.UUIDField().to_representation(externalid.external_user_id)
+
+    def get_status(self, user):
+        """
+        Get user status string value.
+        """
+        return 'Active' if user.is_active else 'Inactive'
```
giovannicimolin, line 407: @shimulch Exclude the inactive users from the API. The spec doesn't require us to return them, and that will be somewhat useless data.

lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
```diff
@@ -374,3 +376,89 @@ class LtiDlImageSerializer(serializers.Serializer):
...
+                roles += LTI_1P3_CONTEXT_ROLE_MAP[accessrole.role]
+
+        return set(roles)
+
+
+class LtiNrpsContextMemberPIISerializer(LtiNrpsContextMemberBasicSerializer):
+    """
+    Personally identifiable information serializer for Context Member.
+    """
+    name = serializers.CharField(source='profile.name', required=False)
```
giovannicimolin, line 428: @shimulch You're not prefetching the `profile` user model here. This will cause a query explosion, you need to prefetch the profile to avoid performance issues on big courses.

lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
giovannicimolin: @shimulch This method actually uses the `user.profile` related model, which will cause trouble if you're not prefetching `profile` in the queryset. Add a test that counts queries on this and you'll see what I mean. If there's more users, more queries will be performed.

lti_consumer/plugin/compat.py
```diff
@@ -121,3 +121,40 @@ def user_course_access(*args, **kwargs):
...
+def get_or_create_externalid(user):
+    """
+    Given a list of user, returns corresponding external id's
+    """
+    # pylint: disable=import-error,import-outside-toplevel
+    from openedx.core.djangoapps.external_user_ids.models import ExternalId
+    return ExternalId.add_new_user_id(user, 'lti')
+
+
+def get_course_members(*args, **kwargs):
```
giovannicimolin, line 135: @shimulch I think we're messing too much with internals in the core in a way that is not providing a nice reusable function. I'm considering a few alternatives:
1. Keep the current implementation in the core.
2. Move the course members queryset function inside the `lti_consumer` repo and simplify it.
What do you think is the best approach?

lti_consumer/plugin/urls.py
```diff
@@ -16,12 +16,15 @@
...
     deep_linking_content_endpoint,
+    # LTI NRPS URLs
+    LtiNrpsContextMembershipViewSet,
 )
 
 
 # LTI 1.3 APIs router
 router = routers.SimpleRouter(trailing_slash=False)
 router.register(r'lti-ags', LtiAgsLineItemViewset, basename='lti-ags-view')
+router.register(r'lti-nrps/memberships', LtiNrpsContextMembershipViewSet, basename='lti-nrps-memberships-view')
```
giovannicimolin, line 27: @shimulch You can simplify this to just `memberships` :)
```suggestion
router.register(r'memberships', LtiNrpsContextMembershipViewSet, basename='lti-nrps-memberships-view')
```

lti_consumer/plugin/views.py
```diff
@@ -415,3 +416,113 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+
+        matched_role = None
+        if role_filter_param:
+            for edx_role, context_roles in LTI_1P3_CONTEXT_ROLE_MAP.items():
+                if role_filter_param in context_roles:
+                    matched_role = edx_role
+
+        return matched_role
+
+    def queryset_helper(self, kwarg_overrides):
```
giovannicimolin, line 454: This might need to be changed depending on the direction of https://github.com/edx/edx-platform/pull/25843

giovannicimolin (changes requested):
@shimulch Left a few comments. I'm open to discuss some suggestions.
Overall the code works, but some bits are not optimized and might cause query explosions on large datasets, nothing that we can't fix with a few prefetches though.

lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
shimulch: Now it will return only `Active` as per the data api.

lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
shimulch: @giovannicimolin, since the data API now prefaces `profile`, this won't be a problem.

lti_consumer/plugin/compat.py
shimulch: @giovannicimolin, the core API now follows the edX guideline to return serialized data.

lti_consumer/plugin/views.py
shimulch: @giovannicimolin, yes this has been changed as the data API implementation got changed.

lti_consumer/plugin/views.py
```diff
@@ -415,3 +418,130 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+            # exclude students if filtered by any access role
+            kwargs['include_students'] = False
+
+        return self.get_memberships(kwargs)
+
+    def preprocess(self, data):
+        """
+        Preprocess the output of get_membership method.
+        For example - append external ids for user.
+        """
```
giovannicimolin, line 497: The only purpose of this is to add the external ids to the users, so rename this function accordingly.
```suggestion
    def attach_external_id_information(self, data):
        """
        A helper method that attaches the external id information to the user queryset.
        """
```

Also, add a better function docstring explaining the parameters and output.

lti_consumer/plugin/views.py
```diff
@@ -415,3 +418,130 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+
+        kwargs = {
+            'access_roles': [],
+            'include_students': True,
+        }
+
+        if kwarg_overrides and isinstance(kwarg_overrides, dict):
+            kwargs.update(kwarg_overrides)
+
+        return partial(compat.get_course_members, course_key, **kwargs)
```
giovannicimolin, line 471: This seems overly complicated. Following my suggestion of simplification in the other comment, we can completely remove this function in favor of only using `get_queryset`.

lti_consumer/plugin/views.py
```diff
@@ -415,3 +418,130 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+
+    def get_queryset(self):
+        """
+        Get User queryset related to current course. Also filter via role if provided.
+        """
+        kwargs = {}
```
giovannicimolin, line 477: Avoid using a generic parameter dict, create separate variables for each required parameter.
```
access_roles = []
include_students = False
limit = self.request.query_params.get('limit', None)
page = self.request.query_params.get('p', None)
course_key = ...

...

return attach_external_id_information(
    compat.get_course_members(...)
)
```

lti_consumer/plugin/views.py
```diff
@@ -415,3 +418,130 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+    # Custom permission classes for LTI APIs
+    authentication_classes = [Lti1p3ApiAuthentication]
+    permission_classes = [LtiNrpsContextMembershipsPermissions]
+
+    # Renderer classes to accept LTI NRPS content types
+    renderer_classes = [
+        MembershipResultRenderer,
+    ]
+
+    serializer_class = LtiNrpsContextMembershipSerializer
```
giovannicimolin, line 437: Pagination and the DRF request limit don't seem to be working for me. Can you test if the `PAGE_SIZE` and `limit` parameters work for you? They are not working for me.

I think it's because you're not passing the `limit` and page parameters to the LMS retrieval method.

lti_consumer/plugin/views.py
```diff
@@ -415,3 +418,130 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+    def list(self, *args, **kwargs):
+        """
+        Overrides default list method of ModelViewSet.
+        Checks if enrollment limit satisfies for current request.
+        """
+        members = self.get_memberships()(limit=1)
+        if members['count'] <= lti_nrps_enrollment_limit():
+            return super().list(*args, **kwargs)  # pylint: disable=no-member
+        else:
+            raise Http404
```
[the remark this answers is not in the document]
giovannicimolin: Additionally a `log.warning` here would be nice: `LTI NRPS Error: the tool can't retrieve more than %max_limit enrollments (tried %count)`.

lti_consumer/plugin/views.py
```diff
@@ -415,3 +418,130 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+        """
+        role_filter_param = self.request.query_params.get('role', None)
+
+        matched_role = None
+        if role_filter_param:
+            for edx_role, context_roles in LTI_1P3_CONTEXT_ROLE_MAP.items():
+                if role_filter_param in context_roles:
+                    matched_role = edx_role
+
+        return matched_role
```
giovannicimolin, line 454:
```suggestion
            for edx_role, context_roles in LTI_1P3_CONTEXT_ROLE_MAP.items():
                if role_filter_param in context_roles:
                    return edx_role
```

giovannicimolin (changes requested):
@shimulch Nice work here. Left quite a number of comments. 
The API is already working with a few minor caveats, but the code still warrants some simplification.

Let me know if you have any questions.

lti_consumer/lti_1p3/extensions/rest_framework/pagination.py
```diff
@@ -0,0 +1,88 @@
...
+"""
+from rest_framework.pagination import PageNumberPagination
+from rest_framework.response import Response
+
+
+class LinkHeaderPagination(PageNumberPagination):
+    """
+    Pagination via link header.
+    """
+
```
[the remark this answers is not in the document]
shimulch: Added. Thanks for pointing that out.

lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
```diff
@@ -374,3 +375,79 @@ class LtiDlImageSerializer(serializers.Serializer):
...
+    status = serializers.CharField(default='Active')
+    user_id = serializers.CharField(source='external_id')
+    roles = serializers.SerializerMethodField()
+
+    def get_roles(self, user):
+        """
+        Prepare and return Context Roles for user.
+        """
+        roles = []
+        if len(user['enrollments']) > 0:
```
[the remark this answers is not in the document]
shimulch: Updated

lti_consumer/plugin/views.py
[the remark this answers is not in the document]
shimulch: Added both log and 403 responses with the error message.

lti_consumer/plugin/views.py
```diff
@@ -415,3 +418,130 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+            'expose_pii_fields': expose_pii_fields(self.request.lti_configuration.location.course_key)
+        })
+        return context
+
+    def list(self, *args, **kwargs):
+        """
+        Overrides default list method of ModelViewSet.
+        Checks if enrollment limit satisfies for current request.
+        """
+        members = self.get_memberships()(limit=1)
```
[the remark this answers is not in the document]
shimulch: Moved the `get_course_membership` call and `lti_nrps_enrollment_limit` check to the custom pagination class. I know it's kind of against what DRF pagination suppose to do, but in this case, separating these concerns is hard. And in my opinion, there is no separate concern as `get_course_membership` already paginates for us.

lti_consumer/plugin/views.py
shimulch: Renamed and updated docstring.

lti_consumer/plugin/views.py
shimulch: All are removed. Moved this logic to the custom pagination class as I've mentioned [here](https://github.com/edx/xblock-lti-consumer/pull/124#discussion_r597808087).

lti_consumer/plugin/views.py
shimulch: Now using explicit variables rather than generic kwarg dict.

lti_consumer/plugin/views.py
shimulch: This has been fixed now. Thanks for reporting the issue.

lti_consumer/plugin/views.py
shimulch: Updated.

lti_consumer/lti_1p3/extensions/rest_framework/pagination.py
```diff
@@ -0,0 +1,158 @@
...
+        Parse role filter from query param and map it to edx role.
+        """
+        role_filter_param = self.request.query_params.get('role', None)
+
+        if role_filter_param:
+            for edx_role, context_roles in LTI_1P3_CONTEXT_ROLE_MAP.items():
+                if role_filter_param in context_roles:
+                    return edx_role
+
+        return None
```
giovannicimolin, line 87: This is the implicit return value. You can keep it here only if you add an explanation.
```suggestion

        # In case no filters match, bla bla....
        return None
```

shimulch: Added explanation.

lti_consumer/plugin/views.py
```diff
@@ -416,3 +425,196 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+
+    def get_serializer_context(self):
+        """
+        Adds additional contexts for serializers. Ex: if PII fileds can be exposed.
+        """
+        context = super().get_serializer_context()
+        context.update({
+            'expose_pii_fields': expose_pii_fields(self.request.lti_configuration.location.course_key)
+        })
+        return context
```
giovannicimolin, line 531: This is not needed anymore.
```suggestion
```

lti_consumer/plugin/views.py
```diff
@@ -416,3 +425,196 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+                    'id': course_key
+                },
+                'members': self.preprocess(data['result']),
+            }
+
+            serializer_cls = self.get_serializer_class()
+
+            serializer = serializer_cls(result)
+
+            return self.get_paginated_response(serializer.data)
```
giovannicimolin, line 604:
```suggestion
            # Serialize and return data NRPS reponse.
            serializer = self.get_serializer_class()(result)
            return self.get_paginated_response(serializer.data)
```

lti_consumer/plugin/views.py
```diff
@@ -416,3 +425,196 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+
+        # prepare datastructure supported by serializer
+        result = {
+            'id': self.request.build_absolute_uri(),
+            'context': {
+                'id': self.request.lti_configuration.location.course_key
+            },
+            'members': self.preprocess(data),
+        }
+        return super().get_serializer(result)
```
giovannicimolin, line 521: It doesn't seem like this is needed anymore, this is being done on the `list` method.

```suggestion
```

lti_consumer/plugin/views.py
```diff
@@ -416,3 +425,196 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+        if nrps_filter_role:
+            for edx_role, context_roles in LTI_1P3_CONTEXT_ROLE_MAP.items():
+                if nrps_filter_role in context_roles:
+                    return edx_role
+
+        # if no NRPS filter role provided or there was no match in LTI_1P3_CONTEXT_ROLE_MAP
+        return None
+
+    @staticmethod
+    def build_filter_kwargs(filter_by=None):
```
giovannicimolin, line 461:
```suggestion
    def get_filter_params(lti_role=None):
```

lti_consumer/plugin/views.py
```diff
@@ -416,3 +425,196 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+    # Renderer classes to accept LTI NRPS content types
+    renderer_classes = [
+        MembershipResultRenderer,
+    ]
+
+    # Custom pagination class to support pagination via Link Header
+    pagination_class = LinkHeaderPagination
+
+    @staticmethod
+    def parse_role_filter(nrps_filter_role=None):
```
giovannicimolin, line 448:
```suggestion
    def map_lti_to_lms_role(lti_role=None):
```

lti_consumer/plugin/views.py
```diff
@@ -416,3 +425,196 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+    def build_filter_kwargs(filter_by=None):
+        """
+        Helper method to build filter arguments for get_course_membership python API
+        Returns:
+            access_roles (list): A list of roles if there any, otherwise None
+            include_students (bool)
+        """
+        access_roles = None
+        include_students = True
+
```
giovannicimolin, line 470:
```suggestion
        role_filter = self.map_lti_to_lms_role(lti_role)

        access_roles = None
        include_students = True
```

lti_consumer/plugin/views.py
```diff
@@ -416,3 +425,196 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+        # we need to set the request object for paginator ourselves.
+        self.paginator.request = self.request
+
+        # get course key
+        course_key = self.request.lti_configuration.location.course_key
+
+        # parse and build filter parameters for `get_course_member` python API
+        nrps_filter_role = self.request.query_params.get('role', None)
+        filter_by = self.parse_role_filter(nrps_filter_role=nrps_filter_role)
+        access_roles, include_students = self.build_filter_kwargs(filter_by=filter_by)
```
giovannicimolin, line 559: These steps do not need to happen here. See this + my previous suggestions on the static methods:
```suggestion
        # Parse and build filter parameters for `get_course_member` python API
        access_roles, include_students = self.get_filter_params(
            self.request.query_params.get('role', None)
        )
```

giovannicimolin (changes requested):
There's still a few extra comments that I need to put here, but this is already in a good shape.
My last few nitpicks are around the paginator internals in the `list` method.
I'll experiment and leave comments here.

lti_consumer/plugin/views.py
shimulch: @giovannicimolin, since this is a static method we don't have `self` here. Do you want those as regular methods instead?

lti_consumer/plugin/views.py
shimulch: Sorry, I forgot to remove that. Removed now.

lti_consumer/plugin/views.py
shimulch: Simplified.

lti_consumer/plugin/views.py
shimulch: Sorry, I forgot to remove that. Removed now.

lti_consumer/plugin/views.py
shimulch: Changed

lti_consumer/plugin/views.py
shimulch: Changed

lti_consumer/plugin/views.py
shimulch: Simplified, but not exact. Due to my previous comment [here](https://github.com/edx/xblock-lti-consumer/pull/124#discussion_r598113187).

lti_consumer/plugin/views.py
```diff
@@ -416,3 +425,169 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+        filter_by = self.map_lti_to_lms_role(
+            self.request.query_params.get('role', None)
+        )
+        access_roles, include_students = self.get_filter_params(lti_role=filter_by)
+
+        # get the page size
+        page_size = self.paginator.get_page_size(self.request)
+
+        # get current page number
+        page_number = self.request.query_params.get(self.paginator.page_query_param, 1)
```
giovannicimolin, line 540:
```suggestion
        # Get the page size and number from parameters
        page_size = self.paginator.get_page_size(self.request)
        page_number = self.request.query_params.get(self.paginator.page_query_param, 1)
```

lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
```diff
@@ -374,3 +375,60 @@ class LtiDlImageSerializer(serializers.Serializer):
...
+            roles += LTI_1P3_CONTEXT_ROLE_MAP['student']
+
+        for accessrole in user['course_access_roles']:
+            if LTI_1P3_CONTEXT_ROLE_MAP.get(accessrole['role']):
+                roles += LTI_1P3_CONTEXT_ROLE_MAP[accessrole['role']]
```
giovannicimolin, line 405: This is not being covered by any tests.

lti_consumer/lti_1p3/consumer.py
```diff
@@ -492,6 +496,18 @@ def lti_ags(self):
 
         return self.ags
 
+    @property
+    def lti_nrps(self):
+        """
+        Returns LTI NRPS class or throw exception if not set up.
+        """
+        if not self.nrps:
+            raise exceptions.LtiAdvantageServiceNotSetUp(
```
giovannicimolin, line 505: This is not covered by tests.

lti_consumer/plugin/views.py
giovannicimolin: @shimulch Can you make only `get_filter_params` a regular method? You can leave the other one as static.

lti_consumer/plugin/views.py
```diff
@@ -416,3 +425,169 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+
+        try:
+
```
giovannicimolin, line 543:
```suggestion
```

lti_consumer/plugin/views.py
```diff
@@ -416,3 +425,169 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+                },
+                'members': self.preprocess(data['result']),
+            }
+
+            # Serialize and return data NRPS reponse.
+            serializer = self.get_serializer_class()(result)
+            return self.get_paginated_response(serializer.data)
+
+        except LtiNRPSEnrollmentLimitExceed as ex:
+
```
giovannicimolin, line 580:
```suggestion
```

lti_consumer/plugin/views.py
```diff
@@ -416,3 +425,169 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+            log.warning("LTI NRPS Error: %s", ex)
+            return Response({
+                "error": "above_response_limit",
+                "explanation": "The number of retrieved users is bigger than the maximum allowed in the configuration.",
+            }, status=HTTP_403_FORBIDDEN)
+
+        except InvalidPage as ex:
+
```
giovannicimolin, line 588:
```suggestion
```

lti_consumer/plugin/views.py
```diff
@@ -416,3 +425,169 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+            processed_data.append(user)
+
+        return processed_data
+
+    def get_serializer_class(self):
+        """
+        Overrides ModelViewSet's `get_serializer_class` method.
+        Checks if PII fields can be exposed and returns appropiate serializer.
+        """
+
```
giovannicimolin, line 512:
```suggestion
```

giovannicimolin (changes requested):
Last round, this is working as expected and the NRPS response is behaving correctly.
Here's a few nitpicks, then it'll be good to go.

lti_consumer/lti_1p3/consumer.py
shimulch: Covered.

lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
shimulch: Covered

lti_consumer/plugin/views.py
shimulch: Updated

shimulch:
@giovannicimolin, I've addressed your review comments.

lti_consumer/plugin/views.py
```diff
@@ -416,3 +425,163 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+                course_key,
+                access_roles=access_roles,
+                include_students=include_students,
+                page=page_number,
+                limit=page_size,
+            )
+
+            # check if total number of result is larger than specified limit
+            enrollment_limit = lti_nrps_enrollment_limit()
+            if data['count'] > lti_nrps_enrollment_limit():
```
giovannicimolin, line 550: Nit:
```suggestion
            if data['count'] > enrollment_limit:
```

lti_consumer/plugin/views.py
```diff
@@ -416,3 +425,163 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+        """
+        Maps given lti role to corresponding edx role.
+        """
+        if lti_role:
+            for edx_role, context_roles in LTI_1P3_CONTEXT_ROLE_MAP.items():
+                if lti_role in context_roles:
+                    return edx_role
+
+        # if no NRPS filter role provided or there was no match in LTI_1P3_CONTEXT_ROLE_MAP
+        return None
```
giovannicimolin, line 458: @shimulch This does not filter the queryset properly since it returns on the first match. 
See my output below:
```
{
  "id": "http://localhost:18000/api/lti_consumer/v1/lti/4/memberships?role=http%3A%2F%2Fpurl.imsglobal.org%2Fvocab%2Flis%2Fv2%2Fmembership",
  "context": {
    "id": "course-v1:OpenCraft+LTI202+2021"
  },
  "members": [
    {
      "status": "Active",
      "user_id": "a1752805-5281-437c-ae82-4f2430b3384c",
      "roles": [
        "http://purl.imsglobal.org/vocab/lis/v2/membership#Learner",
        "http://purl.imsglobal.org/vocab/lis/v2/membership#Administrator",
        "http://purl.imsglobal.org/vocab/lis/v2/membership#Instructor"
      ],
      "name": "full name test",
      "email": "edx@example.com"
    },
    {
      "status": "Active",
      "user_id": "bc73ed9f-10c5-4378-9ae9-14a57ed14592",
      "roles": [
        "http://purl.imsglobal.org/vocab/lis/v2/membership#Learner"
      ],
      "name": "",
      "email": "honor@example.com"
    }
  ]
}
```

The second entry should not have been returned in this case.
We need a new map from IMS to Open edX roles in the constants file.

giovannicimolin (changes requested):
Found a small issue in the role filter where a student matched in the admin filter. This needs to be fixed before we submit it for review.

lti_consumer/plugin/views.py
shimulch: @giovannicimolin, the `role` filter arg is not correct. `http://purl.imsglobal.org/vocab/lis/v2/membership` is not a valid role. From the [spec](https://www.imsglobal.org/spec/lti/v1p3/#lis-vocabulary-for-context-roles), the role will be one -

```
http://purl.imsglobal.org/vocab/lis/v2/membership#Administrator
http://purl.imsglobal.org/vocab/lis/v2/membership#ContentDeveloper
http://purl.imsglobal.org/vocab/lis/v2/membership#Instructor
http://purl.imsglobal.org/vocab/lis/v2/membership#Learner
http://purl.imsglobal.org/vocab/lis/v2/membership#Mentor
```

But yes there is a bug, the API should not have returned any member at all.

lti_consumer/plugin/views.py
shimulch: Thanks, updated.

lti_consumer/plugin/views.py
shimulch: @giovannicimolin, I've pushed a fix for this bug according to my previous comment. Also, I suggest to urlencode the role URI first before submitting the `GET` request. The `#` part gets discarded if not urlencoded.

giovannicimolin (approved):
:+1:

- [x] I tested this:
1. Checked that the NRPS API returns the response in the correct format, as described in the LTI spec.
2. Tested that the `filter`, `limit`, and `page` parameters work as expected.
3. Checked that the PII WaffleFlag is respected.
4. Tested the enrollment limit for the API.
5. Using `django-silk`, I've checked that there are no query "explosions" with 10, 100 and 500 users enrolled (always 16 or 17 queries per API call - including the LMS ones). 
- [x] I read through the code
- [x] I checked for accessibility issues: extra option added to block settings works ok.
- [x] Includes documentation: as docstrings.

@shimulch After countless rounds of testing, this is good to go.

Here is the performance testing data (using full serializer with PII data):
* 10 enrolled users:
image
* 500 enrolled users (`PAGE_SIZE=300`):
image

lti_consumer/plugin/views.py
```diff
@@ -416,3 +425,168 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+            else:
+                # exclude students if filtered by any access role
+                include_students = False
+                access_roles = [lti_role]
+
+        return access_roles, include_students
+
+    def preprocess(self, data):
+        """
+        Preprocess the output of get_membership method. It now does followings -
```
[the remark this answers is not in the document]
shimulch: Updated.

lti_consumer/plugin/views.py
```diff
@@ -416,3 +425,168 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
             status=status.HTTP_201_CREATED,
             headers=headers
         )
+
+
+class LtiNrpsContextMembershipViewSet(viewsets.ReadOnlyModelViewSet):
+    """
+    LTI NRPS Context Membership Service endpoint.
```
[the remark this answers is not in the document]
shimulch: Updated.

lti_consumer/plugin/views.py
```diff
@@ -420,3 +424,106 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+        profile_images = self.get_profile_images(users)
+
+        # import pdb; pdb.set_trace()
+        for userid, user_info in data.items():
+
+            # append external ids to user dictonary
+            user_info['external_id'] = external_ids[userid].external_user_id
+
+            if profile_images.get(userid):
+                user_info['picture'] = profile_images[userid]
```
giovannicimolin, line 483:
```suggestion
        users = get_user_model().objects.filter(id__in=user_ids)

        # get external ids
        external_ids = compat.batch_get_or_create_externalids(users)

        # import pdb; pdb.set_trace()
        for userid, user_info in data.items():

            # append external ids to user dictonary
            user_info['external_id'] = external_ids[userid].external_user_id
```

lti_consumer/plugin/views.py
```diff
@@ -420,3 +424,106 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+        """
+        Overrides default list method of ModelViewSet. Calls LMS `get_course_members`
+        API and returns result.
+        """
+
+        # get course key
+        course_key = self.request.lti_configuration.location.course_key
+
+        try:
+            data = compat.get_course_members(course_key)
```
giovannicimolin, line 509: @shimulch How are you paginating these results? This still needs to be paginated here to respect the LTI spec parameters.

shimulch: @giovannicimolin, I thought we've decided to dump all data at once since we only going to support a certain limit on enrollment. So this API returns all data at once.

I think it is respecting LTI spec for [pagination](https://www.imsglobal.org/spec/lti-nrps/v2p0#limit-query-parameter), as we are not sending `rel="next"` in the header, it means there is no data. And the pagination spec seems optional.

lti_consumer/plugin/views.py
shimulch: Removed.

lti_consumer/plugin/views.py
giovannicimolin: Makes sense, I've read the spec and checked that this is not a required parameter and that the LMS can ignore it.

lti_consumer/plugin/views.py
```diff
@@ -420,3 +424,87 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+
+        # import pdb; pdb.set_trace()
+        for userid, user_info in data.items():
+
+            # append external ids to user dictonary
+            user_info['external_id'] = external_ids[userid].external_user_id
+
+            processed_data.append(user_info)
+
+        return processed_data
```
giovannicimolin, line 468: Simplify function, rename it to a proper name and add a better description. 
My suggestion also avoids copying the dictionary (a potentially large dataset) - the current code doubles the execution time compared to it.

Suggestion: 
```suggestion
    def attach_external_user_ids(self, data):
        """
        Process the output of `get_membership` method and append external ids to each
        user in the object.
        """
        # batch get or create external ids for all users
        user_ids = data.keys()
        users = get_user_model().objects.filter(id__in=user_ids)
        external_ids = compat.batch_get_or_create_externalids(users)
       
        # Attaches external id information to the original dict
        for userid in data:
            data[userid]['external_id'] = external_ids[userid].external_user_id
```

lti_consumer/plugin/views.py
```diff
@@ -420,3 +424,87 @@ def scores(self, request, *args, **kwargs):  # pylint: disable=unused-argument
...
+            data = compat.get_course_members(course_key)
+
+            # build correct format for the serializer
+            result = {
+                'id': self.request.build_absolute_uri(),
+                'context': {
+                    'id': course_key
+                },
+                'members': self.preprocess(data),
+            }
```
giovannicimolin, line 499:
```suggestion
            # Attach external_id information to results
            self.attach_external_user_ids(data)
            
            # build correct format for the serializer
            result = {
                'id': self.request.build_absolute_uri(),
                'context': {
                    'id': course_key
                },
                'members': data,
            }
```

giovannicimolin (changes requested):
@shimulch I've left some comments to simplify the function and lower the resource footprint.
Other than that this is good to go :+1: 

I can leave a final :+1: once the nits are addressed.

lti_consumer/plugin/views.py
shimulch: Done.

lti_consumer/plugin/views.py
shimulch: Done!

shimulch:
@giovannicimolin Thanks for those performance tips! I have addressed your review comments.

giovannicimolin (approved):
:+1:

- [x] I tested this: 
1. Set up an LTI 1.3 test tool and enabled LTI-NRPS.
2. Tested the LTI NRPS endpoint and checked that all course enrollments were returned.
3. Tested the endpoint on a course with 500 enrollments and checked that the resource usage/queries were controlled.
4. Tested modifying `COURSE_MEMBER_API_ENROLLMENT_LIMIT` to 300 and checked that the API responded with an error:
```
{
  "error": "above_response_limit",
  "explanation": "The number of retrieved users is bigger than the maximum allowed in the configuration."
}
```
5. Compared the response format to the spec and checked all required fields are there.
- [x] I read through the code
- [x] Includes documentation

@shimulch Good to go.
@nedbat This is ready for review and conformance testing.

lti_consumer/lti_1p3/exceptions.py
```diff
@@ -57,5 +57,13 @@ class LtiAdvantageServiceNotSetUp(Lti1p3Exception):
     pass
 
 
+class LtiNRPSServiceNotSetUp(Lti1p3Exception):
```
nedbat, line 60: Super-nit: elsewhere we use "Nrps" as the word in the Python name.  Other initialisms in this file are all Camelcase.

lti_consumer/lti_1p3/extensions/rest_framework/permissions.py
```diff
@@ -7,7 +7,36 @@
 from rest_framework import permissions
 
 
-class LtiAgsPermissions(permissions.BasePermission):
+class LTIBasePermissions(permissions.BasePermission):
```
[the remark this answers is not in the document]
nedbat: s/LTI/Lti/ :)

lti_consumer/lti_1p3/consumer.py
```diff
@@ -488,6 +492,18 @@ def lti_ags(self):
 
         return self.ags
 
+    @property
+    def lti_nrps(self):
```
nedbat, line 496: This is unusual to have a property just to raise an exception if the attribute is None.  Does it simplify much code elsewhere?

lti_consumer/lti_1p3/tests/test_consumer.py
```diff
@@ -726,3 +727,37 @@ def test_set_dl_content_launch_parameters(self):
...
+            self.lti_consumer.lti_nrps  # pylint: disable=pointless-statement
+
+    def test_enable_nrps(self):
+        """
+        Test enabling LTI NRPS and checking that required parameters are set.
+        """
+        self.lti_consumer.enable_nrps("http://example.com/20/membership")
+
+        # Check that the NRPS class was properly instanced and set
+        self.assertEqual(type(self.lti_consumer.nrps), LtiNrps)
```
nedbat, line 745:
```suggestion
        self.assertIsInstance(self.lti_consumer.nrps, LtiNrps)
```

lti_consumer/lti_1p3/tests/test_consumer.py
```diff
@@ -726,3 +727,37 @@ def test_set_dl_content_launch_parameters(self):
...
+
+        # Check retrieving class works
+        lti_nrps_class = self.lti_consumer.lti_nrps
+        self.assertEqual(self.lti_consumer.nrps, lti_nrps_class)
```
nedbat, line 749: Is this testing an internal detail of the class?  Is "nrps" private or not?

lti_consumer/plugin/views.py
```diff
@@ -24,35 +26,37 @@
...
 )
-from lti_consumer.lti_1p3.extensions.rest_framework.permissions import LtiAgsPermissions
 from lti_consumer.lti_1p3.extensions.rest_framework.authentication import Lti1p3ApiAuthentication
 from lti_consumer.lti_1p3.extensions.rest_framework.renderers import (
     LineItemsRenderer,
     LineItemRenderer,
     LineItemScoreRenderer,
-    LineItemResultsRenderer
+    LineItemResultsRenderer,
+    MembershipResultRenderer,
```
nedbat, line 51: Thank you for the trailing commas! <3

lti_consumer/plugin/compat.py
```diff
@@ -129,3 +129,41 @@ def user_course_access(*args, **kwargs):
...
+    Returns a dict containing all users associated with the given course
+    """
+    # pylint: disable=import-error,import-outside-toplevel
+    from lms.djangoapps.course_api.api import get_course_members as core_get_course_members
+    from lms.djangoapps.course_api.exceptions import OverEnrollmentLimitException
+
+    try:
+        return core_get_course_members(course_key)
+    except OverEnrollmentLimitException as ex:
+        raise LtiError('NRPS is not available for this course!') from ex
```
nedbat, line 160: "this course" will be better if we include the course_key in the error message.

lti_consumer/tests/unit/plugin/test_views_lti_nrps.py
```diff
@@ -0,0 +1,317 @@
...
+
+        # name & email should not be exposed.
+        member_fields = response.data['members'][0].keys()
+        self.assertEqual(all([
+            'user_id' in member_fields,
+            'roles' in member_fields,
+            'status' in member_fields,
+            'email' not in member_fields,
+            'name' not in member_fields,
+        ]), True)
```
nedbat, line 270: This would be shorter, clearer, and more informative if you wrote out the asserts individually.

lti_consumer/utils.py
```diff
@@ -25,6 +26,24 @@ def lti_deeplinking_enabled():
     return settings.FEATURES.get('LTI_DEEP_LINKING_ENABLED', False) is True  # pragma: no cover
 
 
+def lti_nrps_enabled():
+    """
+    Returns `true` if LTI NRPS is enabled for instance.
+    """
+    return settings.FEATURES.get('LTI_NRPS_ENABLED', False) is True  # pragma: no cover
```
nedbat, line 33:
```suggestion
    return settings.FEATURES.get('LTI_NRPS_ENABLED', False)  # pragma: no cover
```

nedbat (changes requested):
I love that we are close to done! Just some small tweaks.

lti_consumer/tests/unit/plugin/test_views_lti_nrps.py
shimulch: Updated as requested.

lti_consumer/plugin/compat.py
shimulch: Added `course_key` in the log string.

lti_consumer/lti_1p3/tests/test_consumer.py
shimulch: `nrps` is not supposed to be used directly but through `lti_nrps`. This tests if `lti_nrps` returns the same object as `nrps` attribute.

lti_consumer/lti_1p3/exceptions.py
shimulch: Updated.

shimulch:
@nedbat Thanks for your review. I've addressed your review comments. Could you take a look? 

cc @giovannicimolin

nedbat (approved):

stvstnfrd (approved):
:+1: on behalf of @nedbat

sarina:
From Ned: `You can close that pull request with a comment that it’s on master through other means.` Congrats folx!

PR #112, opened 2020-10-20
[BD-24][TNL-7607] BB-3072: Partial Deep Linking Implementation 

giovannicimolin:
This PR adds the initial base code that allows performing LTI Deep Linking configuration launches.

This PR includes:
- Deep Linking related models.
- Support for Deep Link Launches (configuration) from LTI XBlock.
- Simplification of LTI launch flows, reducing the required number of launch steps.

_TODO:_
- [x] Full test coverage
- [x] Add missing token validation
- [x] Split into smaller diffs

**Break down in smaller PRs:**

I'm working on reducing the PR size by breaking this one into multiple smaller PRs. They need to be merged in the following order to avoid issues: 
1. Launch simplification PR: https://github.com/edx/xblock-lti-consumer/pull/125
2. Deep Linking Launch implementation: https://github.com/edx/xblock-lti-consumer/pull/126
3. Deep Linking Response endpoint implementation and `ltiResourceLink` support: https://github.com/edx/xblock-lti-consumer/pull/127

**Testing instructions:**
1. Check out this branch.
2. Apply migrations.
3. Enable `LTI_1P3_ENABLED` and `LTI_DEEP_LINKING_ENABLED` on `studio.yml` *FEATURES* dict.
4. Set up an LTI 1.3 tool and make sure to also copy over the `Tool Deep Link URL` from IMS's reference implementation.
5. Check that the normal LTI launch works.
6. Since there's no way of doing deep linking launches yet, apply a small patch to test the deep link launch:
```
diff --git a/lti_consumer/lti_xblock.py b/lti_consumer/lti_xblock.py
index df0cf16..3e5cadd 100644
--- a/lti_consumer/lti_xblock.py
+++ b/lti_consumer/lti_xblock.py
@@ -1378,7 +1378,7 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
             # Retrieve and set LTI 1.3 Launch start URL
             lti_block_launch_handler = get_lti_1p3_launch_start_url(
                 block=self,
-                deep_link_launch=False,
+                deep_link_launch=True,
                 hint=str(self.location)  # pylint: disable=no-member
             )
```
7. Perform the LTI Deep Link Launch, it should work the same way as the other launch.

**Reviewers:**
- [ ] @shimulch 
- [ ] @nedbat 

**Notes:**
I'm still working on this, and the diff size is likely to increase, should I start breaking this down into smaller PRs?
This is NOT a full deep linking implementation.

PR #125, opened 2020-12-07
[BD-24][TNL-7607] BB-3072: Simplify LTI 1.3 Launches

giovannicimolin:
Doing so by removing the OIDC view, which just rendered a form that performed a GET request.

**Testing instructions:**
1. Run a LTI Launch using this branch.
2. Check that the Launch is valid in the IMS tool.

**Reviewers:**
- [x] @shimulch 
- [x] @nedbat


edx-webhook:
Thanks for the pull request, @giovannicimolin! I've created [BLENDED-678](https://openedx.atlassian.net/browse/BLENDED-678) to keep track of it in Jira.   More details are on the [BD-24](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1545076784/BD-24+LTI+v1.3+Improvements+Advantage+Support) project page.


test_settings.py
```diff
@@ -11,4 +11,9 @@
...
 
 # LMS Urls - for LTI 1.3 testing
-LMS_ROOT_URL = "https://example.com"
\ No newline at end of file
+LMS_ROOT_URL = "https://example.com"
+
+# Dummy FEATURES dict
+FEATURES = {
+    'LTI_1P3_ENABLED': False,
+}
```
shimulch, line 19: @giovannicimolin, can you add a new line at the end of this file?

lti_consumer/api.py
```diff
@@ -53,36 +53,51 @@ def get_lti_consumer(config_id=None, block=None):
         # Since the block was passed, preload it to avoid
         # having to instance the modulestore and fetch it again.
         lti_config.block = block
+    else:
+        raise Exception('Either a config_id or block is required to get or create an LTI Configuration.')
```
shimulch, line 57: Maybe an ``LtiError`` would be better than the ``Exception`` class?

lti_consumer/api.py
```diff
@@ -36,12 +36,12 @@ def _get_or_create_local_lti_config(lti_version, block_location):
     return lti_config
 
 
-def get_lti_consumer(config_id=None, block=None):
+def _get_lti_config(config_id=None, block=None):
```
shimulch, line 39: Why are we defaulting to ``None`` for ``config_id`` and ``block``? Is there any use case for that?

lti_consumer/api.py
```diff
@@ -53,36 +53,51 @@ def get_lti_consumer(config_id=None, block=None):
...
 
 
 def get_lti_1p3_launch_info(config_id=None, block=None):
     """
     Retrieves the Client ID, Keyset URL and other urls used to configure a LTI tool.
     """
-    if not config_id and not block:
-        raise Exception("You need to pass either a config_id or a block.")
+    # Retrieve LTI Config
+    lti_config = _get_lti_config(config_id, block)
```
shimulch, line 79: @giovannicimolin, on previous implementation we could get launch info by passing either a ``config_id`` or a ``block``. But new implementation uses ``block`` just for [caching purposes](https://github.com/edx/xblock-lti-consumer/pull/125/files#diff-6e408862179502a0ac809fab914e028b18311a17a3b0874469bd6416bc4a449fR55). This changes old behavior. Is/will there be any case that we might only have ``block`` but not ``config_id`` to get launch info?

lti_consumer/lti_xblock.py
```diff
@@ -1357,10 +1337,20 @@ def _get_context_for_template(self):
...
-        lti_block_launch_handler = self.runtime.handler_url(self, 'lti_launch_handler').rstrip('/?')
-        if self.lti_version == 'lti_1p3':
-            lti_block_launch_handler = self.runtime.handler_url(self, 'lti_1p3_launch_handler').rstrip('/?')
+        if self.lti_version == 'lti_1p1':
+            # Set launch handler depending on LTI version
+            lti_block_launch_handler = self.runtime.handler_url(self, 'lti_launch_handler').rstrip('/?')
+        else:
+            # Runtime import since this will only run in the
+            # Open edX LMS/Studio environments.
+            from lti_consumer.api import get_lti_1p3_launch_start_url  # pylint: disable=import-outside-toplevel
```
shimulch, line 1346: If this import requires LMS/Studio environment, should we move this import to ``compat.py``?

shimulch (commented):
@giovannicimolin, I've tested the LTI launch and it works correctly. I have some comments on the code. Could you address those?

lti_consumer/lti_xblock.py
giovannicimolin: @shimulch This doesn't need any LMS API, but when run inside the XBlock's test environment it'll fail because workbench is not allowed to access Django imports, so any internal `api.py` calls from the XBlock needs to be imported at runtime. 

There's a similar usage of this in this file already, here: https://github.com/edx/xblock-lti-consumer/blob/master/lti_consumer/lti_xblock.py#L838

lti_consumer/api.py
giovannicimolin: @shimulch Because we can retrieve a configuration either using the `block` or the `config_id`, and there's no need to use both.

Eventually, we'll only use `config_id` but there's no full model support for it now, hence the current implementation.
I've just added an extra method to abstract re-usable code for the deep linking implementation.

> Is there any use case for that?

Yes, LTI-AGS and other advantage APIs rely on this feature.

lti_consumer/api.py
giovannicimolin: @shimulch The behavior is exactly the same as previously, and the only thing that changed is that I have pushed the following code into a reusable function (that will be used extensively).

I've moved this code to `get_lti_config`:
```
    if config_id:
        lti_config = LtiConfiguration.objects.get(pk=config_id)
    elif block:
        lti_config = _get_or_create_local_lti_config(
            block.lti_version,
            block.location,
        )
        # Since the block was passed, preload it to avoid
        # having to instance the modulestore and fetch it again.
        lti_config.block = block
```

giovannicimolin: @shimulch Missed your last question:

>Is/will there be any case that we might only have block but not config_id to get launch info?

Every time the API is called from whiting the XBlock, it will **only** pass the `block`, that's because there's no LTI config id storage in the LTI XBlock for now. LTI APIs only use LTI Config ID in the other hand.

We want to get rid of the Block requirement for XBlock launches, but that's on a future implementation.

shimulch (approved):
LGTM :+1:

- [x] I tested this on local devstack. LTI Launch works as expected.
- [x] I read through the code
- [x] Includes documentation

giovannicimolin:
Thanks @shimulch!

@nedbat This is ready for review. This is a part of the Deep Linking implementation from #112.

test_settings.py
```diff
@@ -11,4 +11,9 @@
 ROOT_URLCONF = 'test_urls'
 
 # LMS Urls - for LTI 1.3 testing
-LMS_ROOT_URL = "https://example.com"
\ No newline at end of file
+LMS_ROOT_URL = "https://example.com"
+
+# Dummy FEATURES dict
+FEATURES = {
+    'LTI_1P3_ENABLED': False,
```
nedbat, line 18: Just for my own curiosity: why did you need to add this, and why is it False?

README.rst
```diff
@@ -298,6 +298,11 @@ Please do not report security issues in public. Send security concerns via email
 Changelog
 =========
 
+2.4.1 - 2020-12-07
```
nedbat, line 301: Do we want to increment the version for each pull request?  There's no advantage to deploying these intermediate steps, right? (Other than finding out sooner if there was some bad interaction out in the real world).

nedbat (approved):

README.rst
giovannicimolin: @nedbat For this specific PR I think it's good to have a version bump.

For the LTI Deep Linking ones there's no need.

giovannicimolin:
@nedbat this is ready for review and merge.

test_settings.py
giovannicimolin: Instead of relying on the values that the code defaults to, I've decided to explicitly add it here.

PR #126, opened 2020-12-07
[BD-24][TNL-7607] BB-3072: Implement Deep Linking Launch flow.

giovannicimolin:
This PR partially implements the LTI Advantage Deep Linking Launch flow.
The LTI consumer and LTI Launch sequence is modified to handle messages that contain deep link requests and launch the tool using the proper endpoint and parameters.

**Split from https://github.com/edx/xblock-lti-consumer/pull/112.**

**Testing instructions:**
1. Check out this branch.
2. Set up a test LTI tool that supports Deep Linking (such as IMS's reference tool, https://github.com/IMSGlobal/lti-1-3-php-example-tool or https://github.com/dmitry-viskov/pylti1.3-django-example).
3. Enable LTI_1P3_ENABLED and LTI_DEEP_LINKING_ENABLED on studio.yml FEATURES dict.
4. Check that the normal LTI launch works.
5. Set the LTI Deep link launch url (from IMS's tool is https://lti-ri.imsglobal.org/lti/tools/TOOL_ID/deep_link_launches)
6. Since there's no way of doing deep linking launches yet, apply a small patch to test the deep link launch:
```
diff --git a/lti_consumer/lti_xblock.py b/lti_consumer/lti_xblock.py
index df0cf16..3e5cadd 100644
--- a/lti_consumer/lti_xblock.py
+++ b/lti_consumer/lti_xblock.py
@@ -1378,7 +1378,7 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
                 block=self,
             )
```
7. Perform the LTI Deep Link Launch (just do a normal LTI launc), it should work and land on a `Valid` page (if you're using IMS's tool) or in a 404 page in the LMS (if you're using a tool that does the entire deep linking flow).

Reviewers:
- [ ] @shimulch
- [ ] @nedbat

### DIFF
--- a/lti_consumer/lti_1p3/constants.py
+++ b/lti_consumer/lti_1p3/constants.py
@@ -52,6 +52,9 @@ LTI_1P3_ACCESS_TOKEN_SCOPES = [
 ]
 
 
+LTI_DEEP_LINKING_ACCEPTED_TYPES = []
+
+
 class LTI_1P3_CONTEXT_TYPE(Enum):  # pylint: disable=invalid-name
     """ LTI 1.3 Context Claim Types """
     group = 'http://purl.imsglobal.org/vocab/lis/v2/course#CourseGroup'
--- a/lti_consumer/tests/unit/test_lti_xblock.py
+++ b/lti_consumer/tests/unit/test_lti_xblock.py
@@ -432,6 +432,23 @@ class TestEditableFields(TestLtiConsumerXBlock):
         )
         lti_1p3_enabled_mock.assert_called()
 
+    @patch('lti_consumer.lti_xblock.lti_1p3_enabled', return_value=True)
+    @patch('lti_consumer.lti_xblock.lti_deeplinking_enabled', return_value=True)
+    def test_lti_deeplinking_fields_appear_when_enabled(self, lti_1p3_enabled_mock, lti_deeplinking_enabled_mock):
+        """
+        Test that LTI 1.3 XBlock's fields appear when `lti_1p3_enabled` returns True.
+        """
+        self.assertTrue(
+            self.are_fields_editable(
+                fields=[
+                    'lti_advantage_deep_linking_enabled',
+                    'lti_advantage_deep_linking_launch_url'
+                ]
+            )
+        )
+        lti_1p3_enabled_mock.assert_called()
+        lti_deeplinking_enabled_mock.assert_called()
+
 
 class TestGetLti1p1Consumer(TestLtiConsumerXBlock):
     """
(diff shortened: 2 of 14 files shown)


edx-webhook:
Thanks for the pull request, @giovannicimolin! I've created [BLENDED-679](https://openedx.atlassian.net/browse/BLENDED-679) to keep track of it in Jira.   More details are on the [BD-24](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1545076784/BD-24+LTI+v1.3+Improvements+Advantage+Support) project page.


giovannicimolin:
@shimulch This is ready for review. I can rebase once the launch simplification PR gets merged.

lti_consumer/lti_1p3/constants.py
```diff
@@ -52,6 +52,9 @@
 ]
 
 
+LTI_DEEP_LINKING_ACCEPTED_TYPES = []
```
shimulch, line 55: @giovannicimolin, I think you've missed adding ``ltiResourceLink`` in ``LTI_DEEP_LINKING_ACCEPTED_TYPES``. Deep Linking Launch fails without that.

shimulch (changes requested):
@giovannicimolin, can you check the review comment? Apart from this, all other things seem fine to me. I can approve as soon as you address that.

lti_consumer/lti_1p3/constants.py
giovannicimolin: @shimulch This was by design, the main PR has the required change, but adding it to this list meant that I would need to add the serializers and support for saving `ltiResourceLink` here.

The changes you are looking for are on https://github.com/edx/xblock-lti-consumer/pull/127, where I implement the models and the `ltiResourceLink` support.

shimulch: @giovannicimolin, I think that might be ok. As the whole reason for separate PRs is just to ease the review process and all three need to get merged for Deep linking to work anyway. It would've been nice if these PRs were self-sufficient without dependency, but that might make the splitting process really hard and time-consuming. 

You should add this to the test instruction, maybe on the patch section (step 6).

shimulch (approved):
LGTM :+1:

- [x] I tested this on local devstack. I've used both the IMS tool & https://github.com/dmitry-viskov/pylti1.3-django-example for testing LTI Deep Linking Launch. Works as expected.
- [x] I read through the code
- [x] Includes documentation

lti_consumer/tests/unit/test_lti_xblock.py
```diff
@@ -1329,6 +1329,40 @@ def test_launch_callback_endpoint_deep_linking(self):
...
+            "state=state_test_123&" +
+            "nonce=nonce&" +
+            "login_hint=oidchint&" +
+            "lti_message_hint=deep_linking_launch"
+        )
+
+        response = self.xblock.lti_1p3_launch_callback(request)
+
+        # Check response
+        self.assertEqual(response.status_code, 400)
```
shimulch, line 1364: Should we use ``403`` here instead? As we need a specific role for users to do an LTI Launch, ``403`` seems more appropriate to me. What do you think?

shimulch (changes requested):

lti_consumer/tests/unit/test_lti_xblock.py
giovannicimolin: @shimulch This is a student facing error page, not sure we should throw/display a permission denied error for a student.
Students will only hit this error if they are trying to bypass the `staff` requirement and make a Deep Link Launch without permissions.

giovannicimolin:
@nedbat Rebased the PR and checked that the LTI launches are working as expected. 

This is ready for review.

nedbat:
These changes were part of #127

PR #127, opened 2020-12-09
[BD-24][TNL-7607] BB-3072: Deep Linking response endpoint and model implementation 

giovannicimolin:
This PR implements the Deep Linking response endpoint, which tools use to send back information to the LMS after the user selects the content and configuration in the Tool's UI. 

Split from #112 

**Testing instructions:**
__Incomplete, tool configuration instructions TBD__
1. Checkout this branch.
2. Set up a local LTI tool (recommended: https://github.com/dmitry-viskov/pylti1.3-django-example) and ngrok tunnel.
3. Apply this patch to allow Deep Link launches:
```
diff --git a/lti_consumer/lti_xblock.py b/lti_consumer/lti_xblock.py
index df0cf16..3e5cadd 100644
--- a/lti_consumer/lti_xblock.py
+++ b/lti_consumer/lti_xblock.py
@@ -1378,7 +1378,7 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
                 block=self,
             )
```
4. Do a LTI launch (from the LMS).
5. Select an option in the LTI tool's UI.
6. Go to http://localhost:18000/admin/lti_consumer/ltidlcontentitem/ and check that the option you selected is stored in the model. Make note of the ID.
7. Do the launch again, and check that the old option was deleted and only the new one is stored.

**Author notes:**
This does not implement content presentation, just stores the data the user selected in the models.


edx-webhook:
Thanks for the pull request, @giovannicimolin! I've created [BLENDED-681](https://openedx.atlassian.net/browse/BLENDED-681) to keep track of it in Jira.   More details are on the [BD-24](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1545076784/BD-24+LTI+v1.3+Improvements+Advantage+Support) project page.


shimulch (approved):
LGTM :+1: 

- [x] I tested this in my local devstack. I've used the python LTI tool mentioned in the test instruction. LTI launch works correctly. And, clicking a DL link from the LTI Tool correctly redirects me to the DL response endpoint on LMS. I've checked, the DL response content is saved and gets replaced by a new one when performing the DL flow again.
- [x] I read through the code
- [x] Includes documentation

giovannicimolin:
@nedbat This is ready for review again. I've fixed the tests and added one extra test case to cover a few missing lines.

About logging, can I implement it in a separate follow-up PR? This is already pretty big.

lti_consumer/lti_1p3/consumer.py
```diff
@@ -493,3 +515,101 @@ def enable_ags(
...
+        if not message_type == "LtiDeepLinkingResponse":
+            raise exceptions.InvalidClaimValue("Token isn't a Deep Linking Response message.")
+
+        # Check if supported contentitems were returned
+        content_items = deep_link_response.get(
+            'https://purl.imsglobal.org/spec/lti-dl/claim/content_items',
+            # If not found, return empty list
+            [],
+        )
+        if any([
```
nedbat, line 608: Python nit: no need for square brackets here.  any() will work with a generator expression.

lti_consumer/lti_1p3/deep_linking.py
```diff
@@ -0,0 +1,75 @@
...
+        # Consctruct Deep Linking Claim
+        deep_linking_claim = {
+            "accept_types": accept_types_claim,
+            "accept_presentation_document_targets": [
+                "iframe",
+                "window",
+                "embed"
+            ],
+            # Only accept a single item return from Deep Linking operation.
+            "accept_multiple": True,
```
nedbat, line 56: This line doesn't match the comment before it.  Which is right?

lti_consumer/lti_1p3/deep_linking.py
```diff
@@ -0,0 +1,75 @@
...
+            # Automatically saves Content Items without asking to user
+            "auto_create": True,
+            # Other parameters
+            "title": title,
+            "text": description,
+            "deep_link_return_url": self.deep_linking_return_url
```
nedbat, line 62: You'll be glad later if you include trailing commas now :)

lti_consumer/lti_1p3/consumer.py
giovannicimolin: Fixed!

lti_consumer/lti_1p3/deep_linking.py
giovannicimolin: You're right, thanks for catching the issue.

lti_consumer/lti_1p3/deep_linking.py
giovannicimolin: Fixed. Is there a way we configure `pylint` or `pycodestyle` to catch these in the future?

nedbat:
I've tested this with the IMS Validator. It passes both the Core and the Deep Linking sections.

PR #128, opened 2020-12-14
[BD-24] [BB-3374] LTI Deep Linking Content Presentation Endpoint

shimulch:
This PR adds an endpoint for presenting/rendering Deep Linking Content. This PR depends on #127 

There are a couple of related PR's that should be reviewed and merged in the following order -

1. This PR
2. https://github.com/edx/xblock-lti-consumer/pull/131 - Implements ``link`` content type
3. https://github.com/edx/xblock-lti-consumer/pull/132 - Implements ``html`` content type
4. https://github.com/edx/xblock-lti-consumer/pull/133 - Implements ``image`` content type

**JIRA tickets**: 
https://tasks.opencraft.com/browse/BB-3374

~~**Discussions**:~~

**Dependencies**: None

~~**Screenshots**:~~

~~**Sandbox URL**:~~

**Merge deadline**: None 

**Testing instructions**:

1. Checkout this branch.
2. Set up a local LTI tool (recommended: https://github.com/dmitry-viskov/pylti1.3-django-example) and ngrok tunnel.
3. Change the following in ``lti_consumer/lti_xblock.py`` to allow Deep Link launches:
```
                 block=self,
             )
```
4. Do a LTI launch (from the LMS).
5. Select an option in the LTI tool's UI.
6. Visit - ``/api/lti_consumer/v1/lti/<LTI_CONFIG_ID>/lti-dl/content``. It should return a ``200`` response.

1. This PR only contains the implementation of the Content Presentation endpoint. The actual rendering of different content types will be on separate PRs.

**Reviewers**
- [ ] @farhaanbukhsh
- [ ] edX reviewer[s] TBD

### DIFF
--- a/lti_consumer/plugin/compat.py
+++ b/lti_consumer/plugin/compat.py
@@ -96,10 +96,28 @@ def publish_grade(block, user, score, possible, only_if_higher=False, score_dele
     )
 
 
-def user_has_staff_access(user, course_key):
+def user_has_access(*args, **kwargs):
     """
-    Check if an user has write permissions to a given course.
+    Import and run `has_access` from LMS
     """
     # pylint: disable=import-error,import-outside-toplevel
     from lms.djangoapps.courseware.access import has_access
-    return has_access(user, "staff", course_key)
+    return has_access(*args, **kwargs)
+
+
+def get_course_by_id(course_key):
+    """
+    Import and run `get_course_by_id` from LMS
+    """
+    # pylint: disable=import-error,import-outside-toplevel
+    from lms.djangoapps.courseware.courses import get_course_by_id as lms_get_course_by_id
+    return lms_get_course_by_id(course_key)
+
+
+def user_course_access(*args, **kwargs):
+    """
+    Import and run `check_course_access` from LMS
+    """
+    # pylint: disable=import-error,import-outside-toplevel
+    from lms.djangoapps.courseware.courses import check_course_access
+    return check_course_access(*args, **kwargs)
--- a/lti_consumer/plugin/views.py
+++ b/lti_consumer/plugin/views.py
@@ -7,6 +7,7 @@ from django.db import transaction
 from django.views.decorators.csrf import csrf_exempt
 from django.views.decorators.http import require_http_methods
 from django.views.decorators.clickjacking import xframe_options_sameorigin
+from django.shortcuts import render
 from django_filters.rest_framework import DjangoFilterBackend
 from opaque_keys import InvalidKeyError
 from opaque_keys.edx.keys import UsageKey
@@ -43,8 +44,46 @@ from lti_consumer.lti_1p3.extensions.rest_framework.parsers import (
 from lti_consumer.plugin.compat import (
     run_xblock_handler,
     run_xblock_handler_noauth,
-    user_has_staff_access,
+    get_course_by_id,
+    user_course_access,
+    user_has_access,
 )
+
+
+def user_has_staff_access(user, course_key):
+    """
+    Check if an user has write permissions to a given course.
+    """
+    return user_has_access(user, "staff", course_key)
+
+
+def has_block_access(user, block, course_key):
+    """
+    Checks if a user has access to given xblock.
+
+    ``has_access`` doesn't checks for course enrollment. On the otherhand, ``get_course_with_access``
+    only checks for the course itself. There is no way to check access for specific xblock. This function
+    has been created to perform a combination of check for both enrollment and access for specific xblock.
+
+    Args:
+        user: User Object
+        block: xblock Object to check permission for
+        course_key: A course_key specifying which course run this access is for.
+
+    Returns:
+        bool: True if user has access, False otherwise.
+    """
+    # Get the course
+    course = get_course_by_id(course_key)
+
+    # Check if user is authenticated & enrolled
+    course_access = user_course_access(course, user, 'load', check_if_enrolled=True, check_if_authenticated=True)
+
+    # Check if user has access to xblock
+    block_access = user_has_access(user, 'load', block, course_key)
+
+    # Return True if the user has access to xblock and is enrolled in that specific course.
+    return course_access and block_access
 
 
 @require_http_methods(["GET"])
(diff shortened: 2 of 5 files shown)


edx-webhook:
Thanks for the pull request, @shimulch! I've created [BLENDED-682](https://openedx.atlassian.net/browse/BLENDED-682) to keep track of it in Jira.   More details are on the [BD-24](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1545076784/BD-24+LTI+v1.3+Improvements+Advantage+Support) project page.


lti_consumer/plugin/compat.py
```diff
@@ -94,3 +94,12 @@ def publish_grade(block, user, score, possible, only_if_higher=False, score_dele
         score_deleted=score_deleted,
         grader_response=comment,
     )
+
+
+def has_access(*args, **kwargs):
+    """
+    Import and run the `has_access` method from courseware module.
```
farhaanbukhsh, line 101: I feel an example of how can this function be used for example what all parameters can be passed if we include that it will save a lof of time in future to find the ussage.

lti_consumer/plugin/views.py
```diff
@@ -175,6 +177,38 @@ def deep_linking_response_endpoint(request, lti_config_id=None):
...
+@require_http_methods(['GET'])
+def deep_linking_content_endpoint(request, lti_config_id=None):
+    """
+    Deep Linking endpoint for rendering Deep Linking Content Items.
+    """
+    try:
+        # Get LTI Configuration
+        lti_config = LtiConfiguration.objects.get(id=lti_config_id)
+
+        # This checks if user can perform LTI launch. If yes then user can also view LTI-DL content
```
farhaanbukhsh, line 189: Checks if user can perform LTI launch. If yes then user can also view LTI-DL content

lti_consumer/plugin/views.py
```diff
@@ -175,6 +177,38 @@ def deep_linking_response_endpoint(request, lti_config_id=None):
...
+        allowed = has_access(request.user, 'load', lti_config.block, lti_config.location.course_key)
+        if not allowed:
+            return HttpResponse(status=403)
+
+        # Get all LTI-DL contents
+        content_items = LtiDlContentItem.objects.filter(lti_configuration=lti_config)
+
+        # If no LTI-DL contents found for current configuration, throw 404 error
+        if content_items.count() == 0:
```
farhaanbukhsh, line 198: I feel doing a exists will be better than checking count 
https://stackoverflow.com/questions/6868939/in-django-what-is-the-most-efficient-way-to-check-for-an-empty-query-set

lti_consumer/tests/unit/plugin/test_views_lti_deep_linking.py
```diff
@@ -152,3 +152,52 @@ def test_lti_deep_linking_valid_request(self):
...
+    """
+
+    def setUp(self):
+        super().setUp()
+        self.url = '/lti_consumer/v1/lti/{}/lti-dl/content'.format(self.lti_config.id)
+
+    @patch('lti_consumer.plugin.views.has_access', return_value=False)
+    def test_forbidden_access(self, has_access_patcher):  # pylint: disable=unused-argument
+        """
+        Test if throws 403 when user doesn't have access.
```
farhaanbukhsh, line 169: Test if 403 is returned when a user doesn't have access.

farhaanbukhsh (changes requested):
These are probably some nit picks but it's good if you can have a look :)

lti_consumer/plugin/compat.py
shimulch: @farhaanbukhsh The ``compat.py`` is just an abstraction to help mock platform API calls during testing. This function doesn't contain any logic itself. I think it is better to force the user to check platform documentation. Otherwise, it will have duplicate documentation for the same function, and add the overhead of maintaining the docstring if anything changes in platform python API. You can check ``compat.py`` and see that most functions are doing the same thing, calling platform function.

lti_consumer/plugin/views.py
shimulch: Sounds better. Updated.

lti_consumer/plugin/views.py
shimulch: Yup, ``exists`` would be sufficient in this case.

lti_consumer/tests/unit/plugin/test_views_lti_deep_linking.py
shimulch: Updated. ``throws`` kind of means an exception, but in this case ``returned`` is a more appropriate word.

lti_consumer/plugin/compat.py
farhaanbukhsh: The point of maintaining the docstring seems legit.

farhaanbukhsh (changes requested):
The permission seems to have some issue, when I am creating a new user and trying to hit the endpoint with existing configuration it gives me a 200.

The new user is not enrolled in the course, this doesn't seem to be the right behaviour here can you have a look?

shimulch:
@farhaanbukhsh ``has_access`` doesn't check for course enrollment. On the other hand, ``get_course_with_access`` only checks for the course itself. There is no way to check access for specific xblock. 

A new function has been created to perform a combination of checks for both enrollment and access for specific xblock.

Thanks a lot for catching that. Can you check now if the new access function works correctly?

lti_consumer/plugin/compat.py
```diff
@@ -96,10 +96,34 @@ def publish_grade(block, user, score, possible, only_if_higher=False, score_dele
...
+    Returns:
+        bool: True if has access, False otherwise.
     """
     # pylint: disable=import-error,import-outside-toplevel
-    from lms.djangoapps.courseware.access import has_access as course_has_access
-    return course_has_access(*args, **kwargs)
+    from lms.djangoapps.courseware.courses import get_course_by_id, check_course_access
+    from lms.djangoapps.courseware.access import has_access
+
+    # get the course
```
farhaanbukhsh, line 119: Nit: Comments should start with caps keeping it consistent throughout

lti_consumer/plugin/compat.py
```diff
@@ -96,10 +96,34 @@ def publish_grade(block, user, score, possible, only_if_higher=False, score_dele
...
+    course = get_course_by_id(course_key)
+
+    # check if user is authenticated & enrolled
+    course_access = check_course_access(course, user, 'load', check_if_enrolled=True, check_if_authenticated=True)
+
+    # check if user has access to xblock
+    block_access = has_access(user, 'load', block, course_key)
+
+    # return True if have access for both course and xblock
```
farhaanbukhsh, line 128: Nit: Return True if the user has access to xblock and is enrolled in that specific course.

farhaanbukhsh (approved):
👍 

I have left few nits but  overall it looks decent, thanks for the work

- [x] I tested this: I was able to create the LTI configuration. Users who are enrolled in the course are able to access the configuration and users who are not enrolled in the course are not able to access the configuration.
- [x] I read through the code
- ~~[ ] I checked for accessibility issues~~
- [x] Includes documentation
- ~~[ ] I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository.~~

shimulch:
@farhaanbukhsh Nits are addressed. Thanks for the thorough review 👍🏽

lti_consumer/plugin/compat.py
```diff
@@ -94,3 +94,36 @@ def publish_grade(block, user, score, possible, only_if_higher=False, score_dele
         score_deleted=score_deleted,
         grader_response=comment,
     )
+
+
+def user_has_access(user, block, course_key):
```
giovannicimolin, line 99: @shimulch In the merged deep linking implementation, I already make use of the `has_access` method. You'll have to rework it a bit to reuse the code.
Also, break this down into two smaller functions, each only running a small piece of LMS code. This is meant to be a compatibility layer, so no logic is to be done here.

shimulch: @giovannicimolin, I've refactored the ``compat.py`` and added a couple of smaller functions to keep logic out of it. Then added reusable functions in ``views.py`` so that we can make use of those functions inside different views. Let me know if you think the current approach is ok :-)

nedbat (approved):

PR #129, opened 2020-12-17
Skip migration when no LTI XBlock location exists

stvstnfrd:
otherwise this blows up, trying to migrate a record with no location
(the modulestore lookup throws).

Presumably this isn't (wasn't?) an issue in prod already, right?
But I did hit this in my devstack, having previously created an
LtiConfiguration record via Django Admin (without a location).

### DIFF
--- a/lti_consumer/migrations/0005_migrate_keyset_to_model.py
+++ b/lti_consumer/migrations/0005_migrate_keyset_to_model.py
@@ -23,6 +23,8 @@ def forwards_func(apps, schema_editor):
     )
     with transaction.atomic():
         for lti_config in lti_configs:
+            if not lti_config.location:
+                continue
             block = _load_block(lti_config.location)
 
             # If client_id exists, move it to model


stvstnfrd:
@giovannicimolin I hit this running migrations in my devstack. Presumably an edge case; I'm guessing the `location` typically gets set via the normal flows, but not here since I manually created it via the Django Admin and did not provide a location.

We probably don't need to cut a new release/tag just for this (and I have fixed my local devstack), but it probably doesn't hurt to include this on master. Let me know what you think and if I'm missing anything. Thanks!

  (CC: @xitij2000 @davidjoy )

giovannicimolin:
@stvstnfrd This only happens because you are creating the LTI config through the admin, which is a suitable use case (just not one supported right now).`LtiConfiguration` objects were only creating using a python API before (see `lti_consumer/api.py`), so this edge case would never happen. 

The migration is already applied on edX.org, but it won't hurt to add this change here since this is just a data migration.

giovannicimolin (approved):
:+1: 

- [ ] ~~I tested this~~ Not tested in devstack 
- [x] I read through the code: LGTM, the patch makes sense and fixes an edge case when using `LtiConfiguration` outside the scope of XBlocks.

@stvstnfrd Thanks for spotting this! I missed it :)

FYI @nedbat

PR #130, opened 2020-12-21
[BD-03] [TNL-7742] [BB-3370] Add LTI 1.1 config on model

xitij2000:
Add support for storing LTI1.1 configuration in the database.


edx-webhook:
Thanks for the pull request, @xitij2000! I've created [BLENDED-688](https://openedx.atlassian.net/browse/BLENDED-688) to keep track of it in Jira.


lti_consumer/models.py
```diff
@@ -60,52 +62,89 @@ class LtiConfiguration(models.Model):
...
+        (CONFIG_ON_XBLOCK, _('Configuration Stored on XBlock fields')),
+        (CONFIG_ON_DB, _('Configuration Stored on this model')),
     ]
     config_store = models.CharField(
         max_length=255,
         choices=CONFIG_STORE_CHOICES,
         default=CONFIG_ON_XBLOCK,
     )
 
+    # A secondary ID for this configuration that can be used in URLs without leaking primary id.
```
stvstnfrd, line 76: What is the primary ID now?

lti_consumer/migrations/0006_add_on_model_config_for_lti_1p1.py
```diff
@@ -0,0 +1,46 @@
...
+        ),
+        migrations.AddField(
+            model_name='lticonfiguration',
+            name='lti_config',
+            field=jsonfield.fields.JSONField(blank=True, default=dict, dump_kwargs={'cls': jsonfield.encoder.JSONEncoder, 'separators': (',', ':')}, help_text='LTI configuration data.', load_kwargs={}),
+        ),
+        migrations.AlterField(
+            model_name='lticonfiguration',
+            name='config_store',
+            field=models.CharField(choices=[('CONFIG_ON_XBLOCK', 'Configuration Stored on XBlock fields'), ('CONFIG_ON_DB', 'Configuration Stored on this model')], default='CONFIG_ON_XBLOCK', max_length=255),
```
stvstnfrd, line 44: Probably not a big deal, but I wonder if the default should now be `CONFIG_ON_DB`..

stvstnfrd (commented):
I don't know LTI particularly well, but this seems reasonable from my understanding of what we need on the discussions side.

I defer to someone that knows LTI better, though.

lti_consumer/models.py
xitij2000: The original default auto-generated ID is till the primary ID.

lti_consumer/migrations/0006_add_on_model_config_for_lti_1p1.py
xitij2000: The config on db path isn't fully developed, and doesn't support LTI1.3, so until all the paths for that are supported I think we should default to config on XBlock.

xitij2000:
@stvstnfrd I upgraded the dependencies to pull in the latest version of edx-lint and that raised a lot of quality warnings. I tried to fix the dependencies but it was a bit fragile so I decided to instead use that time fixiing the quality issues. Those are in a separate commit to make it easier to review the actual content of the PR.

lti_consumer/tests/unit/test_models.py
```diff
@@ -90,6 +113,7 @@ def test_lti_consumer_ags_enabled(self):
 
         # Get LTI 1.3 consumer
         consumer = self.lti_1p3_config.get_lti_consumer()
+        self.maxDiff = None
 
         # Check that LTI claim was included in extra claims
         self.assertEqual(
```
shimulch, line 119: You need to update LTI configuration id to ``3`` for the LTI AGS test [here](https://github.com/open-craft/xblock-lti-consumer/blob/kshitij/lti-on-model-config-lti1.1/lti_consumer/tests/unit/test_models.py#L129). Because your change on this PR creates a new LTI configuration and ``lti_1p3_config`` is the last one. I think this is why the test is failing.

shimulch (changes requested):

lti_consumer/tests/unit/test_models.py
xitij2000: Done.

shimulch (approved):
LGTM :+1:

- [x] I tested this on the sandbox attached with https://github.com/edx/edx-platform/pull/25910. That PR has a dependency on this PR. I've created LTI 1.1 configuration from Django Admin. It works fine.
- [x] I read through the code
- [x] Includes documentation

stvstnfrd (approved):
LGTM

sarina:
My understanding is this needs a rebase, then we can merge?

stvstnfrd:
@xitij2000 Looks like this just needs squashed/rebased and then is good to go?
Let me know if you/open-craft don't have the merge bit here, and I can hit the button :+1:

xitij2000:
@stvstnfrd I had already rebased it into 2 commits, I've now brought it down to 1. 

I don't think anyone at OC has merge rights here, so you'll need to do the merging.

PR #133, opened 2020-12-29
[BD-24] [BB-3374] LTI Deep Linking Content Presentation - image

shimulch:
Depends on https://github.com/edx/xblock-lti-consumer/pull/132

This PR implements the ``image`` content type for LTI Deep Linking content presentation. Spec - http://www.imsglobal.org/spec/lti-dl/v2p0#image

**JIRA tickets**: https://tasks.opencraft.com/browse/BB-3374

**Discussions**: N/A

**Dependencies**: None

**Screenshots**:
Screenshot from 2020-12-29 23-05-38

**Sandbox URL**: N/A

**Merge deadline**: None

**Testing instructions**:
1. Assuming you have already tested https://github.com/edx/xblock-lti-consumer/pull/128 in your local devstack, because this PR depends on that.
2. Go to ``http://localhost:18000/admin/lti_consumer/ltidlcontentitem``
3. Create a new item, select the LTI configuration, from the content type select ``Image``. Put following in the attributes -
```
{"url":"https://via.placeholder.com/400x200","title":"Present Image","text":"Sample Description","width":"400px","height":"200px"}
```
4. Visit - /api/lti_consumer/v1/lti/<LTI_CONFIG_ID>/lti-dl/content. It should render the Image with defined ``width`` and ``height``.

1.  I think it reasonable to add ``icon`` and ``thumbnail`` as hyperlinked to the actual image. Clicking on ``icon`` and ``thumbnail`` will open the actual image. I also didn't find any reason to add both of them at the same time. So ``thumbnail`` will take precedence over ``icon`` if both exist and only one will show.

**Reviewers**
- [ ] @farhaanbukhsh
- [ ] edX reviewer[s] TBD

### DIFF
--- a/lti_consumer/templates/html/lti-dl/render_image.html
+++ b/lti_consumer/templates/html/lti-dl/render_image.html
@@ -1,0 +1,14 @@
+{% if attrs.title %}<h2>{{ attrs.title }}</h2>{% endif %}
+{% if attrs.text %}<p>{{ attrs.text }}</p>{% endif %}
+
+{% if attrs.thumbnail %}
+<a href="{{ attrs.url }}" {% if attrs.title %}alt="{{ attrs.title }}"{% endif %}>
+    <img src="{{ attrs.thumbnail.url }}" {% if attrs.title %}alt="{{ attrs.title }}"{% endif %} width="{{ attrs.thumbnail.width }}" height="{{ attrs.thumbnail.height }}" />
+</a>
+{% elif attrs.icon %}
+<a href="{{ attrs.url }}" {% if attrs.title %}alt="{{ attrs.title }}"{% endif %}>
+    <img src="{{ attrs.icon.url }}" {% if attrs.title %}alt="{{ attrs.title }}"{% endif %} width="{{ attrs.icon.width }}" height="{{ attrs.icon.height }}" />
+</a>
+{% else %}
+<img src="{{ attrs.url }}" {% if attrs.title %}alt="{{ attrs.title }}" {% endif %}{% if attrs.width %}width="{{ attrs.width }}" {% endif %}{% if attrs.height %}height="{{ attrs.height }}"{% endif %} />
+{% endif %}
--- a/lti_consumer/lti_1p3/constants.py
+++ b/lti_consumer/lti_1p3/constants.py
@@ -56,7 +56,7 @@ LTI_DEEP_LINKING_ACCEPTED_TYPES = [
     'ltiResourceLink',
     'link',
     'html',
-    # TODO: implement "image" support,
+    'image',
     # TODO: implement "file" support,
 ]
 
--- a/lti_consumer/lti_1p3/constants.py
+++ b/lti_consumer/lti_1p3/constants.py
@@ -54,8 +54,8 @@ LTI_1P3_ACCESS_TOKEN_SCOPES = [
 
 LTI_DEEP_LINKING_ACCEPTED_TYPES = [
     'ltiResourceLink',
+    'link',
     # TODO: implement "image" support,
-    # TODO: implement "link" support,
     # TODO: implement "file" support,
     # TODO: implement "html" support,
 ]
--- a/lti_consumer/lti_1p3/constants.py
+++ b/lti_consumer/lti_1p3/constants.py
@@ -55,9 +55,9 @@ LTI_1P3_ACCESS_TOKEN_SCOPES = [
 LTI_DEEP_LINKING_ACCEPTED_TYPES = [
     'ltiResourceLink',
     'link',
+    'html',
     # TODO: implement "image" support,
     # TODO: implement "file" support,
-    # TODO: implement "html" support,
 ]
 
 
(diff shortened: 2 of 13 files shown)


edx-webhook:
Thanks for the pull request, @shimulch! I've created [BLENDED-692](https://openedx.atlassian.net/browse/BLENDED-692) to keep track of it in Jira.   More details are on the [BD-24](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1545076784/BD-24+LTI+v1.3+Improvements+Advantage+Support) project page.


lti_consumer/templates/html/lti-dl/render_image.html
```diff
@@ -0,0 +1,4 @@
+{% if attrs.title %}<h2>{{ attrs.title }}</h2>{% endif %}
+{% if attrs.text %}<p>{{ attrs.text }}</p>{% endif %}
+
+<img src="{{ attrs.url }}" {% if attrs.width %}width="{{ attrs.width }}"{% endif %} {% if attrs.height %}height="{{ attrs.height }}"{% endif %} />
```
farhaanbukhsh, line 4: Do you think we can add alt property with title or text here?

farhaanbukhsh (changes requested):
I think the point of having icon and thumbnail is on need basis, like if some courses wanted to show or have the ability to show thumbnail and same goes for icon, but this definitely should accompany on some ways to specify if they are supported.

lti_consumer/templates/html/lti-dl/render_image.html
shimulch: Added title as alt if exists.

shimulch:
> I think the point of having icon and thumbnail is on need basis, like if some courses wanted to show or have the ability to show thumbnail and same goes for icon, but this definitely should accompany on some ways to specify if they are supported.

I've updated my PR and added ``icon`` and ``thumbnail`` support. I think it reasonable to add them as hyperlinked to the actual image. Clicking on ``icon`` and ``thumbnail`` will open the actual image. I also didn't find any reason to add both of them at the same time. So ``thumbnail`` will take precedence over ``icon`` if both exist and only one will show.

farhaanbukhsh (approved):
👍 

- [ ] I tested this: I have added the attribute for image and checked if the image is displayed properly, also I checked if thumbnail and icons are added it becomes a link to the image.
- [x] I read through the code
- ~~[ ] Includes documentation~~
- ~~[ ] I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository.~~

lti_consumer/templates/html/lti-dl/render_image.html
```diff
@@ -1,4 +1,14 @@
...
+<a href="{{ attrs.url }}" {% if attrs.title %}alt="{{ attrs.title }}"{% endif %}>
+    <img src="{{ attrs.thumbnail.url }}" {% if attrs.title %}alt="{{ attrs.title }}"{% endif %} width="{{ attrs.thumbnail.width }}" height="{{ attrs.thumbnail.height }}" />
+</a>
+{% elif attrs.icon %}
+<a href="{{ attrs.url }}" {% if attrs.title %}alt="{{ attrs.title }}"{% endif %}>
+    <img src="{{ attrs.icon.url }}" {% if attrs.title %}alt="{{ attrs.title }}"{% endif %} width="{{ attrs.icon.width }}" height="{{ attrs.icon.height }}" />
+</a>
+{% else %}
+<img src="{{ attrs.url }}" {% if attrs.title %}alt="{{ attrs.title }}" {% endif %}{% if attrs.width %}width="{{ attrs.width }}" {% endif %}{% if attrs.height %}height="{{ attrs.height }}"{% endif %} />
+{% endif %}
```
farhaanbukhsh, line 14: can you add a new line here please

shimulch: Added.

giovannicimolin (approved):
:+1: 

This is an approval for the **serializer** portion of the code (commit: 9b0532c079558e00683c762dd28632c864b98401).

- [x] I tested this: 
1. Checked out this branch.
2. Submitted a image component though the DL response endpoint and made sure it was saved correctly.
3. Tested a few invalid cases and checked that the request failed (no url, invalid url).
- [x] I read through the code
- [ ] I checked for accessibility issues
- [ ] Includes documentation

**Notes:**
I'm not sure if when using `icon` and `thumbnail` all of their sub-properties are required. Something to test once we find a tool that implements it.

nedbat:
Are we sure it's OK to use guessable ids in URLs like `/api/lti_consumer/v1/lti/<LTI_CONFIG_ID>/lti-dl/content` ?

nedbat (approved):

nedbat:
I've merged this manually, thanks!

shimulch:
> Are we sure it's OK to use guessable ids in URLs like /api/lti_consumer/v1/lti/<LTI_CONFIG_ID>/lti-dl/content ?

Even though it's guessable, access being checked properly. One improvement we can do in the future is to use UUID as the primary key for the LTI Config model. That might be a way to make it non-guessable.

giovannicimolin:
@shimulch @nedbat Since we only use a single content type per launch (see https://github.com/edx/xblock-lti-consumer/blob/master/lti_consumer/lti_1p3/deep_linking.py#L56) we can use the block id to locate the LTI Deep Linking content. 

Though I think that using a unique identifier would be better and leave one less place to refactor code if we end up allowing multiple content types in the future.

I'll schedule a follow up ticket to clean up the LTI Advantage implementation and will add this point in the scope.

PR #140, opened 2021-02-04
fix: Migrate the UUID backfill again

stvstnfrd:
To correct the, in our case, single database entry that snuck in without
a UUID during the schema change.

Steps:
- Make the field editable
- Add `config_id` to all models missing it
- Make the field noneditable again

### DIFF
--- a/lti_consumer/migrations/0008_fix_uuid_backfill.py
+++ b/lti_consumer/migrations/0008_fix_uuid_backfill.py
@@ -1,0 +1,31 @@
+from django.db import migrations, models
+import uuid
+
+
+def create_config_ids(apps, schema_editor):
+    LtiConfiguration = apps.get_model('lti_consumer', 'LtiConfiguration')
+    broken = LtiConfiguration.objects.filter(config_id__isnull=True)
+    for config in broken:
+        config.config_id = uuid.uuid4()
+        config.save()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('lti_consumer', '0007_ltidlcontentitem'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='lticonfiguration',
+            name='config_id',
+            field=models.UUIDField(default=uuid.uuid4, editable=True, unique=True),
+        ),
+        migrations.RunPython(create_config_ids),
+        migrations.AlterField(
+            model_name='lticonfiguration',
+            name='config_id',
+            field=models.UUIDField(default=uuid.uuid4, editable=False, unique=True),
+        ),
+    ]
--- a/README.rst
+++ b/README.rst
@@ -298,6 +298,12 @@ Please do not report security issues in public. Send security concerns via email
 Changelog
 =========
 
+2.5.4 - 2021-02-XX
+------------------
+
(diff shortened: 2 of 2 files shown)


stvstnfrd:
TL;DR: We have a single entry in the database that doesn't have a UUID set; this should fix it.

Note: It's hard to pin-point the exact time this entry was created, as the rows have no created/modified timestamp.
That said, it is recent enough, relative to the table size, that it seems likely.

Warning: I still need to properly test this in my devstack; I need to mock up the data in this half-state, but I _think_ this is the correct approach.

stvstnfrd:
@xitij2000 @davidjoy

lti_consumer/migrations/0008_fix_uuid_backfill.py
```diff
@@ -0,0 +1,31 @@
...
+            model_name='lticonfiguration',
+            name='config_id',
+            field=models.UUIDField(default=uuid.uuid4, editable=True, unique=True),
+        ),
+        migrations.RunPython(create_config_ids),
+        migrations.AlterField(
+            model_name='lticonfiguration',
+            name='config_id',
+            field=models.UUIDField(default=uuid.uuid4, editable=False, unique=True),
+        ),
```
xitij2000, line 30: Can't this just be handled with just the runpython script? I don't see why the AlterFields are needed.

stvstnfrd: Are you suggesting that the update should go through even when `editable=True`?

sarina:
Are we open to this error again and again if a new config is added during the migration? Is there any way to mitigate that?

lti_consumer/migrations/0008_fix_uuid_backfill.py
xitij2000: I think `editable=False` only applies to the Django admin. It is still possible to edit the field through code.

xitij2000:
> Are we open to this error again and again if a new config is added during the migration? Is there any way to mitigate that?

That won't  happen again, because now that field has a default value that's a dynamically generated uuid4. Any new config added now will always have a valid uuid4 config_id. It's only during the previous migration that there was a window during which blank values could be added. 

Rather than a migration this change could also be applied via management command.

sarina:
Yes, duh, my brain farted while writing that comment. Is there a reason why this hasn't been merged for 11 days?

davidjoy (approved):

marcotuts:
I can try Piazza I believe as soon as this is merged.

lti_consumer/migrations/0008_fix_uuid_backfill.py
```diff
@@ -0,0 +1,31 @@
...
+
+    dependencies = [
+        ('lti_consumer', '0007_ltidlcontentitem'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='lticonfiguration',
+            name='config_id',
+            field=models.UUIDField(default=uuid.uuid4, editable=True, unique=True),
```
Agrendalath, line 23: @stvstnfrd, shouldn't this have `null=True` instead of `unique=True`?

PR #142, opened 2021-02-17
[BD-24][TNL-7919] LTI Advantage Names and Roles provisioning service implementation direction

giovannicimolin:
This PR adds an ADR to discuss the possible implementations of the LTI NRPS service.

Related PRs:
https://github.com/edx/xblock-lti-consumer/pull/124
https://github.com/edx/edx-platform/pull/25843
https://github.com/edx/edx-platform/pull/25844

@ormsbee Since you raised the scalability issue of the API.
@nedbat @shimulch

### DIFF
--- a/docs/decisions/0004-lti-advantage-nrps.rst
+++ b/docs/decisions/0004-lti-advantage-nrps.rst
@@ -81,4 +81,4 @@ Placeholder.
 Consequences
 ============
 
-Depends on choosen approach.
+Depends on chosen approach.
--- a/docs/decisions/0004-lti-advantage-nrps.rst
+++ b/docs/decisions/0004-lti-advantage-nrps.rst
@@ -57,7 +57,7 @@ we're still left with an API pattern that would be actively encouraging clients 
 to crawl enrollment data.
 
 .. _`Names and Roles Provisioning Services (NRPS)`: http://www.imsglobal.org/spec/lti-nrps/v2p0
-.. _`even with pagination enable as defined in the spec`: http://www.imsglobal.org/spec/lti-nrps/v2p0#limit-query-parameter
+.. _`even with pagination as defined in the spec`: http://www.imsglobal.org/spec/lti-nrps/v2p0#limit-query-parameter
 
 Proposed solutions
 ==================
--- a/docs/decisions/0004-lti-advantage-nrps.rst
+++ b/docs/decisions/0004-lti-advantage-nrps.rst
@@ -56,8 +56,8 @@ students in a given course. Given the scale of some courses in edX.org, even if 
 we're still left with an API pattern that would be actively encouraging clients out there to make hundreds of thousands of successive requests in order
 to crawl enrollment data.
 
-.. _`Names and Roles Provisioning Services (NRPS)`: http://www.imsglobal.org/spec/lti-nrps/v2p0
-.. _`even with pagination as defined in the spec`: http://www.imsglobal.org/spec/lti-nrps/v2p0#limit-query-parameter
+.. _Names and Roles Provisioning Services (NRPS): http://www.imsglobal.org/spec/lti-nrps/v2p0
+.. _even with pagination as defined in the spec: http://www.imsglobal.org/spec/lti-nrps/v2p0#limit-query-parameter
 
 Decision
 ========
--- a/docs/decisions/0004-lti-advantage-nrps.rst
+++ b/docs/decisions/0004-lti-advantage-nrps.rst
@@ -1,0 +1,84 @@
+LTI Advantage NRPS (Names and Roles provisioning services)
+----------------------------------------------------------
+
+Status
+======
+
+In Review
+
+Context
+=======
+
+One of the LTI Advantage services if the `Names and Roles Provisioning Services (NRPS)`_, which allows LTI tools
+to retrieve a user list associated with the learning context and internally provision resources for them.
+
+From the LTI-NRPS Specification:
+
+    The Names and Role Provisioning Services is based on IMS Learning Information Services (LIS) [LIS-20]
+    and W3C Organization Ontology [W3C-ORG]. It is concerned with providing access to data about users’ roles
+    within organizations, a course being an example of an organization. So a very common purpose for this service
+    is to provide a roster (list of enrolments) for a course.
+
+The LTI-NRPS services has two services which provide user data in the following format:
+
+.. code-block:
+    {
+        "id" : "https://lms.example.com/sections/2923/memberships",
+        "context": {
+            "id": "2923-abc",
+            "label": "CPS 435",
+            "title": "CPS 435 Learning Analytics",
+        },
+        "members" : [
+            {
+                "status" : "Active",
+                "name": "Jane Q. Public",
+                "picture" : "https://platform.example.edu/jane.jpg",
+                "given_name" : "Jane",
+                "family_name" : "Doe",
+                "middle_name" : "Marie",
+                "email": "jane@platform.example.edu",
+                "user_id" : "0ae836b9-7fc9-4060-006f-27b2066ac545",
+                "lis_person_sourcedid": "59254-6782-12ab",
+                "roles": [
+                    "http://purl.imsglobal.org/vocab/lis/v2/membership#Instructor"
+                ]
+            }
+        ]
+    }
+
+Where the only required fields for each member are :code:`user_id` and :code:`roles`, to avoid unintended PII transmission to a third party.
+But since one of the purposes of this API is to enable tools access to user PII, the implementation should provide a flag to enable the consumer
+to pass additional PII parameters such as name, email, and a profile picture url.
+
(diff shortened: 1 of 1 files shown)


edx-webhook:
Thanks for the pull request, @giovannicimolin! I've created [BLENDED-769](https://openedx.atlassian.net/browse/BLENDED-769) to keep track of it in Jira.   More details are on the [BD-24](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1545076784/BD-24+LTI+v1.3+Improvements+Advantage+Support) project page.


docs/decisions/0004-lti-advantage-nrps.rst
```diff
@@ -0,0 +1,84 @@
...
+Implement NRPS for courses up to a predefined number of students
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Placeholder.
+
+Implement NRPS limiting the context of the data retrieved
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Placeholder.
+
+Consequences
+============
```
ormsbee, line 82: Another option is to pre-calculate the enrollment data (say on a nightly basis) and make a giant file that we push up to S3 and redirect to with an authorized link. That would allow us to serve the data quickly, though it would mean that the data was stale. Also, there's a good chance that getting a single file with a million enrollments would choke any potential consumer... but it is an option.

docs/decisions/0004-lti-advantage-nrps.rst
```diff
@@ -0,0 +1,84 @@
...
+
+Don't implement NRPS
+~~~~~~~~~~~~~~~~~~~~
+Placeholder.
+
+Implement NRPS gated by course
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Placeholder.
+
+Implement NRPS for courses up to a predefined number of students
```
shimulch, line 73: This seems a good option to me. One suggestion I have is to use the NRPS spec (pagination) based API for courses up to a fixed number of students. And Implement the pre-calculation approach suggested by ``ormsbee`` for other courses.

docs/decisions/0004-lti-advantage-nrps.rst
shimulch: We can actually pre-calculate multiple files in a scheduled task. And use those files as paginations. That will not decrease the number of requests, but requests will not do any db queries.

ormsbee:
So before we dive too deeply on the merits of particular approaches, I'd really like this ADR to articulate the use cases we're solving for in implementing this part of the spec for Open edX courses. There are certain use cases I could imagine (e.g. check to see if this user is really allowed to use some service), for which having one-day-stale data may be unacceptable. If it's mostly used by proctoring and other premium services, maybe limiting it to a subset of enrollments makes sense. Those are completely unvalidated examples I made up on the spot, but my point is that I really don't know how different categories of clients are going to use this information, so it's hard to frame any discussion around technical tradeoffs.

docs/decisions/0004-lti-advantage-nrps.rst
```diff
@@ -0,0 +1,84 @@
...
+From the LTI-NRPS Specification:
+
+    The Names and Role Provisioning Services is based on IMS Learning Information Services (LIS) [LIS-20]
+    and W3C Organization Ontology [W3C-ORG]. It is concerned with providing access to data about users’ roles
+    within organizations, a course being an example of an organization. So a very common purpose for this service
+    is to provide a roster (list of enrolments) for a course.
+
+The LTI-NRPS services has two services which provide user data in the following format:
+
+.. code-block:
```
nedbat, line 24:
```suggestion
.. code-block:: json
```

docs/decisions/0004-lti-advantage-nrps.rst
```diff
@@ -0,0 +1,84 @@
...
+
+Status
+======
+
+In Review
+
+Context
+=======
+
+One of the LTI Advantage services if the `Names and Roles Provisioning Services (NRPS)`_, which allows LTI tools
```
nedbat, line 12:
```suggestion
One of the LTI Advantage services is the `Names and Roles Provisioning Services (NRPS)`_, which allows LTI tools
```

docs/decisions/0004-lti-advantage-nrps.rst
```diff
@@ -0,0 +1,84 @@
...
+Placeholder.
+
+Implement NRPS limiting the context of the data retrieved
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Placeholder.
+
+Consequences
+============
+
+Depends on choosen approach.
```
nedbat, line 84:
```suggestion
Depends on chosen approach.
```

nedbat (changes requested):

docs/decisions/0004-lti-advantage-nrps.rst
```diff
@@ -0,0 +1,116 @@
...
+                ]
+            }
+        ]
+    }
+
+Where the only required fields for each member are :code:`user_id` and :code:`roles`, to avoid unintended PII transmission to a third party.
+But since one of the purposes of this API is to enable tools access to user PII, the implementation should provide a flag to enable the consumer
+to pass additional PII parameters such as name, email, and a profile picture url.
+
+The main issue with enabling this API (`even with pagination as defined in the spec`_) is that it would allow tools to scrape enrollments for all
```
nedbat, line 54:
```suggestion
The main issue with enabling this API (`even with pagination as defined in the spec`_) is that it would allow external third-party tools to scrape enrollments for all
```

docs/decisions/0004-lti-advantage-nrps.rst
```diff
@@ -0,0 +1,116 @@
...
+
+   * - Toggle name
+     - Type
+     - Behavior
+   * - LTI_NRPS_ENABLED
+     - Feature flag
+     - Enables and disables LTI NRPS globally in the Open edX instance. Is disabled by default.
+   * - LTI_NRPS_ACTIVE_ENROLLMENT_LIMIT
+     - Django setting
+     - Controls the allowed number of active enrollments allowed where the API is still available.
```
nedbat, line 83:
```suggestion
     - Controls the allowed number of active enrollments when the API is enabled.
```

docs/decisions/0004-lti-advantage-nrps.rst
```diff
@@ -0,0 +1,116 @@
...
+But since one of the purposes of this API is to enable tools access to user PII, the implementation should provide a flag to enable the consumer
+to pass additional PII parameters such as name, email, and a profile picture url.
+
+The main issue with enabling this API (`even with pagination as defined in the spec`_) is that it would allow tools to scrape enrollments for all
+students in a given course. Given the scale of some courses in edX.org, even if the API is made more performant with cursor pagination,
+we're still left with an API pattern that would be actively encouraging clients out there to make hundreds of thousands of successive requests in order
+to crawl enrollment data.
+
+.. _`Names and Roles Provisioning Services (NRPS)`: http://www.imsglobal.org/spec/lti-nrps/v2p0
+.. _`even with pagination as defined in the spec`: http://www.imsglobal.org/spec/lti-nrps/v2p0#limit-query-parameter
```
nedbat, line 60:
```suggestion
.. _Names and Roles Provisioning Services (NRPS): http://www.imsglobal.org/spec/lti-nrps/v2p0
.. _even with pagination as defined in the spec: http://www.imsglobal.org/spec/lti-nrps/v2p0#limit-query-parameter
```

docs/decisions/0004-lti-advantage-nrps.rst
```diff
@@ -0,0 +1,116 @@
...
+     - CourseWaffleFlag
+     - Allows the tool to access student and instructor PII (username, email, full name, profile picture).
+       Defaults to False.
+
+
+Consequences
+============
+
+* Small courses will be able to use LTI NRPS endpoints on their tools, up to a limited amount of users.
+* Big MOOCs (with millions of users) won't be able to scrape enrollment data on edX and potentially cause stability issues.
```
nedbat, line 95:
```suggestion
* Big MOOCs (with millions of users) won't won't allow scraping of enrollment data on edX and potentially cause stability issues.
```

docs/decisions/0004-lti-advantage-nrps.rst
```diff
@@ -0,0 +1,116 @@
...
+*
```
nedbat, line 96:
```suggestion

```

nedbat (changes requested):
Just some tweaks of English to clarify who is acting.

giovannicimolin:
@nedbat I addressed your comments in the last commit.

nedbat (approved):

PR #143, opened 2021-02-24
[BD-24][TNL-7977][BB-3789]: General improvements to LTI 1.3 implementation

giovannicimolin:
This PR improves logging and adds more information to the studio UI:
- d49965b925753c3fc2dacb86b6d4343b2b7e805d: Improve error and exception logging and pass tracebacks forward. Use the django built in functions for 404 errors instread of `HttpResponse`
- 3492fa2dd5d1ad3bd90d5ca02236bacee08b4a99: Remove misleading LTI information from Studio UI and more information when Deep linking is already set up.
- 65d8096bfbdbc3ab915d1bff083914a2ee171764: Internationalize LTI 1.3 templates
- 065de1c778b85f5ad74ecfb5b4c99f829863c9a5: Add i18n tooling

**Testing instructions:**
1. Checkout this branch.
2. Set up an LTI tool that **supports deep linking** (either IMS tool with user logged in or django pylti tool).
3. Check the studio view.
4. Run a deep linking launch.
5. Refresh the studio view page and check that the LTI Deep Linking information is displayed. 

**Reviewers:**
- [ ] @shimulch 
- [ ] @nedbat 

**Useful links:**
- https://openedx.atlassian.net/wiki/spaces/COMM/pages/753532962/Native+XBlocks+Morning+Session
- https://github.com/edx/i18n-tools


edx-webhook:
Thanks for the pull request, @giovannicimolin! I've created [BLENDED-776](https://openedx.atlassian.net/browse/BLENDED-776) to keep track of it in Jira.   More details are on the [BD-24](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1545076784/BD-24+LTI+v1.3+Improvements+Advantage+Support) project page.


giovannicimolin:
@shimulch @nedbat This is ready for review.

lti_consumer/templates/html/lti-dl/dl_response_error.html
```diff
@@ -0,0 +1,17 @@
+<!DOCTYPE HTML>
+<html>
+    <head>
+        <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
+        <title>LTI Deep Linking failed.</title>
```
shimulch, line 5: Should we use translated text here?

lti_consumer/templates/html/lti-dl/dl_response_saved.html
```diff
@@ -0,0 +1,15 @@
...
+    <head>
+        <meta http-equiv="Content-Type" content="text/html; charset=UTF-8" />
+        <title>LTI Deep Linking</title>
+    </head>
+    <body>
+        <p>
+            <b>The LTI Deep Linking content was successfully saved in the LMS.</b>
+        </p>
+        <p>
+            You can safely close this page now.
```
shimulch, line 12: Also, can we use translations for these and other static texts too?

shimulch (changes requested):

giovannicimolin:
@shimulch Done, internationalized the LTI 1.3 pages and added some translation tooling to the repo.

giovannicimolin:

lti_consumer/translations/locale
```diff
@@ -0,0 +1 @@
+locale
```
OmarIthawi, line 1: This file is a symlink to itself? I don't think this is correct.

OmarIthawi (approved):
Thanks @giovannicimolin for adding i18n tooling support. It looks good to me except one little symlink that I don't think it makes sense.

lti_consumer/translations/locale
giovannicimolin: @OmarIthawi I've commited the wrong symlink. Fixing it soon :)

giovannicimolin: Fixed!

OmarIthawi:
Thanks @giovannicimolin for the fix! Now it looks good! Please consider checking the `es` locale after generating dummy variables. I or Ned can help if this is your first time with `i18n-tools`.

giovannicimolin:
@OmarIthawi I've checked the `fr`and `ar` locales and the current translations seem to be working. My LMS is not picking up `es_419`, not sure why.

OmarIthawi:
> @OmarIthawi I've checked the `fr`and `ar` locales and the current translations seem to be working. My LMS is not picking up `es_419`, not sure why.

I don't think you've broken anything i18n with this PR. I meant `eo` instead of `es`. But if `ar` works, then all other languages are fine in my opinion.

lti_consumer/plugin/views.py
```diff
@@ -111,8 +117,9 @@ def public_keyset_endpoint(request, usage_id=None):
         response = JsonResponse(lti_config.lti_1p3_public_jwk)
         response['Content-Disposition'] = 'attachment; filename=keyset.json'
         return response
-    except (LtiError, InvalidKeyError, ObjectDoesNotExist):
-        return HttpResponse(status=404)
+    except (LtiError, InvalidKeyError, ObjectDoesNotExist) as exc:
+        log.info("Error while retrieving keyset for given usage_id.")
```
nedbat, line 121: Why is this `info` if it is an error?  

More importantly, and throughout all the logging: we should include data in the message.  What usage_id? The message from the exception would be helpful:
```
log.info("Error while retrieving keyset for usage_id %r: %s", usage_id, exc)
```

lti_consumer/plugin/views.py
```diff
@@ -135,8 +142,9 @@ def launch_gate_endpoint(request, suffix):
             handler='lti_1p3_launch_callback',
             suffix=suffix
         )
-    except Exception:  # pylint: disable=broad-except
-        return HttpResponse(status=404)
+    except Exception as exc:
+        log.warning("Error preparing LTI 1.3 launch.")
```
nedbat, line 146:
```
log.warning("Error preparing LTI 1.3 launch with hint %r: %s", usage_key_str, exc)
```
etc

nedbat (changes requested):
The translation and UI stuff looks great.

For the logging, I want more information throughout.

lti_consumer/plugin/views.py
giovannicimolin: @nedbat 
>Why is this info if it is an error?

These are known errors that can happen during normal usage of the LTI block, so that's why I didn't want to fill the logs with errors related to normal usage most of the time. I'll change them to `exception` again since our goal is to make the tool easy to debug.

> More importantly, and throughout all the logging: we should include data in the message. 

Makes sense, I'll improve the logs and add more data.

nedbat: You know better than me how likely the exception is.  If this is something that a site operator should be concerned about, then we should make it a warning.  If it's something that can happen because an instructional team made a typo, then we can leave it as `info`.

giovannicimolin: I've replaced some instances of `info` with `warning` in permission denied situations and others where the error handling is done inside the XBlock.

I've also improved the error messages, and checked that the exceptions inside the LTI 1.3 module pass down the full traceback (`except ... as exc: raise ... from exc`).

giovannicimolin:
@nedbat Addressed your review comments and added a few extra logging points in LTI 1.3 methods.

lti_consumer/plugin/views.py
```diff
@@ -118,7 +118,7 @@ def public_keyset_endpoint(request, usage_id=None):
         response['Content-Disposition'] = 'attachment; filename=keyset.json'
         return response
     except (LtiError, InvalidKeyError, ObjectDoesNotExist) as exc:
-        log.info("Error while retrieving keyset for given usage_id.")
+        log.info("Error while retrieving keyset for usage_id %s: %r", usage_id, exc)
```
nedbat, line 121: The `%s` and `%r` are backwards.  We want the repr() for the usage_id where we might be unsure of the type or trailing white space.  The %s is for the exception, to turn it into its message.

nedbat: (sorry, throughout...)

lti_consumer/plugin/views.py
```diff
@@ -226,11 +226,16 @@ def deep_linking_response_endpoint(request, lti_config_id=None):
...
-        log.info("LtiConfiguration does not exist.")
+        log.info("LtiConfiguration %s does not exist: %r", lti_config_id, exc)
         raise Http404 from exc
     # Bad JWT message, invalid token, or any message validation issues
-    except (Lti1p3Exception, PermissionDenied):
-        log.info("Permission denied for user.")
+    except (Lti1p3Exception, PermissionDenied) as exc:
+        log.warning(
+            "Permission on LTI Config %s denied for user %s: %r",
+            lti_config,
```
nedbat, line 235: It's a little tricky using `lti_config` here, since its assignment is inside the try block, so in theory this clause could be triggered before the variable has a value.  But I think it is OK in this case.

lti_consumer/plugin/views.py
```diff
@@ -241,8 +246,8 @@ def deep_linking_response_endpoint(request, lti_config_id=None):
             },
             status=403
         )
-    except KeyError:
-        log.info("LTI Content Type not supported.")
+    except KeyError as exc:
```
nedbat, line 249: This exception type is too generic for so large a try block.  How can we be sure that KeyErrors mean the content type is not supported?  Can we move this error handling to a narrower chunk of code?  And also include the content type in the message.

nedbat (changes requested):

lti_consumer/plugin/views.py
giovannicimolin: I used a specific exception for this case since there was already one in place: `LtiDeepLinkingContentTypeNotSupported`.

lti_consumer/plugin/views.py
giovannicimolin: @nedbat `Lti1p3Exception` and `PermissionDenied` will only ever happen after the lti configuration is retrieved. The only other case where this variable isn't set is covered by `LtiConfiguration.DoesNotExist`

lti_consumer/plugin/views.py
giovannicimolin: Sorry, my mistake. I've fixed the occurrences on the log statements.

nedbat:
Thanks, looks great!

nedbat (approved):

PR #144, opened 2021-02-24
fix: Manually backfill config_id UUIDs again

stvstnfrd:
The previous backfills handled NULL entries;
this handles empty strings.

The difficulty here was that since the empty string in an invalid UUID
(as opposed to None/NULL which is just "blank"), attempts to instantiate
models would fail, given the invalid data. This meant we couldn't use
the Django ORM to handle the migration entirely; we need to craft some
raw SQL to work-around these checks and limitations.

### DIFF
--- a/lti_consumer/migrations/0009_backfill-empty-string-config-id.py
+++ b/lti_consumer/migrations/0009_backfill-empty-string-config-id.py
@@ -1,0 +1,72 @@
+"""
+Backfill empty config_id records
+
+We need to do this with raw SQL,
+otherwise the model fails upon instantiation,
+as the empty string is an invalid UUID.
+"""
+import uuid
+
+from django.db import connection
+from django.db import migrations
+
+
+sql_forward = """\
+UPDATE
+    lti_consumer_lticonfiguration
+SET
+    config_id = %s
+WHERE
+    id = %s
+;\
+"""
+
+sql_select_empty = """\
+SELECT
+    id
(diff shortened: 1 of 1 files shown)


stvstnfrd:
Testing steps:
- Create a new LtiConfiguration:
  - http://127.0.0.1:18000/admin/lti_consumer/lticonfiguration/
  - Note the ID (probably 1, if it's your first)
- Manually set its `config_id` in the database:
  - `make mysql57-shell`
  - `UPDATE lti_consumer_lticonfiguration SET config_id = '' WHERE id=1;`
- Confirm the DiscussionConfig /add page breaks:
  - http://127.0.0.1:18000/admin/discussions/discussionsconfiguration/add/
- Install the XBlock locally and run the migration
  - `pip install -e src/xblock-lti-consumer`
  - `./manage.py lms migrate lti_consumer`
- Confirm the DiscussionConfig /add page does _NOT_ break:
  - http://127.0.0.1:18000/admin/discussions/discussionsconfiguration/add/

sarina:
Nit, I might name the migration `0009_backfill-empty-str-config-id.py` to be a little more descriptive

stvstnfrd:
> Nit, I might name the migration `0009_backfill-empty-str-config-id.py` to be a little more descriptive

Done: `0009_backfill-empty-string-config-id.py`

davidjoy (approved):

PR #145, opened 2021-02-25
fix: Manually backfill lti_config dicts

stvstnfrd:
The difficulty here is that the empty string in an invalid JSON
dictionary (as opposed to None/NULL or '{}' which is just "blank");
attempts to instantiate models would fail, given the invalid data. This
meant we couldn't use the Django ORM to handle the migration entirely;
we need to craft some raw SQL to work-around these checks and
limitations.

Related-To: 42a9e342ef57a9ffa5016f4f57da633a811d2c0f
Related-To: https://github.com/edx/xblock-lti-consumer/pull/144

### DIFF
--- a/lti_consumer/migrations/0010_backfill-empty-string-lti-config.py
+++ b/lti_consumer/migrations/0010_backfill-empty-string-lti-config.py
@@ -1,0 +1,68 @@
+"""
+Backfill empty lti_config records
+
+We need to do this with raw SQL,
+otherwise the model fails upon instantiation,
+as the empty string is an invalid JSON dictionary.
+"""
+import uuid
+
+from django.db import connection
+from django.db import migrations
+
+
+sql_forward = """\
+UPDATE
+    lti_consumer_lticonfiguration
+SET
+    lti_config = %s
+WHERE
+    id = %s
+;\
+"""
+
+sql_select_empty = """\
+SELECT
+    id
+FROM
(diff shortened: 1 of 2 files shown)


stvstnfrd:
Cribbed from #144; can be compared locally against the previous migration with:
- `vimdiff lti_consumer/migrations/00{09,10}*.py`

stvstnfrd:
CC: @xitij2000 @Agrendalath

stvstnfrd:
Testing steps:
- Create a new LtiConfiguration:
  - http://127.0.0.1:18000/admin/lti_consumer/lticonfiguration/
  - Note the ID (probably 1, if it's your first)
- Manually set its `lti_config` in the database:
  - `make mysql57-shell`
  - `UPDATE lti_consumer_lticonfiguration SET lti_config = '' WHERE id=1;`
- Confirm the DiscussionConfig /add page breaks:
  - http://127.0.0.1:18000/admin/discussions/discussionsconfiguration/add/
- Install the XBlock locally and run the migration
  - `pip install -e src/xblock-lti-consumer`
  - `./manage.py lms migrate lti_consumer`
- Confirm the DiscussionConfig /add page does _NOT_ break:
  - http://127.0.0.1:18000/admin/discussions/discussionsconfiguration/add/

davidjoy (approved):

PR #146, opened 2021-03-01
Replace jsonfield2 with jsonfield

iamsobanjaved:
**Description:** We are going to replace jsonfield2 with jsonfield as the earlier one has archived and merged back in the latter one.

**JIRA:** https://openedx.atlassian.net/browse/BOM-2447


iamsobanjaved:
I generated the migration file through the following steps

- Cloned this in `workspace/src/xblock-lti-consumer` with devstack and platform in `workspace/` directory
- Ran `lms-shell` though devstack
- Installed this package through `pip install -e /edx/src/xblock-lti-consumer/`
- Ran the migration command in lms `python manage.py lms makemigrations`
- Pushed the generated migration file

iamsobanjaved:
@stvstnfrd can you review this PR

UsamaSadiq (approved):

stvstnfrd:
@iamsobanjaved I'm going to pull this down locally to test. Can you and @UsamaSadiq confirm that you did test running this migration locally? Admittedly, I've never swapped models on a django model before, so that's one thing that's a grey area for me.

stvstnfrd (changes requested):
Running this locally appeared to be a no-op for me (that's what I expected/hoped), so that seems fine in my tests; please confirm that you have tested on your end too (with both an empty and non-empty configuration jsonfield).

Please rename this migration to describe the migration, instead of relying on the auto-generated filename, `0011_auto_20210301_1452.py`.

idegtiarov (approved):
LGTM!

iamsobanjaved:
We had to merge all `jsonfield` PRs in edX packages at once because this PR will bring latest version of `jsonfield` in platform as well which needs migration in few other packages as well. I may have to revert this PR if anyone needs to release this package as other PRs are awaiting review from owning teams

stvstnfrd:
> We had to merge all `jsonfield` PRs in edX packages at once because this PR will bring latest version of `jsonfield` in platform as well which needs migration in few other packages as well. I may have to revert this PR if anyone needs to release this package as other PRs are awaiting review from owning teams

@iamsobanjaved It looks like this was never released? This commit/merge wasn't tagged (in git, `2.7.2`), so it wasn't uploaded to PyPI, nor was a new version installed in `edx-platform`'s requirements.

I created and pushed the tag. I tried bumping the version installed in `edx-platform` [1], but it looks like this introduces an issue w/ migrations. This may be similar to the migrations here, but I don't have all of the context.

@iamsobanjaved  Can you take a look at this?

This blocks the refactoring work here [2], plus also, the actual release of this changeset ;)

- [1] https://github.com/edx/edx-platform/pull/27308
- [2] https://github.com/edx/edx-platform/pull/27289

iamsobanjaved:
The PR in platform is failing because all 5 packages needs to updated in once, but I have few PRs which are still in review so I have made the revert PR https://github.com/edx/xblock-lti-consumer/pull/155, this will unblock this package. I will make a new  PR and will merge once all are approved.

PR #148, opened 2021-03-10
[BD-24][TNL-8072]: Improve content presentation and add support for displaying multiple content items

giovannicimolin:
This PR improves the Deep Linking content presentation and adds support for presenting multiple content items in the same XBlock. 

**Testing instructions:**
1. Check out this branch.
2. Set up a local LTI tool that supports Deep Linking (recommended: https://github.com/dmitry-viskov/pylti1.3-django-example) and ngrok tunnel.
3. Create a new block, add the LTI consumer and set up the tool.
4. Go to the LMS and check that the LTI launch is working as expected.
5. Go to Studio and click on the deep linking flow to configure content items.
6. Check that the deep link is configured on Django admin.
7. Go to the LMS and check that the launch works (with the correct content from the deep link). The content should be presented directly in the iframe.
image
8. Go back into django admin and duplicate the Deep Link content for the same block.
image
9. Reload the LMS page and check that the two content items are presented, and that they launch correctly when clicking on the links.
image

**IMS conformance testing instructions:**
1. Set up the LTI Reference implementation tool.
2. Do a deep link launch, and select a `ltiResourceLink` content.
3. Go to the LMS and check if the content link selected was launched correctly.
4. Re-do the deep link launch, but this time select more than one content in the IMS tool.
5. Check if the LMS is able to present the supported content types. 

**Reviewers:**
- [ ] @shimulch 
- [ ] @nedbat


edx-webhook:
Thanks for the pull request, @giovannicimolin! I've created [BLENDED-788](https://openedx.atlassian.net/browse/BLENDED-788) to keep track of it in Jira.   More details are on the [BD-24](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1545076784/BD-24+LTI+v1.3+Improvements+Advantage+Support) project page.


giovannicimolin:
@shimulch This is ready for review.

lti_consumer/templates/html/lti-dl/render_lti_resource_link.html
```diff
@@ -0,0 +1,9 @@
+{% load get_dl_lti_launch_url %}
+
+{% if attrs.title %}<h2>{{ attrs.title }}</h2>{% endif %}
+{% if attrs.text %}<p>{{ attrs.text }}</p>{% endif %}
+
+<a href="{{ item | get_dl_lti_launch_url }}" target="_blank">
+    Click here to launch the LTI content in a new tab
+    <i class="icon fa fa-external-link"></i>
+</a>
```
shimulch, line 9: Please add a new line here.

shimulch (changes requested):
@giovannicimolin, The PR looks good. I've tested it on my local devstack and it's working fine.

`LTI_DEEP_LINKING_ENABLED` seems not documented. Can you add that in README?

lti_consumer/templates/html/lti-dl/render_lti_resource_link.html
giovannicimolin: Done!

giovannicimolin:
@shimulch I've addressed your comments and updated the readme, this is ready for review again.

shimulch (approved):
@giovannicimolin Thanks for cleaning this up. LGTM :+1: 
Just a few nits in the documentation.

- [x] I tested this in my local devstack with the https://github.com/dmitry-viskov/pylti1.3-django-example tool. Deep linking working as expected. Also, multiple content item is supported as per the PR description.
- [x] I read through the code
- [x] Includes documentation

giovannicimolin:

lti_consumer/tests/unit/test_api.py
```diff
@@ -184,49 +231,17 @@ def test_retrieve_with_block(self):
                     lti_config.lti_1p3_client_id
                 ),
                 'deep_linking_launch_url': 'https://example.com',
-                'deep_linking_content_items': []
```
nedbat, line 187: This assert uses assertCountEqual, but that won't check everything we need it to. When comparing dicts, it only looks at the keys. This assert would pass: self.assertCountEqual({"a": 1}, {"a": 2}).

Don't we want assertEqual for these tests?

giovannicimolin: @nedbat You're right. This test also had a few incorrect values and I fixed those as well.

nedbat: Thanks!

nedbat (approved):

PR #149, opened 2021-03-18
fix: show the configuration as nicely formatted JSON

nedbat:
@giovannicimolin is it OK to show this as formatted JSON?

### DIFF
--- a/lti_consumer/api.py
+++ b/lti_consumer/api.py
@@ -4,6 +4,9 @@ Python APIs used to handle LTI configuration and launches.
 Some methods are meant to be used inside the XBlock, so they
 return plaintext to allow easy testing/mocking.
 """
+
+import json
+
 from .exceptions import LtiError
 from .models import LtiConfiguration, LtiDlContentItem
 from .utils import (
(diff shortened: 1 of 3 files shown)


giovannicimolin:
@nedbat Each deep linking content type has different and optional parameters, so adding separate rendering templates for Studio would overly complicate that implementation (especially given we're a bit budget constrained on the project).

We can simplify the Studio and display less information (number of content items + their type) - that will be more user-friendly. What do you think? I can implement that change next sprint, along with the documentation in the repo.

nedbat:
> adding separate rendering templates for Studio would overly complicate the Studio view implementation

I'm not sure what this refers to?  I'm not adding separate templates, just JSON-formatting the data on the page.

giovannicimolin:
@nedbat I think I've misunderstood your initial question in the PR. I've read it as `is it OK to show a JSON to the course creators?` as a question for the choice of directly presenting the raw data in Studio.

Presenting it as a nicely formatted JSON is better. :+1:

giovannicimolin (approved):
:+1:

- [x] I tested this: 
1. Checked this out on the devstack.
2. Tested that the JSON output is nicely formatted.
image

- [x] I read through the code
- [ ] ~~Includes documentation~~ NA

PR #150, opened 2021-03-24
[SE-4235] Add missing parameter processors code

pkulkark:
Somewhere between versions 2.0.0 and 2.1.0, the code to handle external parameter processors was dropped. This PR adds it back. Without this, external processors that are specified in the [parameter processors settings](https://github.com/edx/xblock-lti-consumer#configuring-the-parameter-processors-settings) is not called.

**Jira Tickets:**  [OSPR-5686](https://openedx.atlassian.net/browse/OSPR-5686)

**Sandbox URL:** TBD

**Testing Instructions:**

1. Install this branch on your xblock runtime environment (for example xblock-sdk).
2. Test the parameter processors from tahoe-lti as per the [test instructions provided](https://github.com/appsembler/tahoe-lti#development-guide)
3. Verify that the external parameter processors are being called correctly.

**Reviewers**
- [ ] @gabor-boros 
- [ ] edX reviewer TBD

### DIFF
--- a/lti_consumer/lti_1p1/consumer.py
+++ b/lti_consumer/lti_1p1/consumer.py
@@ -146,6 +146,9 @@ class LtiConsumer1p1:
         self.lti_launch_presentation_locale = None
         self.lti_custom_parameters = None
 
+        # Extra claims - used for custom parameter processors
+        self.extra_claims = {}
+
     def set_user_data(
             self,
             user_id,
@@ -241,6 +244,14 @@ class LtiConsumer1p1:
 
         self.lti_custom_parameters = custom_parameters
 
+    def set_extra_claims(self, claim):
+        """
+        Updates launch extra claims using python's dict .update method
+        """
+        if not isinstance(claim, dict):
+            raise ValueError('Invalid extra claim: {!r} is not a dict.'.format(claim))
+        self.extra_claims.update(claim)
+
     def generate_launch_request(self, resource_link_id):
         """
         Signs LTI launch request and returns signature and OAuth parameters.
@@ -287,6 +298,10 @@ class LtiConsumer1p1:
         # Appending custom parameter for signing.
         if self.lti_custom_parameters:
             lti_parameters.update(self.lti_custom_parameters)
+
+        # Extra claims - from custom parameter processors
+        if self.extra_claims:
+            lti_parameters.update(self.extra_claims)
 
         headers = {
             # This is needed for body encoding:
--- a/lti_consumer/lti_xblock.py
+++ b/lti_consumer/lti_xblock.py
@@ -1009,6 +1009,16 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
 
         lti_consumer.set_custom_parameters(self.prefixed_custom_parameters)
 
+        for processor in self.get_parameter_processors():
+            try:
+                default_params = getattr(processor, 'lti_xblock_default_params', {})
+                lti_consumer.set_extra_claims(default_params)
+                lti_consumer.set_extra_claims(processor(self) or {})
+            except Exception:  # pylint: disable=broad-except
+                # Log the error without causing a 500-error.
+                # Useful for catching casual runtime errors in the processors.
+                log.exception('Error in XBlock LTI parameter processor "%s"', processor)
+
         lti_parameters = lti_consumer.generate_launch_request(self.resource_link_id)
         loader = ResourceLoader(__name__)
         context = self._get_context_for_template()
--- a/setup.py
+++ b/setup.py
@@ -49,7 +49,7 @@ with open('README.rst') as _f:
 
 setup(
     name='lti-consumer-xblock',
-    version='3.0.2',
+    version='3.0.3',
     author='Open edX project',
     author_email='oscm@edx.org',
     description='This XBlock implements the consumer side of the LTI specification.',


edx-webhook:
Thanks for the pull request, @pkulkark! I've created [OSPR-5686](https://openedx.atlassian.net/browse/OSPR-5686) to keep track of it in JIRA, where we prioritize reviews. Please note that it may take us up to several weeks or months to complete a review and merge your PR.

- supporting documentation

Please let us know once your PR is ready for our review and all tests are green.

natabene:
@pkulkark Thank you for your contribution. Please let me know once it is ready for our review.

gabor-boros (approved):

pkulkark:
@natabene This is ready for edX's review.

lti_consumer/lti_1p1/consumer.py
```diff
@@ -138,6 +138,7 @@ def __init__(self, lti_launch_url, oauth_key, oauth_secret):
         self.lti_launch_url = lti_launch_url
         self.oauth_key = oauth_key
         self.oauth_secret = oauth_secret
+        self.xblock = xblock
```
giovannicimolin, line 141: @pkulkark @gabor-boros I've been asked by @nedbat to take a look at this PR. 

The approach here doesn't fit the new code structure we're trying to achieve here. 
The `lti_1p1` folder should be a generic LTI 1.1 library, with no knowledge or usage of `xblock` module.

lti_consumer/lti_1p1/consumer.py
```diff
@@ -288,6 +289,17 @@ def generate_launch_request(self, resource_link_id):
...
+            for processor in self.xblock.get_parameter_processors():
+                try:
+                    default_params = getattr(processor, 'lti_xblock_default_params', {})
+                    lti_parameters.update(default_params)
+                    lti_parameters.update(processor(self.xblock) or {})
+                except Exception:  # pylint: disable=broad-except
+                    # Log the error without causing a 500-error.
+                    # Useful for catching casual runtime errors in the processors.
+                    log.exception('Error in XBlock LTI parameter processor "%s"', processor)
+
```
giovannicimolin, line 302: For this specific use case, I'd suggest an approach similar to what I've done for LTI 1.3:
1. Add an extra settings variable to the class: https://github.com/edx/xblock-lti-consumer/blob/master/lti_consumer/lti_1p3/consumer.py#L60 and a public update method for it https://github.com/edx/xblock-lti-consumer/blob/master/lti_consumer/lti_1p3/consumer.py#L448-L454
2. If there's anything on it, update the parameters on launch (same as on https://github.com/edx/xblock-lti-consumer/blob/master/lti_consumer/lti_1p3/consumer.py#L301-L303):
```suggestion
        # Extra claims - from custom parameter processors
        if self.extra_claims:
            lti_message.update(self.extra_claims)
```

Then, on the `xblock` code, when doing an LTI 1.1 launch, add the parameter processors (see https://github.com/edx/xblock-lti-consumer/blob/master/lti_consumer/lti_xblock.py#L1027 for reference):
```
# From XBlock code
for processor in self.get_parameter_processors():
  try:
      default_params = getattr(processor, 'lti_xblock_default_params', {})
      lti_consumer.set_extra_claims(processor(self) or default_params)
  except Exception:  # pylint: disable=broad-except
      # Log the error without causing a 500-error.
      # Useful for catching casual runtime errors in the processors.
      log.exception('Error in XBlock LTI parameter processor "%s"', processor)
```

giovannicimolin (changes requested):
@pkulkark @gabor-boros I left some comments in the code :)

pkulkark:
Thanks @giovannicimolin. I've made the changes as per your suggestions. Could you take a look and let me know if it looks good?

pkulkark:
@giovannicimolin Yup, I tested it on [this sandbox](https://se-4345.stage.opencraft.hosting/). Everything works as expected.

giovannicimolin (approved):
:+1:

- [x] I tested this:
1. I checked the sandbox instance using the `staff` user.
2. Checked data from additional parameter processors were passed to the LTI tool.
- [x] I read through the code
- [ ] ~~Includes documentation~~ NA (bugfix)

@pkulkark Thanks for the fix, this was missed when implementing LTI 1.1 support outside XBlocks for the Discussions epic.

nedbat:
@pkulkark Is this ready for a final review?

pkulkark:
@nedbat Yes, this is good for a final review.

lti_consumer/lti_xblock.py
```diff
@@ -1026,6 +1026,15 @@ def lti_launch_handler(self, request, suffix=''):  # pylint: disable=unused-argu
 
         lti_consumer.set_custom_parameters(self.prefixed_custom_parameters)
 
+        for processor in self.get_parameter_processors():
+            try:
+                default_params = getattr(processor, 'lti_xblock_default_params', {})
+                lti_consumer.set_extra_claims(processor(self) or default_params)
```
nedbat, line 1032: The original lost code did this:
```
                default_params = getattr(processor, 'lti_xblock_default_params', {})
                lti_parameters.update(default_params)
                lti_parameters.update(processor(self.xblock) or {})
```
I don't know if default_params can be a multi-key dict (it sounds like it can).  If default params has a key that the processor() result does not, then the old code would include that key, but the new code would not include it.  What is the correct behavior?

lti_consumer/lti_1p1/consumer.py
```diff
@@ -241,6 +244,14 @@ def set_custom_parameters(self, custom_parameters):
 
         self.lti_custom_parameters = custom_parameters
 
+    def set_extra_claims(self, claim):
+        """
+        Sets launch extra claims
+        """
+        if not isinstance(claim, dict):
+            raise ValueError('Invalid extra claim: is not a dict.')
```
nedbat, line 252: It's helpful to provide data about the bad value.  Perhaps:
```
raise ValueError('Invalid extra claim: {!r} is not a dict.'.format(claim))
```

nedbat (changes requested):

lti_consumer/lti_xblock.py
pkulkark: I'm not too sure, but looking at the [readme doc](https://github.com/edx/xblock-lti-consumer/blob/master/README.rst#defining-an-lti-parameter-processor) I think the default params are to be used if the processor hits an exception (which should happen with the current logic as well). In any case, please let me know if it's better to explicitly include the default params anyway.

lti_consumer/lti_1p1/consumer.py
pkulkark: Sounds good. Added :+1:

gabor-boros:
@nedbat, I think this is ready for another round of reviews 😊

nedbat:
@gabor-boros @giovannicimolin @pkulkark I still have [this question](https://github.com/edx/xblock-lti-consumer/pull/150/files#r657315012).

giovannicimolin:
@nedbat I scheduled some time to re-review and test your comment in the next sprint (starting on Monday). 
Sorry for the delay in the reply!

nedbat:
@giovannicimolin When you get the answer to my question, you should feel free to review and merge if you are comfortable with the changes (as core contributor).

sarina:
@giovannicimolin as Ned is out of town it'd be great if you could review this week!

lti_consumer/lti_xblock.py
giovannicimolin, line 1032: @pkulkark @nedbat is right on this one. 

The previous behavior was that all the default parameters were included and then overwritten by the ones provided by the parameter processor. To keep it working the same way, the code should be like this:
```suggestion
                default_params = getattr(processor, 'lti_xblock_default_params', {})
                lti_consumer.set_extra_claims(default_params)
                lti_consumer.set_extra_claims(processor(self) or {})
```

Maybe `set_extra_claims` is misleading here: it's not a `setter` since it **updates** a dictionary using python's `.update` method instead of overwriting content. Can you add a note in the method's description to clarify the behavior?

giovannicimolin (changes requested):
@pkulkark One comment left to address.

Can you also rebase, and version bump the block to `3.0.3`? I can merge and tag a release once thay is done.

lti_consumer/lti_xblock.py
pkulkark: @giovannicimolin Updated the code, rebased to latest master and updated the version :+1:

giovannicimolin (approved):
:+1:

- [x] I tested this: 
1. Tested that the parameter processors are now working.
2. Checked that the default attributes are now included.
- [x] I read through the code
- [ ] ~~Includes documentation~~ NA: This is a bugfix for a feature previously working

lti_consumer/lti_xblock.py
giovannicimolin: This is now fixed! :+1:

giovannicimolin:
@pkulkark I just realized that squash & merge is not enabled for this repo. Can you squash your commits down to a single one following OEP-51?

I can merge after that.

giovannicimolin:
@sarina @nedbat Merging this as all comments have been addressed.

louis-s-29:
@pkulkark @giovannicimolin - Hi, currently testing out the new xblock-lti-consumer 3.0.3 and tahoe-lti.

Per the "how to test" instructions of tahoe-lti, I was wondering if this PR + Tahoe-lti will work only for Lti1.1 or also Lti1.3?

pkulkark:
@louis-s-29 Tahoe-lti only supports LTI 1.1 as far as I know.

giovannicimolin:
@louis-s-29 Parameter processors are currently not implemented in LTI 1.3, but they can be plugged [in the LTI 1.3 launch handler](https://github.com/edx/xblock-lti-consumer/blob/master/lti_consumer/lti_xblock.py#L1030) along with the custom parameter support.

Upon a quick check, the parameter processor code seems to be generic enough so that it isn't too complex to add support for on LTI 1.3. If you are going to work on adding support, let me know, I can help/guide you through the implementation process.

PR #153, opened 2021-04-06
docs: ADR to decide the future of the CourseEditLTIFieldsEnabledFlag [BD-03] [TNL-7945]

xitij2000:
This ADR proposes that we use the existing CourseEditLTIFieldsEnabledFlag
for LTI Course tabs in addition LTI XBlocks, and move it to this repo as
intended via EDUCATOR-121.

### DIFF
--- a/docs/decisions/0005-lti-pii-sharing-flag.rst
+++ b/docs/decisions/0005-lti-pii-sharing-flag.rst
@@ -51,7 +51,7 @@ The arguments for using the same flag are:
 3. It's confusing to have two flags with very similar purposes.
 4. There is already a double layer of flags here. Enabling this flag will enable
    configuring other option which in turn enable sending PII data. Adding
-   another separate flag doesn't help the situaiton. Even if the flag is only
+   another separate flag doesn't help the situation. Even if the flag is only
    needed for either tabs or XBlocks, there is another step to enabling sharing
    PII which can simply be skipped if undesired.
 
@@ -74,6 +74,9 @@ of just controlling if those fields are editable.
 
 Additionally, any PII sharing via LTI1.3 can also be folded under the same flag.
 
+We should also add an indication to users that this flag is the reason for their
+inability to edit these fields in UI instead of just hiding the fields. Users
+can be given information about getting this flag enabled for their course.
 
 Consequences
 ============
--- a/docs/decisions/0005-lti-pii-sharing-flag.rst
+++ b/docs/decisions/0005-lti-pii-sharing-flag.rst
@@ -1,0 +1,104 @@
+Unified Flag for Enabling Sharing of PII in LTI
+-----------------------------------------------
+
+Status
+======
+
+Proposal
+
+Context
+=======
+
+LTI integrations can be enhanced by sharing some level of personal information
+about users, such as their username and email. Sharing such information allows
+for a smoother registration flow among other benefits.
+
+The `LTI XBlock has options
+<https://github.com/edx/xblock-lti-consumer/blob/edec2a68282a2a1fc2b2036e1989d60688fa6b19/lti_consumer/lti_xblock.py#L471-L487>`_
+called ``ask_to_send_username`` and ``ask_to_send_email`` that enable sending
+such PII to the external tool. However, `by default these options are not
+visible
+<https://github.com/edx/xblock-lti-consumer/blob/edec2a68282a2a1fc2b2036e1989d60688fa6b19/lti_consumer/lti_xblock.py#L588-L599>`_
+to course authors, and there is no way to send PII to LTI tools. In order to
+even enable these options to appear in the LTI XBlock configuration, you need
+to first set a `configuration flag
+<https://github.com/edx/edx-platform/blob/e19ba34f5a564285b3a20a7298c20ca640ca5aa0/cms/djangoapps/xblock_config/models.py#L38-L86>`_
+via Django admin in studio.
+
+This configuration flag unlocks the editing of the above fields after which
+those fields can be set to allow sharing of PII. The configuration flag works
+on a course-wide level, while the flags above work on a per-LTI-XBlock-basis.
+
+With the recent addition of LTI Course tabs, there is need for a similar
+mechanism for enabling sending of PII. What needs to be determined is whether
+we should use the same flag for the purpose, or create a new flag for this use
+with LTI tabs.
+
+The arguments for creating a new flag are:
+
+1. The scope of the current flag doesn't match the new use case.
+2. Currently the flag is implemented in studio, while the flag now needs to be
+   used in LMS and MFEs.
+3. Keeping flags separate allows controlling this flag separately for course
+   tabs vs XBlocks.
+
+The arguments for using the same flag are:
+
+1. As part of `EDUCATOR-121 <https://openedx.atlassian.net/browse/EDUCATOR-121>`_
+   this config flag is already supposed to be moved to the LTI XBlock, which
+   would make it usable from LMS and Studio.
+2. The name and scope can both be changed while retaining the data.
+3. It's confusing to have two flags with very similar purposes.
+4. There is already a double layer of flags here. Enabling this flag will enable
+   configuring other option which in turn enable sending PII data. Adding
+   another separate flag doesn't help the situaiton. Even if the flag is only
+   needed for either tabs or XBlocks, there is another step to enabling sharing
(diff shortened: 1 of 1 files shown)


edx-webhook:
Thanks for the pull request, @xitij2000! I've created [BLENDED-813](https://openedx.atlassian.net/browse/BLENDED-813) to keep track of it in Jira.   More details are on the [BD-03](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1544913061/BD-03+Discussions+Improvements+Integrations+Phase+1) project page.


giovannicimolin (changes requested):
@xitij2000 Since this is related to all LTI integrations does it make sense to also consider LTI 1.3 integrations and advantage extensions?

References with PII on LTI 1.3/Advantage: 
- https://github.com/edx/xblock-lti-consumer/blob/master/docs/decisions/0004-lti-advantage-nrps.rst: We won't implement NRPS now, but there's a flag to gate PII planned.
- http://www.imsglobal.org/spec/lti/v1p3#user-identity-claims: Currently we're only setting the `required` attributes of the spec, but if this goes into effect, we might need to implement the PII fields here as well to keep the implementation consistent (at least `name` and `email`). There's already support in the 1.3 library (https://github.com/edx/xblock-lti-consumer/blob/master/lti_consumer/lti_1p3/consumer.py#L111) but the XBlock does not use it (https://github.com/edx/xblock-lti-consumer/blob/master/lti_consumer/lti_xblock.py#L1040)

docs/decisions/0005-lti-pii-sharing-flag.rst
```diff
@@ -0,0 +1,101 @@
...
+flag will now be accessible via LMS and Studio, which is what is needed.
+
+The LTI course tabs code will be modified to check this value before allowing
+the sending of PII even if the tab's configuration allows sending PII.
+
+Currently, if this flag is enabled for a course, and the `ask_to_send_*` flags
+are enabled for an XBlock, disabling this flag will not stop the XBlock from
+sending PII. This flag will *only* disable editing the `ask_to_send_*` flags.
+
+This might not be desirable behaviour. So we can also modify the scope of this
```
Agrendalath, line 71:
```suggestion
sending PII. This flag will *only* disable editing the `ask_to_send_*` flags. This might not be desirable behavior. So we can also modify the scope of this
```

Agrendalath: "This might not be desirable behaviour." - I believe it's a follow-up to the previous paragraph, right? If yes, then it might make more sense to merge them.

docs/decisions/0005-lti-pii-sharing-flag.rst
```diff
@@ -0,0 +1,101 @@
...
+<https://github.com/edx/xblock-lti-consumer/blob/edec2a68282a2a1fc2b2036e1989d60688fa6b19/lti_consumer/lti_xblock.py#L588-L599>`_
+to course authors, and there is no way to send PII to LTI tools. In order to
+even enable these options to appear in the LTI XBlock configuration, you need
+to first set a `configuraiton flag
+<https://github.com/edx/edx-platform/blob/e19ba34f5a564285b3a20a7298c20ca640ca5aa0/cms/djangoapps/xblock_config/models.py#L38-L86>`_
+via Django admin in studio.
+
+This configuration flag unlocks the editing of the above fields after which
+those fields can be set to allow sharing of PII. The configuraiong flag works
+on a course-wide level, while the flags above work on a per-LTI-XBlock-basis.
```
Agrendalath, line 30: Nit: would it make sense to use a different word than `flag` for `ask_to_send_username` and `ask_to_send_email`? This XBlock does not refer to any of [its fields](https://github.com/edx/xblock-lti-consumer/blob/edec2a68282a2a1fc2b2036e1989d60688fa6b19/lti_consumer/lti_xblock.py) this way but uses this word only in the context of the configuration flags (specified in settings). I believe it would be a bit easier for the reader to distinguish the configuration flag from an XBlock field this way.

docs/decisions/0005-lti-pii-sharing-flag.rst
```diff
@@ -0,0 +1,101 @@
...
+
+The LTI course tabs code will be modified to check this value before allowing
+the sending of PII even if the tab's configuration allows sending PII.
+
+Currently, if this flag is enabled for a course, and the `ask_to_send_*` flags
+are enabled for an XBlock, disabling this flag will not stop the XBlock from
+sending PII. This flag will *only* disable editing the `ask_to_send_*` flags.
+
+This might not be desirable behaviour. So we can also modify the scope of this
+flag to be the final arbiter of whether PII sharing is allowed or not instead
```
Agrendalath, line 72: In addition to this, it would be user-friendly to show the course authors that there is such a configuration flag, which is currently disabled for a course, and that's why these fields are not being displayed. What do you think?

Agrendalath (approved):
:+1:

* [x]  I tested this: read the docs
* [x]  I read through the code
* [x]  I checked for accessibility issues: n/a
* [x]  Includes documentation
* [x] I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository: n/a

docs/decisions/0005-lti-pii-sharing-flag.rst
xitij2000: Will do.

docs/decisions/0005-lti-pii-sharing-flag.rst
xitij2000: I've currently based it on the way the XBloock work, which is to just hide the fields. Since they are hidden by default, only the people who need to be aware of them will ever see them.

docs/decisions/0005-lti-pii-sharing-flag.rst
```diff
@@ -0,0 +1,101 @@
...
+even enable these options to appear in the LTI XBlock configuration, you need
+to first set a `configuraiton flag
+<https://github.com/edx/edx-platform/blob/e19ba34f5a564285b3a20a7298c20ca640ca5aa0/cms/djangoapps/xblock_config/models.py#L38-L86>`_
+via Django admin in studio.
+
+This configuration flag unlocks the editing of the above fields after which
+those fields can be set to allow sharing of PII. The configuraiong flag works
+on a course-wide level, while the flags above work on a per-LTI-XBlock-basis.
+
+With the recent addition of LTI Course tabs, there is need for a similar
```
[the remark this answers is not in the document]
xitij2000: I think this is OK as is :-)

docs/decisions/0005-lti-pii-sharing-flag.rst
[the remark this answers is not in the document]
xitij2000: This change will included in a folloring comment.

docs/decisions/0005-lti-pii-sharing-flag.rst
```diff
@@ -0,0 +1,104 @@
...
+The arguments for using the same flag are:
+
+1. As part of `EDUCATOR-121 <https://openedx.atlassian.net/browse/EDUCATOR-121>`_
+   this config flag is already supposed to be moved to the LTI XBlock, which
+   would make it usable from LMS and Studio.
+2. The name and scope can both be changed while retaining the data.
+3. It's confusing to have two flags with very similar purposes.
+4. There is already a double layer of flags here. Enabling this flag will enable
+   configuring other option which in turn enable sending PII data. Adding
+   another separate flag doesn't help the situaiton. Even if the flag is only
```
davidjoy, line 54: Nit: situaiton -> situation

docs/decisions/0005-lti-pii-sharing-flag.rst
davidjoy: I agree that we should improve the user-friendliness here.  It makes sense to me that the flag would be the final arbiter/override the ask_to_send_* settings, and it may also be worth giving users a note that says how they can get it enabled if necessary.

davidjoy (approved):
This looks great.  Thanks Kshitij!

docs/decisions/0005-lti-pii-sharing-flag.rst
xitij2000: Sure I've updated the doc to include this. Would this apply to the XBlock as well? Or just new UI for the MFEs?

xitij2000:
@davidjoy Could you merge this since it has been approved?

PR #154, opened 2021-04-08
fix: Try importing LMS helper from new path

stvstnfrd:
Fall-back to the old path.

Once the LMS has fully switched over to this new path [1],
we can remove the legacy (LMS) import support here.

- [1] https://github.com/edx/edx-platform/pull/27289

### DIFF
--- a/lti_consumer/plugin/compat.py
+++ b/lti_consumer/plugin/compat.py
@@ -108,9 +108,17 @@ def user_has_access(*args, **kwargs):
 def get_course_by_id(course_key):
     """
     Import and run `get_course_by_id` from LMS
+
+    TODO: Once the LMS has fully switched over to this new path [1],
+    we can remove the legacy (LMS) import support here.
+
+    - [1] https://github.com/edx/edx-platform/pull/27289
     """
     # pylint: disable=import-error,import-outside-toplevel
-    from lms.djangoapps.courseware.courses import get_course_by_id as lms_get_course_by_id
+    try:
+        from openedx.core.lib.courses import get_course_by_id as lms_get_course_by_id
+    except ImportError:
+        from lms.djangoapps.courseware.courses import get_course_by_id as lms_get_course_by_id
     return lms_get_course_by_id(course_key)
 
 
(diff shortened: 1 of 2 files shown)


lti_consumer/plugin/compat.py
```diff
@@ -110,7 +110,10 @@ def get_course_by_id(course_key):
     Import and run `get_course_by_id` from LMS
     """
     # pylint: disable=import-error,import-outside-toplevel
-    from lms.djangoapps.courseware.courses import get_course_by_id as lms_get_course_by_id
+    try:
+        from openedx.core.lib.courses import get_course_by_id as lms_get_course_by_id
+    except ImportError:
+        from lms.djangoapps.courseware.courses import get_course_by_id as lms_get_course_by_id
```
sarina, line 116: `Once the LMS has fully switched over to this new path, we can remove the legacy (LMS) import support here.` -> might add that as a comment or even a TODO in the code

stvstnfrd, line 116: Technically, this would still work, even after the move in `edx-platform` because that LMS module also imports the function anyway (from `openedx.core`), since it's used by another function in that LMS module.

stvstnfrd: @sarina Done, TODO comment added to docstring.

sarina (approved):
Not sure if you're going to add a test but the change itself looks good to me - also gj on commit messages

stvstnfrd:
> Not sure if you're going to add a test

@sarina Nah.. Since the XBlock is tested in isolation and is itself a dependency of `edx-platform`, these tests don't run in the same virtualenv as `edx-platform`, so we can't actually test this here. Well, we could mock the import but that seems silly because then we're really just testing that we've mocked things correctly; it has no bearing on the actual, underlying import path :)

sarina:
I figured it was going to be something like that, but I was asking more of a "I don't know if there's going to be more code to review, but here's a thumb anyways"

stvstnfrd:
I also created tag `2.7.3`; this should now auto-upload to PyPI

PR #155, opened 2021-04-13
Revert "chore: Replace jsonfield2 with jsonfield"

iamsobanjaved:
This reverts commit 2fd330f8a62bc2fcc4951d546db5ca4920a5c1ec.

Reverting only `jsonfield` related changes from that PR


awais786 (approved):

nedbat:
When reverting, please explain why the revert is needed.

nedbat:
Also, this revert deletes a migration.  That migration might have already been applied to a live database. Isn't it true that we should never delete a migration?

stvstnfrd:
Revert commits should _only_ revert code; they should never introduce new code;
this revert also increases the version number [1][2][3].

If we need to make additional changes, these should happen in a separate commit;
squashing the functional change into the revert muddies the commit history and can make future maintenance more difficult.
For example, if we later want to revert this commit (the revert commit, because we think it should work then), then reverting the commit actually downgrades the version from `2.7.4` to `2.7.3`; this is unexpected and should be avoided.

In situations like this, I would probably recommend just reverting the entire commit first.
Then after it has merged, we can decide to reintroduce the subset of cosmetic changes in a new, separate PR.
Especially since part of the issue here is that the version upgrade (to `2.7.2`) never happened (neither tagged nor deployed), it follows that we would revert/remove the entire thing until we know what's going on. It's riskier to manually author new commits directly alongside the removal of known problematic code.


- [1] https://github.com/edx/xblock-lti-consumer/pull/155/commits/3b507a3499c8e07f794346a7f1c79eed496bc24b#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R52
- [2] Worst case, if the commit being reverted also increased the version number, we would want to restore the version number to what it was pre-revert. We don't want to actually downgrade the version number; we want to remove a commit from the current/future version.
- [3] But really, this is why we should just increase the version number in a separate commit (or should have); that way reverting functionality doesn't mess with build numbers.

stvstnfrd:
> Also, this revert deletes a migration. That migration might have already been applied to a live database. Isn't it true that we should never delete a migration?

@nedbat Just to say without commenting on software development best practices, I do not believe edx has run this in production [1], as that (failure trying to install the new version) is what started this chain.

And given that _trying_ to install this causes the test suite to fail, it's _highly_ _unlikely_ that anyone in the community is running this yet; they'd be out in front of us.

I have not confirmed anything though (particularly state of the migrations table), I leave that to @iamsobanjaved .

"Don't remove migrations" is probably a good general principle, but I _suspect_ we can overlook it here, given that the failure here is explicitly an integration failure.

- [1] https://github.com/edx/xblock-lti-consumer/pull/146#issuecomment-818242859

nedbat:
@stvstnfrd thanks for the clarification.  This is the sort of explanation that would be very helpful in the initial pull request description.

PR #156, opened 2021-04-13
[BD-24] Version bump and update docs.

giovannicimolin:
This PR updates the documentation related to the LTI-AGS service and adds a missing version bump.

**Reviewer:**
- [ ] @nedbat

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -329,6 +329,7 @@ Changelog
 
 * LTI Advantage - AGS Service: Added support for programmatic grade management by LTI tools.
 * Improved grade publishing to LMS when using LTI-AGS.
+* Increase LTI 1.3 token validity to 1h.
 
 2.7.0 - 2021-02-16
 ------------------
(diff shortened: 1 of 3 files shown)


edx-webhook:
Thanks for the pull request, @giovannicimolin! I've created [BLENDED-824](https://openedx.atlassian.net/browse/BLENDED-824) to keep track of it in Jira.   More details are on the [BD-24](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1545076784/BD-24+LTI+v1.3+Improvements+Advantage+Support) project page.


README.rst
```diff
@@ -318,6 +324,12 @@ Please do not report security issues in public. Send security concerns via email
 Changelog
 =========
 
+2.7.5 - 2021-02-16
```
nedbat, line 327: This date should be updated, right?  And the number should be 2.8.0, since this is a change in behavior.

giovannicimolin: >This date should be updated, right?

Yes, I've missed that while copying the changelog entry.

>And the number should be 2.8.0, since this is a change in behavior.

Agreed :+1: 
I initially went for the minor version bump since the previous one already deployed this change, but you're right, a separate minor version is better.

nedbat (approved):

PR #168, opened 2021-06-11
build: save common_constraints.txt locally

stvstnfrd:
so that dependabot can access the constraints file and automatically
create security PRs; it fails if the constraints file is specified via
remote URL.


stvstnfrd:
I'm going to merge this now, to confirm it unblocks dependabot..

PR #188, opened 2021-08-10
chore: upgrade requirements and fix pylint violations

stvstnfrd:
This replaces #186 and fixes the failing tests.


lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
```diff
@@ -19,12 +19,11 @@ class UsageKeyField(serializers.Field):
     from the platform like so:
     `from openedx.core.lib.api.serializers import UsageKeyField`
     """
-    # pylint: disable=arguments-differ
-    def to_representation(self, data):
+    def to_representation(self, value):
```
sarina, line 22: what was the reasoning behind naming the arg `data` ? I assume that changing it to `value` doesn't affect anything

sarina (approved):

lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
stvstnfrd: This changed "upstream" in the parent class; we get a warning without the change (though it technically shouldn't break _unless_ it was mistakenly called with a `kwarg`)

stvstnfrd: Slightly more context in the commit message: https://github.com/edx/xblock-lti-consumer/pull/188/commits/1e17937cb0532de1d26dfda2da93487b31c137ec

sarina: 👍  ty

PR #192, opened 2021-08-24
test: suppress noise

nedbat:
The test suite was generating error log messages when tests would cause
error conditions.  This cluttered the output needlessly, because it
looked like test failures, but were actually tests succeeding.  Now all
log messages are suppresed during the test suite.

In addition, the XBlock runtime generates a warning about needing
IdGenerator in the future.  I'm suppressing this warning also.

Now the test suite just outputs 397 nice dots.

### DIFF
--- a/test.py
+++ b/test.py
@@ -4,7 +4,9 @@ Run tests for the LTI Consumer XBlock
 """
 
 import os
+import logging
 import sys
+import warnings
 
 if __name__ == '__main__':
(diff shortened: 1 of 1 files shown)


nedbat:
@Jawayria what do you think?

nedbat:
@giovannicimolin isn't this better?

sarina (approved):

giovannicimolin:
@nedbat The test output is much better now. :+1:

giovannicimolin (approved):

PR #204, opened 2021-11-04
fix: Long-term fix for cross-origin iFrames

connorhaugh:
For [TNL-8584](https://openedx.atlassian.net/wiki/spaces/PT/pages/3221028876/TNL-8584+Discovery+Long-term+fix+for+cross-origin+iFrames) I need to make a proof of concept of a "Open resource in new window" modal dialog. 

Context: Our current LTI Component XBlock provides an option to "View resource in a New Window", which uses JS to launch a new window/tab. These capabilities were removed in Chrome v92. Here's some referances to those changes: TL;DR: there is a temporary cookie workaround, but this problem also started occurring on Safari 15. 

https://developer.chrome.com/origintrials/#/view_trial/2541156089743802369

. Chrome provided a temporary work-around which we now use in production. But we now need to design and implement a longer term fix.

These issues have since migrated to Safari, and “before other browsers implement this change, which they have indicated that they will” we need to resolve this issue with something other than band-aids. Google Chrome’s and Safari’s move, however, is aligned with making iFrame technology more secure, which we can read as a signal that the industry will continue to advance in the direction of iFrames.

This is that. It obviously needs some styling fixes, as well as internationalization & a11y work, but it works as expected.

screenshot:
Screen Shot 2021-11-04 at 11 30 12 AM

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -318,6 +318,12 @@ Please do not report security issues in public. Send security concerns via email
 Changelog
 =========
 
+3.0.2 - 2021-11-12
+-------------------
+
+* The modal to confirm information transfer on open of lti in new tab/window has been updated
+  because of a change in how browsers handle iframe permissions.
+
 3.0.1 - 2021-07-09
 -------------------
 
--- a/lti_consumer/static/js/xblock_lti_consumer.js
+++ b/lti_consumer/static/js/xblock_lti_consumer.js
@@ -106,17 +106,58 @@ function LtiConsumerXBlock(runtime, element) {
 
             // If this instance is configured to require username and/or email, ask user if it is okay to send them
             // Do not launch if it is not okay
+
+            function confirmDialog(message) {
+                var def = $.Deferred();
+                $('<div></div>').appendTo('body') // TODO: this will need some cute styling. It looks like trash but it works.
+                  .html('<div><h6>' + message + '</h6></div>')
+                  .dialog({
+                    modal: true,
+                    title: 'Confirm', //TODO: int8lize these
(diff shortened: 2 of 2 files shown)


giovannicimolin:
@connorhaugh Does this only impact launches to new windows/tabs? Will the change impact other types of LTI launches? 

Also, can you add any references to the Chrome/Safari change? I can't view the TNL ticket.

connorhaugh:
> @connorhaugh Does this only impact launches to new windows/tabs? Will the change (in Chrome) have any impact on other types of LTI launches?
> 
> Also, can you add any references to the Chrome/Safari change? I can't view the TNL ticket. 

Hi @giovannicimolin, Here's the chrome issue: https://www.chromestatus.com/feature/5148698084376576 https://developer.chrome.com/origintrials/#/view_trial/2541156089743802369 (a temporary workaround exists with a cookie but that cookie will be removed soon) The safari issue is new, and less documented, but it is something we have had reported from some users. 

This only affects LTI launches in a new tab/window. This should move out of the proof of concept state soon, as time is of the essence (all LTI in new window might break in January)

connorhaugh:
Note: due to the more urgent need to get a quick fix out for all users on safari, I am now looking to get this code reviewed and deployed, with improvements to the UX to follow. The Functionality is keyboard and screen reader navigable, for now, but will undergo a11y review presently. For those who are interested in a prettier fix, it will come sometime next week, but the need to get safari 15 users back on LTI is large.

kenclary (approved):

PR #210, opened 2021-11-17
BB-5051: Pluggable/reusable LTI configuration

giovannicimolin:
This ADR describes the implementation of pluggable and reusable LTI configurations.

### DIFF
--- a/docs/decisions/0006-pluggable-lti-configuration.rst
+++ b/docs/decisions/0006-pluggable-lti-configuration.rst
@@ -1,0 +1,117 @@
+Pluggable and re-usable LTI configuration
+-----------------------------------------
+
+
+Status
+======
+
+Proposal
+
+Context
+=======
+
+LTI Tools provide extra funcionality and allows course creators to embed external interactive content
+in the LMS through a standardized setup.
+
+Currently in the Open edX platform:
+
+* Each LTI tool needs to be configured individually, sometimes requiring multiple valiables to be set to get a working integration. This makes the course creation process cumbersome on courses that heavily rely on LTI components.
+* There's no support for instance-wide LTI tool configuration or any sort of pluggable LTI configurations.
+
+This proposal aims to solve these issues by introducing a mechanism to allow LTI configuration pluggability 
+and changes to the XBlock to allow re-using configurations to reduce instructor friction.
+
+
+Decision
+========
+
+The decision is to enable LTI reusability and pluggability using a filter hook (as defined in `OEP-50`_) to fetch re-usable LTI configurations
+from external plugins. 
+
+With this implementation the plugins are responsible for storing the LTI configuration and credentials, while the LTI consumer handles displaying the content.
+
+To enable this, the LTI Configuration model needs to be modified to:
+
+#. Allow externally configured tools (a new :code:`CONFIG_EXTERNAL` choice for the configuration storage).
+#. A new fields to store an namespaced tools id in the format (:code:`plugin_name:identifier`).
+#. Changes to the :code:`get_lti_consumer` methods to allow instancing a :code:`LtiConsumer` from externally managed configurations.
+
+The filter will allow the XBlock LTI Consumer to query re-usable and/or plugabble LTI configurations and display them to course 
+creators as configuration options in Studio. 
+The filter response should contain all the necessary information for the XBlock to locate and and use the externally managed LTI configuration.
+
+The inputs for the filter pipeline should be:
+
+* The course ID, so that each filter is able to evaluate and return only configurations available to that context.
+* A dictionary in which LTI configurations will be added/modified by each plugin. 
+
+Filter responses are accumulative and receive the output of the previous function (`Filter tooling - pipeline behaviour`_),
+so the filters will initially recieve an empty dictionary and each filter will incrementally add its available configurations,
+using a namespaced key with the plugin name plus an internal identifier (Example: :code:`lti_plugin_1:random-identifier`).
+
+1. Initial state of filter input
+
+.. code-block:: json
+
+    {}
+
+
+2. After the first filter is run:
+
+.. code-block:: json
+
+    {
+        "lti_plugin_1:a-plugin-interal-indentifier": {
+            "lti_version": "LTI_1P1",
+            "lti_1p1_client_key": "LTI KEY",
+            "lti_1p1_client_secret": "LTI SECRET",
+            "launch_url": "https://saltire.lti.app/tool",
+            "tool_name": "Saltire LTI test tool",
+        },
+        "lti_plugin_1:another-indentifier": {
+            ...
+        }
+    }
+
+This filter has two purposes:
+
+#. Allow the XBlock LTI Consumer to display users a list of available re-usable configurations.
+#. Return LTI configuration data so that the LTI Consumer is able to spawn a `LtiConsumer` class.
+
+When an user launches a re-usable configuration, the LTI consumer model will run the filter pipeline, retrieve the credentials if they are available,
+instance the appropriate :code:`LtiConsumer` and return it to the XBlock for usage.
+
+Related feature flags
+=====================
+
+.. list-table::
+   :widths: auto
+   :header-rows: 1
+
+   * - Toggle name
+     - Type
+     - Behavior
+   * - LTI_ENABLE_REUSABLE_CONFIGURATIONS
+     - Feature flag
+     - This flag enables LTI Configuration reusability and will display an extra field during the LTI 
+       configuration that will allow selection from pre-set configurations.
+
+
+Changes to the LTI Consumer XBlock
+==================================
+
+To allow re-usability, the XBlock will need the following changes:
+
+#. An extra configuration type will need to be added to the *LTI Version field*: :code:`Use predefined LTI configuration`
+#. When the option above is selected, a new field must be presented so that the user can pick one of the available re-usable configurations.
+#. The methods to retrieve the LTI consumer will need to be reworked to factor in re-usable configurations.
+#. Some LTI endpoints that are tied to the XBlock will need to be decoupled from the XBlock runtime urls to remove :code:`LocationKey` dependencies.
+
+
+Implementation concerns/Consequences
+====================================
+
+#. The `LtiConfiguration` model needs to be decoupled from XBlocks so that grades and student information don't get mixed between instances.
+
+.. _OEP-50: https://open-edx-proposals.readthedocs.io/en/latest/architectural-decisions/oep-0050-hooks-extension-framework.html
+.. _Filter tooling - pipeline behaviour: https://github.com/eduNEXT/openedx-filters/blob/main/docs/decisions/0003-hooks-filter-tooling-pipeline.rst


edx-webhook:
Thanks for the pull request, @giovannicimolin! I've created [OSPR-6230](https://openedx.atlassian.net/browse/OSPR-6230) to keep track of it in JIRA.

As a core committer in this repo, you can merge this once the pull request is approved per the [core committer reviewer requirements](https://openedx.atlassian.net/wiki/spaces/COMM/pages/1529675973/Open+edX+Core+Committers#Reviews-and-Ownership) and according to the agreement with your edX Champion.


docs/decisions/0006-pluggable-lti-configuration.rst
```diff
@@ -0,0 +1,116 @@
...
+                    "lti_config_id": 3,     
+                }
+            ]
+        }
+    }
+
+
+Each LTI configuration provider MUST:
+
+* Provide a `get_lti_config` method through the `api_path` parameter conforming with the API parameters described in the `Plugin API interface`.
```
giovannicimolin, line 71: @xitij2000 What I had in mind here was to store a method path in the `LtiConfiguration` models and use `importlib` to call it with well-specified interface that'll describe below.

docs/decisions/0006-pluggable-lti-configuration.rst
```diff
@@ -0,0 +1,116 @@
...
+
+
+Plugin API interface
+====================
+**`get_lti_config`**
+Should return all LTI parameters required to set up an instance of an LTI consumer, 
+`lti_consumer.lti_1p1.consumer.LtiConsumer1p1` for LTI 1.1 or `lti_consumer.lti_1p3.consumer.LtiAdvantageConsumer` for LTI 1.3.
+
+
+Changes to the LTI Consumer XBlock
```
giovannicimolin, line 103: @xitij2000 There is some code from the current implementation that will need to be reworked to allow proper re-usability for LTI 1.3. 

We'll need to move a few urls from Xblock handlers to independent django views so that we're not tying re-usable configuration to a URL that might break if a block is removed.

We can start with LTI 1.1 only as a proof of concept though :)

giovannicimolin:
@marcotuts @nedbat If I recall our discussions correctly, this ADR is aligned with the platform roadmap of having re-usable LTI tools and allowing instructors to pick from already set-up integrations.

We have a client that can partially fund this work (at least the `xblock-lti-consumer` changes), but I'm working on a broad ADR that covers both the community needs and the platform roadmap.

Feel free to pitch in and review the document (it's still WIP though!).

natabene:
@giovannicimolin Thank you for your contribution. Please let me know once this is ready.

docs/decisions/0006-pluggable-lti-configuration.rst
```diff
@@ -0,0 +1,128 @@
...
+                    "lti_config_id": 2,
+                }
+                {
+                    "lti_version": "LTI_1P3",
+                    "display_name": "Reusable LTI tool 2 - LTI 1.3",
+                    "lti_config_id": 3,     
+                }
+            ]
+        }
+    }
```
xitij2000, line 66: I'm not sure I'm understanding this correctly. What if the user has a number of different filters installed? For instance, a filter that fetches the config from the database, another from the settings, another from a vault service? 

I think for LTI 1.1 we have a pretty straightforward mechanism where we can just have the filter directly return the two string values. We should use a simpler filter for that where each filter can simply go through all the existing config options, remove them, and add new ones. 

For LTI 1.3 I guess things are more complex, but what the filter can provide are raw config values that are then copied over to the block/database, after that the block work exactly the same as now.

docs/decisions/0006-pluggable-lti-configuration.rst
```diff
@@ -0,0 +1,128 @@
...
+**`get_lti_config`**
+
+Each plugin must supply a method that returns the necessary LTI parameters settings so that the LTI Configuration 
+can instance one of the following LTI consumers from it:
+
+* :code:`lti_consumer.lti_1p1.consumer.LtiConsumer1p1` for LTI 1.1
+* :code:`lti_consumer.lti_1p3.consumer.LtiAdvantageConsumer` for LTI 1.3.
+
+A method in LTI Configuration will be responsible for instancing the LTI Consumer class and returning it to the XBlock.
+It'll also handle creating any acessory models to enable LTI Advange services.
```
xitij2000, line 109: I think for LTI1.1 at least we can just directly return this from the filter, we don't need to call `get_lti_config` since the filter will already know if there are any credentials configured for this block.

Instead of returning the configured consumer (which will involve each filter having a lot of the same boilerplate) we can just have the filter return a configuration object, which can then be passed to a function that builds the consumer from that.

docs/decisions/0006-pluggable-lti-configuration.rst
xitij2000: I think we can minimise changes using the approach I suggested, which boils down to:

1. Use the filter API as a setup mechanism, not a source of credentials, i.e. it gives you a list of sources for credentials for the block, but once you pick them they are copied over to the DB. There aren't many use cases for needing to change this once it's been set up, but even if they want to they can just edit the block again. 
2. Since the configuration is once again stored on the block or database the rest of the code for the block can remain the same. 

The filter here can be provided the course key, the usage key of the block, the currently-selected configuration, the other available configurations etc.

docs/decisions/0006-pluggable-lti-configuration.rst
giovannicimolin: @xitij2000 I agree with the fact that there will be some boilerplate code in the filters, my initial goal was to reduce the complexity inside the core by removing the logic of configuring the LTI consumer from it (look at the LTI 1.3 method: https://github.com/edx/xblock-lti-consumer/blob/master/lti_consumer/models.py#L247, it's big).

Also, if we retrieve configuration/credentials directly using the filter instead of calling a filter method, we'll have to run the filter pipeline every time the LTI content is launched or one of it's APIs is used (which is less performant than calling a direct function and retrieving the configuration).

What to you think of: 
1. Removing the need for the plugins to return the LTI consumer (just return the credentials + any additional variables required) and leave the LTI consumer instancing to model method/helper.
2. Keep retrieving the credentials using a function name + ID instead of running the filter pipeline on each LTI launch.

docs/decisions/0006-pluggable-lti-configuration.rst
giovannicimolin: @xitij2000 I need to improve the wording here.

I though of "pluggable LTI configurations" as a combination of two things: 
1. One filter pipeline that solely exists to retrieve information about where to fetch a given LTI configuration + some metadata to display to the end user.
2. A standard python API that the xblock-lti-consumer would call to retrieve the actual configuration when LTI launches happened. We can replace the standard API with a second (different) filter pipeline that returns the LTI configuration given a UUID.
  
>What if the user has a number of different filters installed? For instance, a filter that fetches the config from the database, another from the settings, another from a vault service?

The result of the filter will simply contain more results (one for each LTI configuration available). This filter isn't supposed to return the LTI credentials/configuration, but the list of available configurations and how to retrieve data from them.

>For LTI 1.3 I guess things are more complex, but what the filter can provide are raw config values that are then copied over to the block/database, after that the block work exactly the same as now.

:+1: for the filter to provide settings, but not for them to be stored in the database as it wouldn't fulfill the client's requirements.

giovannicimolin:
@xitij2000 After re-thinking what I had in mind initially, I think we can reduce this entire implementation down to a single filter with a few parameters and push the boilerplate code to `xblock-lti-consumer`.

docs/decisions/0006-pluggable-lti-configuration.rst
```diff
@@ -0,0 +1,121 @@
...
+        },
+        ...
+    }
+    
+
+Once a user selects one of the re-usable configurations, the LTI Configuration model is responsible for storing the plugin name and the UUID 
+of the configuration.
+
+When an user launch's a re-usable configuration, the LTI consumer model will run the filter pipeline, retrieve the credentials if they are available,
+instance the appropriate :code:`LtiConsumer` and return it to the XBlock for usage.
```
xitij2000, line 87:
```suggestion
When an user launches a re-usable configuration, the LTI consumer model will run the filter pipeline, retrieve the credentials if they are available,
instance the appropriate :code:`LtiConsumer` and return it to the XBlock for usage.
```

docs/decisions/0006-pluggable-lti-configuration.rst
xitij2000: Sure, what you're saying makes sense. However, I do feel that these credentials aren't rotated as often. So if they need to be rotated, we can provide a different mechanism to bulk-update credentials. 

That said, I think it's probably not what was needed here, so it makes sense to keep the hook at runtime rather than configuration time.

xitij2000 (approved):
:+1: This is looking good.  Just a very minor nit. 

- [x] I read through the ADR
- [x] Includes documentation

docs/decisions/0006-pluggable-lti-configuration.rst
```diff
@@ -0,0 +1,121 @@
...
+#. Changes to the :code:`get_lti_consumer` methods to allow instancing a :code:`LtiConsumer` from externally managed configurations.
+
+The filter will allow the XBlock LTI Consumer to query re-usable and/or plugabble LTI configurations and display them to course 
+creators as configuration options in Studio. 
+The filter response should contain all the necessary information for the XBlock to locate and and use the externally managed LTI configuration.
+
+The inputs for the filter pipeline should be one of the following:
+
+* The course ID, so that each filter is able to evaluate and return only configurations available to that context
+* The unique configuration ID, so the filter returns exactly the configuration required.
```
xitij2000, line 49: I think we should keep in mind how filters operate. The output of a filter should match the input. i.e. you can't just pass a course_id to a filter. What I mean is, filters will run on a sequence, the output from the first filter function will be passed to the next filter function. 

So, for example, you will need to pass to the first filter the `context_key` and a list of `lti_credentaisl`. This filter will then pass on the same parameters, but will add entries to  "lti_credentials" (or remove them).

You don't need to pass the configuration ID. The filter pipeline can just return all valid results, and you can pick the one that matches the stored id.

docs/decisions/0006-pluggable-lti-configuration.rst
```diff
@@ -0,0 +1,121 @@
...
+                        // LTI consumer parameters and configuration
+                    }     
+                }
+            ]
+        },
+        "another_plugin": {
+            "available_configurations": [...]
+        },
+        ...
+    }
```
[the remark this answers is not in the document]
giovannicimolin: @xitij2000 Makes sense to store the configuration on a single field. :+1:
The unique is basically a UUID that is managed by the external tool/plugin, so to avoid the remote possibility of a conflict between two plugins, I've added the extra field.
I'll change the format to `plugin_name:uuid` (plugin_name as in the pipeline function name that is used).

docs/decisions/0006-pluggable-lti-configuration.rst
```diff
@@ -0,0 +1,121 @@
...
+    
+
+Once a user selects one of the re-usable configurations, the LTI Configuration model is responsible for storing the plugin name and the UUID 
+of the configuration.
```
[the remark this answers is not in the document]
giovannicimolin: @xitij2000 I'm thinking of an use case where there's more than one plugin providing LTI configurations at the same time. I know that the possibility of a conflict is low, but I wanted to remove it entirely.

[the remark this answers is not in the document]
xitij2000: If we are namespacing it, we can even allow non-UUID values, i.e. maybe a friendly name like: piazza-main-credentials etc.

docs/decisions/0006-pluggable-lti-configuration.rst
giovannicimolin: @xitij2000 
>I think we should keep in mind how filters operate. The output of a filter should match the input. i.e. you can't just pass a course_id to a filter. What I mean is, filters will run on a sequence, the output from the first filter function will be passed to the next filter function.

That makes sense, should I come up with a django object and let the filters modify it and pass it forward. I found the decision about the accumulative approach [here](https://github.com/eduNEXT/openedx-filters/blob/main/docs/decisions/0003-hooks-filter-tooling-pipeline.rst), but had to read the source to see how it actually worked (here: https://github.com/eduNEXT/openedx-filters/blob/main/openedx_filters/tooling.py#L207).

There's two interesting things there: (1) filter metadata, always passed when instancing the filter and (2) the accumulated output. We can just trigger the pipeline with an empty dict input `{}` and the context key as metadata and let the filters fill and modify it to add LTI configurations. Is my interpretation correct?

The filters payload is still in the "draft" state, but it mentions this:
>Although filters have a public promise status, and thus maintainability is an important design goal, filters by definition permit that extension developers manipulate data that will be passed along to the lms or cms processes.
At the same time, using fixed attr classes such as what is used in events is not convenient when trying to have few restrictions on what developers can do.

I understood that can take some liberties with how we wanted the return format to be, as long as the response is a dict and the filter and takes the accumulated output of previous functions.

>You don't need to pass the configuration ID. The filter pipeline can just return all valid results, and you can pick the one that matches the stored id.

:+1:

docs/decisions/0006-pluggable-lti-configuration.rst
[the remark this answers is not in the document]
giovannicimolin: Makes sense! I'll remove the requirement of it being an uuid. :)

docs/decisions/0006-pluggable-lti-configuration.rst
xitij2000: Yes, we can trigger it with an empty dict, have the filters manipulate the data along the way, and return the data we want. 

I think the idea is that filters could come in any order (you can control the order, but as a filter developer you don't know the order), so the output from one filter will go to the next filter, so it should be able to process it. A few extra keys in a dict shouldn't break a step in the pipeline.

giovannicimolin:
@xitij2000 I addressed your comments and updated the ADR.

@natabene @nedbat @marcotuts This is ready for review now. :grin:

robrap:
Thanks @giovannicimolin! I've been told that you are basically the LTI expert.
1. [question] @marcotuts: Let us know if you have any thoughts or this is good to go on your end.
2. [question] @ormsbee: You may have enough context to make some comments on the approach. Otherwise, we'll just let Giovanni drive this. Let us know either way. Thanks.
3. [inform] @giovannicimolin: No action required on this, but you can use `squash! some commit title` to have your commit messages pass linting if you wish. See https://open-edx-proposals.readthedocs.io/en/latest/best-practices/oep-0051-bp-conventional-commits.html#discussion. Otherwise, you can just wait until you actually squash to get it to pass.

giovannicimolin:
@robrap Sorry for the delay in getting back to you, I was off for the past few weeks and I don't think the auto-responder works on on GitHub pings. 

Thanks for the follow-up and feedback from the team! I'll pick this up again starting next sprint (Jan 11) and move the efforts forward :grin:

regisb:
Hi @giovannicimolin! I'm wondering if there is an expected timeline for implementing this ADR? I'm currently working on a project that would greatly benefit from a better extensibility of the LTI Configuration model.

giovannicimolin:
@robrap @schenedx @regisb I reserved some time to clean up and get a final review pass for this ticket next sprint and merge it. 

>I'm wondering if there is an expected timeline for implementing this ADR? I'm currently working on a project that would greatly benefit from a better extensibility of the LTI Configuration model.

Last I've heard we already have the budget secured to implement this feature, just needed to wait for `openedx-filters` + internal capacity to implement it, but I'll double check with the Epic owner now and get back to you soon.

giovannicimolin:
@regisb 
>I'm wondering if there is an expected timeline for implementing this ADR? I'm currently working on a project that would greatly benefit from a better extensibility of the LTI Configuration model.

No expected timeline, from what I understood the client that needed this feature decided to postpone it's implementation to a later time. That said, we could still drive this implementation using community resources (I mostly spend my CC time here - so I can help implement and review changes). 

CC @Cup0fCoffee as the epic owner.

giovannicimolin:
@xitij2000 Addressed comments and fixes RST syntax issues, this is ready for review.

schenedx (approved):
Looks good to me. I am interested to see how the implementation would look and its impact to the existing UI.

giovannicimolin:
@schenedx Thanks for the review!  

>I am interested to see how the implementation would look and its impact to the existing UI.

I have a few ideas, but ultimately we'll need edX/2U/tCRIL product and technical approval before merging it.

There's a few steps to implement this ADR - we'll probably split the implementation in bite-sized PRs to make review and testing easier.

giovannicimolin:
Awesome! I'll merge this PR now then.

@schenedx @regisb I'll ping you as soon as we start developing this feature.

edx-webhook:
@giovannicimolin 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/xblock-lti-consumer&entry.867055334=https://github.com/openedx/xblock-lti-consumer/pull/210&entry.1484655318=https://github.com/giovannicimolin&entry.752974735=2021-11-17+17:59&entry.1917517419=2022-02-21+14:40&entry.2133058324=Yes) so we can improve your experience in the future.

regisb:
What's the status of this ADR? Is it on someone's roadmap?

Zacharis278:
@regisb this has actually been added already with https://github.com/openedx/xblock-lti-consumer/pull/239

giovannicimolin:
Complementing @Zacharis278's answer: https://github.com/openedx/xblock-lti-consumer/pull/239 added support for LTI 1.1 pluggability. 
LTI 1.3 pluggability support was supposed to be funded by a client that ended up dropping the budget - so the work on this is now stopped.

PR #227, opened 2022-01-31
fix: use launch_url for oidc login target_link_uri parameter

schenedx:
@giovannicimolin I think you would need to review this change.

The `target_link_uri` specified as part of the oidc login preflight url was always edX LTI launch URL. 
According to https://www.imsglobal.org/spec/lti/v1p3#target-link-uri, it needs to be the tool's launch URL instead.

This is because the claim message should have the field `target_link_uri` as the same value as the oidc login preflight parameter.

Even the unit tests on the repository reflect this requirement, therefore I didn't need to write any new unit tests.

@openedx/masters-devs-cosmonauts Please review


lti_consumer/plugin/compat.py
```diff
@@ -60,7 +60,6 @@ def load_block_as_anonymous_user(location):
             student_data=None,
             course_id=location.course_key,
             track_function=None,
-            xqueue_callback_url_prefix="",
```
schenedx, line 63: The code here was changed quite a while ago on edx-platform. Just want to align with edx-platform here.

ashultz0 (approved):

Zacharis278 (approved):
👍

giovannicimolin:
@schenedx Before merging this, can you run the LTI validation suite from IMS? The current code passes it - I think I've bumped into an issue that forced me to change to the current approach.

There's instructions here: https://openedx.atlassian.net/wiki/spaces/COMM/pages/1858601008/How+to+run+the+LTI+Validation+test

Only edX is an IMS member, I don't have access to the validation tool. IIRC @nedbat was handling the IMS accounts?

schenedx:
@giovannicimolin Sounds good. Let me do the validation you pointed out.

schenedx:
@giovannicimolin I made the test using IMS Validator with my code in this PR and it passed the test.
Let me know if you are good with what I have.

giovannicimolin (approved):
:+1:

- [x] I tested this: 
1. I tested these changes using the IMS reference implementation tool (tool: https://lti-ri.imsglobal.org/lti/tools/2618/)
- [x] I read through the code
- [x] Includes documentation: changelog updated :heavy_check_mark: 

@schenedx Good to go! Works perfectly with the reference implementation.

PR #239, opened 2022-03-15
[BB-5559] Adds support for external LTI configurations using openedx-filters

tecoholic:
### Background

Presently the LTI Component XBlock supports integration of LTI Tools by adding the necessary configuration values directly in the XBlock settings. This means things like the launch URL, tool id for Version 1.1 and Login URL, public key and other parameters for version 1.3 need to be entered at every place the XBlock is used.

This causes friction for the course creators while creating and updating the course, as each instance of the XBlock needs to be edited individually.

### Description

This PR is based on the [ADR](https://github.com/openedx/xblock-lti-consumer/blob/master/docs/decisions/0006-pluggable-lti-configuration.rst) aimed at reducing the friction by providing a way for the XBlock to fetch the necessary information for an LTI tool from an external location. 

The PR introduces the following changes:

1. This is achieved by using the events API provided by [openedx-filters](https://github.com/openedx/openedx-filters). We introduce a new filter in the XBlock: ```org.openedx.xblock.lti_consumer.configuration.listed.v1```
    -  As a proof of concept, a new plugin is also created called [openedx_ltistore](https://github.com/open-craft/openedx-ltistore/pull/1) which implements storing of the LTI configurations in the DB and the filter pipeline to return configurations when the filter is run has also been created.
 2. Adds an extra option `XBLOCK_EXTERNAL` to the `LtiConfiguration` model's `config_store` attribute.
 3. Adds an extra option `External Configuration` to the `LTI Version` select in the **Edit** form of the XBlock, which will allow the user to select an external configuration instead of specifying the version and entering related information.
 4. All of the changes are gated behind a Course level Waffle Flag - `lti_consumer.enable_external_config_filter`

### Testing Instructions

The testing instructions you are using the **openedx_devstack** to test.

1. Setup the [openedx_ltistore](https://github.com/open-craft/openedx-ltistore/pull/1) plugin as described in the README and add a couple of configurations with version 1.1. 
    - **Note:** [Saltire](https://saltire.lti.app/) can be used to get some configuration for testing
2. Create a sample configuration with LTI version 1.1 in the [LTI_Store admin|http://localhost:18010/admin/lti_store/] note down the **Filter Key** value of the newly created configuration.We will use this later. eg., `lti_store:1`
3. Setup `xblock-lti-consumer` in the development mode
4. Checkout to this PR branch
5. Restart studio using `make studio-restart`
6. Create a test course, if you don't already have one.
7. Enable the external config feature by creating a course waffle flag in the [admin page](http://localhost:18010/admin/waffle_utils/waffleflagcourseoverridemodel/)
    - FLAG:  `lti_consumer.enable_external_config_filter`
    - Course: `<your course id>`
    - Force: ON
    - Enabled
9. Add `lti_consumer` to the **Advanced Module List** in the **Advanced Settings** of the test course.
10. Add a new unit to the course and add the **LTI Consumer** from the **Advanced** blocks.
11. Click the **Edit** button of the LTI Consumer block and set **Configuration Type** to `Existing Configuration`
12. Set the **External configuration ID** to the filter key value we copied when we setup LTI Store. eg., `lti_store:1`
14. Click Save. 
15. The LTI Tool should load as expected.

### Additional Information

1. [ADR of the implementation](https://github.com/openedx/xblock-lti-consumer/blob/master/docs/decisions/0006-pluggable-lti-configuration.rst)
2. [Compatible Openedx Plugin app - openedx-ltistore](https://github.com/open-craft/openedx-ltistore)


edx-webhook:
Thanks for the pull request, @tecoholic! I've created [OSPR-6522](https://openedx.atlassian.net/browse/OSPR-6522) to keep track of it in JIRA, where we prioritize reviews. Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket as you can:
- supporting documentation

lti_consumer/lti_xblock.py
```diff
@@ -700,6 +711,10 @@ def lti_provider_key_secret(self):
         """
         Obtains client_key and client_secret credentials from current course.
         """
+        if self.lti_version == "external":
```
giovannicimolin, line 714: Why do you need the secret here?

giovannicimolin (changes requested):
A few changes are needed, but this looks promising. Kudos! :smile: 

The main big point here is keeping as much separation as possible between each layer (xblock <--> django app), so we'll have to streamline/simplify the credential fetching mechanism. I'll add more comments soon, ran out of time for today.

lti_consumer/filters.py
```diff
@@ -0,0 +1,54 @@
+"""
+Module that contains the openedx filters for this XBlock
+"""
+from typing import Dict
+
+from openedx_filters.tooling import OpenEdxPublicFilter
```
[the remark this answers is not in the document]
tecoholic: Pip was complaining about something when I added it, so I installed in manually. I was supposed to fix pip issue and seemed to have missed it. Thank you for noticing. Will add it.

lti_consumer/models.py
```diff
@@ -63,14 +64,16 @@ class LtiConfiguration(models.Model):
     )
 
     # Configuration storage
-    # Initally, this only supports the configuration
-    # stored on the block, but should be expanded to
-    # enable storing LTI configuration in this model.
+    # Initally, this only supported the configuration
+    # stored on the block. Now it has been expanded to
+    # enable storing LTI configuration in the model and
```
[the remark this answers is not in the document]
tecoholic: Yup.. Will complete the sentence :joy:

lti_consumer/models.py
```diff
@@ -237,6 +240,11 @@ def _get_lti_1p1_consumer(self):
         if self.config_store == self.CONFIG_ON_XBLOCK:
             key, secret = self.block.lti_provider_key_secret
             launch_url = self.block.launch_url
+        elif self.config_store == self.CONFIG_EXTERNAL:
+            config = get_external_config_from_filter({}, self.block.external_config)
```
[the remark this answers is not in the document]
tecoholic: If it is not stored at the block level, how do we get the user to select the external configuration from the Edit menu? As the parameters of the XBlock are automatically saved at the block level, the external config's id would also be stored alongside right?

lti_consumer/lti_xblock.py
tecoholic: Not sure exactly. I just added this block to maintain consistency with the previous logic of returning the key and secret. The only place I see this function `lti_provider_key_secret` seems to in a debug call.

lti_consumer/models.py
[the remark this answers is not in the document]
giovannicimolin: I want to create some separation between the pluggable interface and the XBlock. You can pass the external config id to `_get_or_create_local_lti_config` and update if it changed in the XBlock (the same way the version is updated when updated - [see code here](https://github.com/openedx/xblock-lti-consumer/pull/239/files#diff-6e408862179502a0ac809fab914e028b18311a17a3b0874469bd6416bc4a449fR39)).

XBlock <--- config_id/XBlock usage key ---> LtiConfiguration <---- external_config_id ---> pluggable filters

The important thing is to have the LTI client instanced using only values from the model and the filter pipeline, without requiring an instanced version of the block.

lti_consumer/lti_xblock.py
giovannicimolin: `lti_provider_key_secret` is only used when the configuration is set to `CONFIG_ON_XBLOCK`, [here](https://github.com/openedx/xblock-lti-consumer/blob/master/lti_consumer/models.py#L238) so there's no to modify this (the `LtiConfiguration` model ensures that the `LtiConsumer` client has the consistent configuration, so it's his job to determine where to fetch the credentials from).

lti_consumer/lti_xblock.py
```diff
@@ -1465,7 +1481,11 @@ def _get_context_for_template(self):
...
         sanitized_comment = bleach.clean(self.score_comment, tags=allowed_tags, attributes=allowed_attributes)
 
-        if self.lti_version == 'lti_1p1':
+        if self.lti_version == "external":
+            config = get_external_config_from_filter({}, self.external_config)
+            self.launch_url = config.get('lti_1p1_launch_url')
+
+        if self.lti_version == 'lti_1p1' or self.lti_version == "external":
             # Set launch handler depending on LTI version
             lti_block_launch_handler = self.runtime.handler_url(self, 'lti_launch_handler').rstrip('/?')
```
giovannicimolin, line 1490: Instead of running the filter here and adding a dependency to the XBlock code, instead retrieve the LTI consume and get the URL from the class:

```suggestion
        if self.lti_version == 'lti_1p1':
            consumer = self._get_lti_consumer()
            
            lti_block_launch_handler = consumer.lti_launch_url.rstrip('/?')
```
While initially, this does look more convoluted, it's simplifying the interface between the XBlock and Django app while delegating the responsibility of a consistent configuration to the `LtiConfiguration` model through `get_lti_consumer`.

lti_consumer/lti_xblock.py
```diff
@@ -264,6 +266,15 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
             "LTI 1.3 and LTI Advantage features."
         ),
     )
+
+    external_config = String(
+        display_name=_("External configuration"),
+        scope=Scope.settings,
+        values=get_external_config_values,
+        help=_("Select the External configuration for the LTI Tool")
+    )
```
tecoholic, line 275: @giovannicimolin While working on removing direct dependencies between the XBlock and the Filters, I got held up here. Here `get_external_config_values` allows the XBlock to fetch a list of available LTI configurations from plugins to display it to the user in the "Edit" page.

Considering your other comment about passing `course_id` to the filters, we can't use this function reference to `values` anymore. So I am thing of adding a AJAX XBlock handler, that will allow the front-end to fetch the list of available LTI Tools. This will allow the XBlock to pass the `course_id` to the `get_external_config_values` function.

But, this still means, there will be a link between XBlock and the Filters. Do you have any idea about how we can get this list without XBlock depending on Filters?

giovannicimolin: Ah, I think I understand the problem now. 

What I had in mind for this field was to have it as a simple `string` field, and [modify the context](https://github.com/openedx/xblock-lti-consumer/blob/master/lti_consumer/lti_xblock.py#L942) or the [method that prepares the fields](https://github.com/openedx/xblock-utils/blob/master/xblockutils/studio_editable.py#L96) to fool the template into rendering a select instead of text input. But I think this is a bit too much for the first iteration.

Let's avoid the complexity of dealing with the XBlock studio view rendering for now: make this a simple `String` field.
Make sure this is gated behind a WaffleFlag, so the option and associated field don't appear if it's globally disabled (use something like this: https://github.com/openedx/xblock-lti-consumer/commit/92cf504eba66577103e3bd6650bdc58a9fa9ffa9).

lti_consumer/lti_xblock.py
```diff
@@ -249,13 +251,13 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
...
     )
 
-    # LTI 1.3 fields
     lti_version = String(
         display_name=_("LTI Version"),
         scope=Scope.settings,
         values=[
             {"display_name": "LTI 1.1/1.2", "value": "lti_1p1"},
             {"display_name": "LTI 1.3", "value": "lti_1p3"},
+            {"display_name": "External Configuration", "value": "external"},
```
giovannicimolin, line 260: @tecoholic Which of the options below do you think is better?
- This solution, with the `external` value here
- A separate `List` variable, where the options are either `New configuration` or `Use external configuration`

lti_consumer/lti_xblock.py
tecoholic: @giovannicimolin Oh! I too tried fooling the system and failed :) I populated the `values` for the field in the `studio_view` before creating the fragment, but somehow the change never persisted. I didn't use any of the XBlock's internal functions though, simply updated the `values` list. Let's keep it simply as a string field as you say. We will get to this UX part later :)

natabene:
@giovannicimolin Are you interested in taking this over the finish line (finalize review and merge)?

lti_consumer/lti_xblock.py
tecoholic: @giovannicimolin This one kind of tripped me up because there is:
1. `launch_url` - which corresponds to the URL of the tool we get from the configuration 
2. there is the the `self.runtime.handler_url` which is the XBlocks launch handler URL.

I [updated the first one](https://github.com/openedx/xblock-lti-consumer/pull/239/files#diff-cc618c2e63b8c2d293d245f9799381d9321e40fe2d05e9a4b2f80bb91995157dR1497) as per the above suggestion, the second one I have left it as it is. Fetching the XBlock's handler URL from the runtime seems appropriate for Version 1.1.

lti_consumer/lti_xblock.py
```diff
@@ -1465,7 +1494,10 @@ def _get_context_for_template(self):
         allowed_attributes = dict(bleach.sanitizer.ALLOWED_ATTRIBUTES, **{'img': ['src', 'alt']})
         sanitized_comment = bleach.clean(self.score_comment, tags=allowed_tags, attributes=allowed_attributes)
 
-        if self.lti_version == 'lti_1p1':
+        if self.config_type == "external":
+            self.launch_url = self._get_lti_consumer().lti_launch_url
+
+        if self.lti_version == 'lti_1p1' or self.config_type == 'external':
```
giovannicimolin, line 1500: Can you add a comment here explaining why we need to check if it's external for now?

lti_consumer/models.py
```diff
@@ -63,21 +64,32 @@ class LtiConfiguration(models.Model):
...
         choices=CONFIG_STORE_CHOICES,
         default=CONFIG_ON_XBLOCK,
     )
 
+    # ID of the configuration if the configuration is obtained from the
+    # external service using filters
+    external_id = models.CharField(
+        max_length=255,
+        blank=True,
+        null=True
```
giovannicimolin, line 90: Django will set this as an empty string by default, no need for the null value here.
```suggestion
```

giovannicimolin:
@natabene Yes, I'm reviewing this as CC.

@schenedx Since this is a big change, can you or someone in your team do a review pass once I give a :+1: here?

lti_consumer/lti_xblock.py
```diff
@@ -248,8 +248,21 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
         default="",
         scope=Scope.settings
     )
+    config_type = String(
+        display_name=_("Configuration Type"),
+        scope=Scope.settings,
+        values=[
+            {"display_name": _("New Configuration"), "value": "new"},
+            {"display_name": _("Existing Configuration"), "value": "external"},
```
giovannicimolin, line 256: The wording here needs to be improved, this can cause some confusion.

lti_consumer/lti_xblock.py
```diff
@@ -264,6 +277,14 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
             "LTI 1.3 and LTI Advantage features."
         ),
     )
+
+    external_config = String(
+        display_name=_("External configuration ID"),
+        scope=Scope.settings,
+        help=_("Enter the External configuration for the LTI Tool")
```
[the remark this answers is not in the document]
giovannicimolin: @marcotuts Same as the other comment: Is the wording here something you can help us with?

lti_consumer/lti_xblock.py
giovannicimolin: I think this calls for product help. @marcotuts Is the wording here something you can help us with?

- New configuration: It's actually when authors configure LTI directly on the block (maybe something like `On block`?).
- Existing Configuration: External configuration, where edX staff/admins provide an LTI config string, which is reusable across the courses.

giovannicimolin (changes requested):

schenedx:
@giovannicimolin Sounds good. I'll get someone on my team to review.

lti_consumer/lti_xblock.py
```diff
@@ -641,7 +662,15 @@ def editable_fields(self):
...
-        editable_fields = self.editable_field_names
+        waffle_blocked = ['config_type', 'external_config']
+        if external_config_filter_enabled(self.location.course_key):
+            editable_fields = self.editable_field_names
+        else:
+            editable_fields = [
+                field
+                for field in self.editable_field_names
+                if field not in waffle_blocked
+            ]
```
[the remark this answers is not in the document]
tecoholic: @giovannicimolin This was a very nice catch. I didn't notice the reference to the `self.editable_field_names` to the inside the loop.

lti_consumer/lti_xblock.py
```diff
@@ -1465,7 +1494,10 @@ def _get_context_for_template(self):
         allowed_attributes = dict(bleach.sanitizer.ALLOWED_ATTRIBUTES, **{'img': ['src', 'alt']})
         sanitized_comment = bleach.clean(self.score_comment, tags=allowed_tags, attributes=allowed_attributes)
 
-        if self.lti_version == 'lti_1p1':
+        if self.config_type == "external":
+            self.launch_url = self._get_lti_consumer().lti_launch_url
```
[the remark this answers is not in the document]
tecoholic: Once reason why I was modified the class attribute was to make sure the `launch_url` was available where ever the property was accessed. I didn't realize the LMS runtime has a restriction to do this. I have used a local variable to handle this as you have suggested. Thank you.

lti_consumer/models.py
tecoholic: @giovannicimolin Removing the `null=True` trips up a lot of the tests with 
```
django.db.utils.IntegrityError: NOT NULL constraint failed: lti_consumer_lticonfiguration.external_id
```
So, I have left it unchanged.

tecoholic:
@giovannicimolin I have addressed the issues you pointed out except one. I have added the explanation of why it couldn't be done. Once we get clarity on the wording for the dropdown, I will update the PR and let you know. Thanks for the review. Some real gotchas there :)

tecoholic:
@giovannicimolin The CI tests are getting cancelled because of the Django 4.0 tests failing. Since there is a requirement pin on Django 3.2, shouldn't Django 4 be removed from the test matrix?

lti_consumer/lti_xblock.py
tecoholic: @marcotuts Can you kindly take a look at this when you have some time? Your inputs would be really helpful in moving this PR forward.

giovannicimolin:
@tecoholic I updated the branch, let's check if tests pass now.

tecoholic:
> @tecoholic I updated the branch, let's check if tests pass now.

@giovannicimolin I found the issue. I have used `pip-compile` instead `make upgrade` to update the requirement files when I added `openedx-filters` to the base file. This had added django 3.2 to the version to the `ci.txt` file making it to clash with tox's Django 4 requirement. 

I have updated the requirements file using `make upgrade` now. Can you kindly approve the workflow now? I think this time it should pass.

lti_consumer/api.py
```diff
@@ -53,10 +59,22 @@ def _get_lti_config(config_id=None, block=None):
...
+            config = get_external_config_from_filter(
+                {"course_key": block.location.course_key},
+                block.external_config
+            )
+            lti_config = _get_or_create_local_lti_config(
+                config.get("version"),
+                block.location,
+                block.external_config
+            )
+            lti_config.lti_config = config
```
giovannicimolin, line 72: @tecoholic Why is this line? We are always using the filter pipeline to retrieve the configuration, so this can become outdated when the credentials are updated in the LTI Store plugin.

lti_consumer/lti_xblock.py
giovannicimolin: @tecoholic Lets change this to:

- `Configuration on block`: default value - config on Xblock.
- `Reusable configuration`: new setting we are introducing with this PR.

lti_consumer/lti_xblock.py
```diff
@@ -248,8 +248,21 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
...
+        scope=Scope.settings,
+        values=[
+            {"display_name": _("New Configuration"), "value": "new"},
+            {"display_name": _("Existing Configuration"), "value": "external"},
+        ],
+        default="new",
+        help=_(
+            "Select 'New Configuration' to enter the details for a LTI Tool Provider. "
+            "If you don't have configuration details for a LTI Tool, select 'Existing Configuration'"
+            " to use one of the pre-configured tools."
```
giovannicimolin, line 262:
```suggestion
            "If the support staff provided you with a pre-configured LTI reusable Tool ID, select "
            "'Reusable Configuration' and enter it in the field below."
```

lti_consumer/lti_xblock.py
```diff
@@ -248,8 +248,21 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
...
+    config_type = String(
+        display_name=_("Configuration Type"),
+        scope=Scope.settings,
+        values=[
+            {"display_name": _("New Configuration"), "value": "new"},
+            {"display_name": _("Existing Configuration"), "value": "external"},
+        ],
+        default="new",
+        help=_(
+            "Select 'New Configuration' to enter the details for a LTI Tool Provider. "
```
giovannicimolin, line 260:
```suggestion
            "Select 'Configuration on block' to configure a new LTI Tool. "
```

lti_consumer/lti_xblock.py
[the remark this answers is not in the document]
giovannicimolin: @tecoholic Change this to:
**Name:** `LTI Reusable Configuration ID`
**Help:** `Enter the reusable LTI external configuration ID provided by the support staff.`

lti_consumer/lti_xblock.py
```diff
@@ -642,6 +663,16 @@ def editable_fields(self):
...
+
+        # If the support for external configuration is not enabled for the course
+        # then let's remove the external config related fields from edit form
+        waffle_blocked = ['config_type', 'external_config']
+        if not external_config_filter_enabled(self.location.course_key):
+            editable_fields = tuple(
+                field
+                for field in editable_fields
+                if field not in waffle_blocked
+            )
```
giovannicimolin, line 675:
```suggestion
        if not external_config_filter_enabled(self.location.course_key):
            editable_fields = tuple(
                field
                for field in editable_fields
                if field not in ['config_type', 'external_config']
            )
```

giovannicimolin (changes requested):
@tecoholic Left small comments and nits - I'll approve once they are addressed.

lti_consumer/api.py
tecoholic: I agree. We don't need this.

tecoholic:
@giovannicimolin I have made the changes based on your comments and fixes the quality issues identified in CI tests. Kindly review.

tecoholic:
@giovannicimolin I noticed that the test coverage was falling below 97%. The code was missing for the parts with "external" configuration. I have added test for `api.py` and `models.py` code blocks which handle the "external" configuration type. Kindly review and rerun the coverage check.

giovannicimolin (approved):
:+1:

- [x] I tested this:
1. Followed testing instructions, and checked that re-usable configuration works as expected.
2. Does not impact older integrations.
- [x] I read through the code
- [ ] ~~Includes documentation~~

@tecoholic Great work here! This is working perfectly for me :grin: 

This is conditionally approved on minor things:
1. PR is rebased and has no conflicts.
2. A major version bump is made (`3.x` -> `4.0.0`)
3. The version release notes are updated.

Zacharis278 (approved):
Got this setup and everything looks good. Thanks for the easy setup instructions!

I'll second the call for a version bump / release notes on this feature but otherwise approved.

tecoholic:
@giovannicimolin @Zacharis278 YAY!! Wonderful. I have updated the PR by:

1. Version is set to `4.0.0`
2. Change-log in `README.rst` updated with details
3. PR rebased on master, there were merge conflicts in requirements files, so requirements have been update with `make upgrade`
4. Commits have been squashed.

Kindly merge :)

openedx-webhooks:
@tecoholic 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/xblock-lti-consumer&entry.867055334=https://github.com/openedx/xblock-lti-consumer/pull/239&entry.1484655318=https://github.com/tecoholic&entry.752974735=2022-03-15+14:25&entry.1917517419=2022-05-09+14:18&entry.2133058324=Yes) so we can improve your experience in the future.

PR #245, opened 2022-05-02
remove unneeded constraints

ashultz0:
constraints file had a couple of entries "matching platform" which platform must have removed a long time ago. Cleaning them up didn't actually change anything in the requirements.txt files but cleaning is worth doing anyway.


requirements/common_constraints.txt
```diff
@@ -20,6 +20,3 @@ Django<4.0
 elasticsearch<7.14.0
 
 setuptools<60
-
-# redis 4 client doesn't play nicely with redis 3 server
-redis<4
```
alangsto, line 25: Do we know if we're not using a redis 3 server anymore?

ashultz0: common_constraints is actually copied out of https://github.com/openedx/edx-lint/blob/master/edx_lint/files/common_constraints.txt by make upgrade, it looks like the redis constraint was recently removed

alangsto: Nice

alangsto (approved):
👍

PR #246, opened 2022-05-06
feat: implement changes to accomodate proctoring specification POC

MichaelRoytman:


lti_consumer/lti_1p3/consumer.py
```diff
@@ -301,7 +306,6 @@ def get_lti_launch_message(
             # Extra claims - From LTI Advantage extensions
             if self.extra_claims:
                 lti_message.update(self.extra_claims)
-
```
MichaelRoytman, line 304: TODO: fix this unnecessary removal

lti_consumer/lti_1p3/README.md
```diff
@@ -100,13 +100,13 @@ def lti_launch_endpoint(request):
 
     # Required user claim data
     lti_consumer.set_user_data(
-      user_id=request.user,
+      user_id=request.user.id,
```
MichaelRoytman, line 103: I'm just updating the README with changes I had to make when borrowing this code.

lti_consumer/lti_1p3/consumer.py
```diff
@@ -540,18 +555,57 @@ def enable_deep_linking(
         """
         self.dl = LtiDeepLinking(deep_linking_launch_url, deep_linking_return_url)
 
+    def enable_proctoring(
+        self,
+        attempt_number,
```
MichaelRoytman, line 560: TODO: Make `session_data` and `resource_link` positional arguments, as they're required.

lti_consumer/lti_1p3/consumer.py
```diff
@@ -586,8 +640,42 @@ def generate_launch_request(
                 )
             }
 
-        # Call LTI Launch if Deep Linking is not
-        # set up or this isn't a Deep Link Launch
+        if self.proctoring and lti_message_hint in ['LtiStartProctoring', 'LtiEndAssessment']:
```
ashultz0, line 643: generate_launch_request should turn into just a router to _generate_deep_linking and _generate_proctoring or similar names, this function now has too many different concerns

lti_consumer/lti_1p3/consumer.py
```diff
@@ -625,6 +713,118 @@ def check_and_decode_deep_linking_token(self, token):
...
+        #       (indicating that it has not been verified by the Proctoring Tool).  Which identity claims a
+        #       Proctoring Tool will verify is subject to agreement outside of the scope of this specification but, at a
+        #       minimum, it is recommended that Proctoring Tools performing identity verification are able to verify the
+        #       given_name, family_name and name claims.
+        #       See 3.3 Transferring the Candidate Back to the Assessment Platform.
+
+        # Check Required LTI Claims
+        # -------------------------
+
+        # Check that the response message_type claim is "LtiStartAssessment".
```
ashultz0, line 738: this stuff could probably be turned into a set of calls to a generic checker. They're mostly response.get(keystring) != None and then compare to some value so you could do it as check(response, keystring, desired value, errormessage)

but I'm not sure that would really be more readable or understandable

lti_consumer/models.py
```diff
@@ -147,6 +177,14 @@ def clean(self):
             raise ValidationError({
                 "config_store": _("LTI Configuration stores on XBlock needs a block location set."),
             })
+        if self.config_store == self.CONFIG_ON_DB:
+            if self.lti_1p3_tool_public_key == "" and self.lti_1p3_tool_keyset_url == "":
```
ashultz0, line 181: presumably None would also be a problem here?

also won't this validation error will pop up for everything that was migrated?

lti_consumer/lti_1p3/consumer.py
```diff
@@ -540,18 +555,57 @@ def enable_deep_linking(
         """
         self.dl = LtiDeepLinking(deep_linking_launch_url, deep_linking_return_url)
 
+    def enable_proctoring(
```
schenedx, line 558: I am wondering if we can create a new class called `LtiProctoringConsumer` instead of enhancing the existing LtiAdvantageConsumer. This way, we can write our Proctoring logic in the new inherited class?

lti_consumer/migrations/0014_add_remaining_fields_to_lti_config.py
```diff
@@ -0,0 +1,33 @@
...
+        ),
+        migrations.AddField(
+            model_name='lticonfiguration',
+            name='lti_1p3_oidc_url',
+            field=models.CharField(blank=True, help_text='This is the OIDC login intitiation URL in the LTI 1.3 flow, which should be provided by the LTI Tool.', max_length=255),
+        ),
+        migrations.AddField(
+            model_name='lticonfiguration',
+            name='lti_1p3_tool_keyset_url',
+            field=models.CharField(blank=True, help_text="This is the Tool's JWK (JSON Web Key) keyset URL. This should be provided by the LTI Tool. One of either lti_1p3_tool_public_key or lti_1p3_tool_keyset_url must not be blank.", max_length=255),
```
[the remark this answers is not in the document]
MichaelRoytman: The `model.clean()` method is one form of model validation. It's called as part of a call to `model.full_clean()`, which calls `model.clean_fields()`, `model.clean()`, and `model.validate_unique()`. You can call `model.clean()` on its own, but it will also be invoked when calling `model.full_clean()`.

When calling `is_valid()` on a `ModelForm`, `model.full_clean()` will be called. Because the Django admin form uses ModelForms, `clean` will be called when saving an object in the Django admin, which is what we want for new objects.

As far as existing objects, if someone goes into the admin and tries to modify the existing object, then they will see an error in the Django admin.

That said, I don't believe that it's even possible to use LTI consumers with the config_store `CONFIG_ON_DB` right now, so this should not occur.

There are 19 objects in the database with the `config_store` set to `CONFIG_ON_DB`, so once we use `CONFIG_ON_DB`, we should determine what they're used for, if at all. If necessary, we can remove them or can manually update these values.

As an aside, `model.clean()` is not called when calling `save()` on a model.

lti_consumer/models.py
MichaelRoytman: For all instances of `LtiConfiguration` that have `config_store="CONFIG_ON_DB"`, yes.

Because both model fields have `blank=True` and `null=False` (by default), not setting a value will be stored as the empty string in the database, not `null`. For that reason, `None` is not a concern here.

lti_consumer/lti_1p3/consumer.py
```diff
@@ -417,6 +421,12 @@ def _validate_preflight_response(self, response):
         :param response: the preflight response to be validated
         """
         try:
+            # TODO: Per the specification, I believe we're meant to also do the following.
+            # 1. validate that the learner represented by "login_hint" is the currently authenticated user
+            #          making this request.
+            # 2. confirm that the redirect_uri matches one of the registered launch endpoints for the Tool
+            #          integration identified by the client_id, as specified during the configuration phase.
+            # TODO: Check in with Giovanni.
```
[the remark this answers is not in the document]
giovannicimolin: @MichaelRoytman 
> validate that the learner represented by "login_hint" is the currently authenticated user

On the LTI consumer, we set `login_hint` as the [block id](https://github.com/openedx/xblock-lti-consumer/blob/master/lti_consumer/lti_xblock.py#L1516) - ideally, we should keep this consistent, but I'm ok to change on both places.

> 2. confirm that the redirect_uri matches one of the registered launch endpoints for the Tool

We did that initially, but a bunch of tools complained or stopped working. When using deep linking, the content URL changes and it would be a bit complicated to verify that here if the consumer is being instanced without looking at the stored deep linking configurations.

See: https://github.com/openedx/xblock-lti-consumer/blame/b80ca5fdecae89ba3857c71d8b86aa7215b42e6b/lti_consumer/lti_1p3/consumer.py#L403

lti_consumer/lti_1p3/consumer.py
giovannicimolin: :+1: for splitting this into a separate module - proctoring doesn't have the same use cases as `LtiAdvantage` in the platform now.

lti_consumer/lti_1p3/consumer.py
```diff
@@ -540,19 +556,45 @@ def enable_deep_linking(
...
+        """
+        Set the optional user data claims, per the Proctoring Services specification.
+        """
+        # As defined in the LTI Core specification [LTI-13], the https://purl.imsglobal.org/spec/lti/claim/roles claim
+        # is required but MAY be empty.
+        # The Start Proctoring message always initiates the proctoring process (and corresponding assessment launch)
+        # regardless of any value specified for this claim, the Proctoring Tool MUST ignore this value.
+        # If the Assessment Platform provides a non-empty value it SHOULD include
+        # http://purl.imsglobal.org/vocab/lis/v2/membership#Learner.
+        # TODO: This is hacky. The roles list should either be empty or should include #Learner. #Leaner is a
```
giovannicimolin, line 584: :+1: for this. I think we can just add it here: https://github.com/openedx/xblock-lti-consumer/blob/master/lti_consumer/lti_1p3/constants.py#L29

lti_consumer/lti_1p3/consumer.py
```diff
@@ -625,6 +721,173 @@ def check_and_decode_deep_linking_token(self, token):
...
+
+        claim_key = "https://purl.imsglobal.org/spec/lti/claim/message_type"
+        self._check_token_claim(
+            proctoring_response,
+            claim_key,
+            "LtiStartAssessment",
+            f"Token's {claim_key} claim should be LtiStartAssessment."
+        )
+
+        # Check that the response version claim is "1.3.0".
```
giovannicimolin, line 769: The next few checks are generic to all messages received from the tool (namely `LtiDeepLinkingResponse`) - so maybe this should be generic?

lti_consumer/lti_1p3/consumer.py
```diff
@@ -625,6 +721,173 @@ def check_and_decode_deep_linking_token(self, token):
...
+
+        # Check that the response deployment_id claim is "1". We do not support multi-tenancy,
+        # so this should always be "1".
+        # deployment_id = proctoring_response.get("https://purl.imsglobal.org/spec/lti/claim/deployment_id")
+        # if deployment_id is None:
+        #     raise exceptions.MissingRequiredClaim("Token is missing required deployment_id claim.")
+        # if deployment_id != "1":
+        #     raise exceptions.InvalidClaimValue("Token's deployment_id claim is not correct.")
+
+        claim_key = "https://purl.imsglobal.org/spec/lti/claim/deployment_id"
```
giovannicimolin, line 792: I wouldn't hardcode this check here and now - the worst it can do with the current implementation is just fail a validation and stop the tool from working.

lti_consumer/lti_xblock.py
```diff
@@ -1121,6 +1121,10 @@ def lti_1p3_launch_callback(self, request, suffix=''):  # pylint: disable=unused
             lti_message_hint = preflight_response.get('lti_message_hint', '')
 
             # Set LTI Launch URL
+            # TODO: I believe this should be the redirect_uri claim in the preflight_response.
+            # See 5.1.1.3 Step 3: Authentication Response of the IMS Security Framework.
+            # https://www.imsglobal.org/spec/security/v1p0/
+            # TODO: Check in with Giovanni.
```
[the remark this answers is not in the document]
giovannicimolin: @MichaelRoytman Yes! Nice find.

lti_consumer/migrations/0014_add_remaining_fields_to_lti_config.py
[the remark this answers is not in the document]
giovannicimolin: >That said, I don't believe that it's even possible to use LTI consumers with the config_store CONFIG_ON_DB right now, so this should not occur.

LTI 1.3 doesn't support `CONFIG_ON_DB` now but that's going to change in the next few weeks - I and @tecoholic  (OpenCraft) are working on pluggable LTI configurations for LTI 1.3 in the current sprint.

`CONFIG_ON_DB` can **currently** be used for any kind of LTI 1.1 integration or embedded anywhere in the platform, and if I'm not mistaken it's used by the new Discussions project integrations. Take care to only delete the `LTI_1P3` versions with `CONFIG_ON_DB` to avoid potentially breaking other integrations.

lti_consumer/models.py
giovannicimolin: @MichaelRoytman Make sure this doesn't trip with LTI 1.1 configurations.
`lti_1p3_tool_public_key` and `lti_1p3_tool_keyset_url` are only valid for LTI 1.3 integrations, not 1.1 ones.

lti_consumer/models.py
```diff
@@ -316,10 +354,33 @@ def _get_lti_1p3_consumer(self):
                 consumer.enable_nrps(get_lti_nrps_context_membership_url(self.id))
 
             return consumer
+        elif self.config_store == self.CONFIG_ON_DB:
+            # TODO: I couldn't quite tell whether the Proctoring Services Specification relies on LTI Advantage or
+            #       just LTI 1.3, but I think, according to this document, it's a proposed LTI Advantage service,
+            #       so I'm using the LtiAdvantageConsumer.
+            #       https://www.imsglobal.org/activity/learning-tools-interoperability
```
giovannicimolin, line 361: @MichaelRoytman All LTI Advantage services only rely on the base LTI 1.3 spec and are completely independent between themselves. 
Come to think of it, it might be best to keep the proctoring service inside `LtiAdvantageConsumer` to avoid and extra fork here when configuring the class, what do you say?

lti_consumer/models.py
```diff
@@ -316,10 +354,33 @@ def _get_lti_1p3_consumer(self):
...
+                # we're not using multi-tenancy.
+                deployment_id="1",
+                # Model Private RSA Key
+                rsa_key=self.lti_1p3_private_key,
+                rsa_key_id=self.lti_1p3_private_key_id,
+                # Model LTI 1.3 Tool Key/Keyset URL
+                tool_key=self.lti_1p3_tool_public_key,
+                tool_keyset_url=self.lti_1p3_tool_keyset_url,
+            )
+            # I am omitting the various other LTI Advantage services that can be set up, as they're not relevant
```
giovannicimolin, line 377: Can you raise an exception if this is anything other than a proctoring configuration?

lti_consumer/models.py
MichaelRoytman: That makes sense. We either need a consumer class for each LTI service (or, at least an extra one for proctoring if we keep the LtiAdvantageConsumer as one) and then have to decide which consumer class to use depending on whether we're using proctoring or not, or we need to have all services implemented in `LtiAdvantageConsumer`, which allows us to avoid the forking but makes it bloated.

On the other hand, once the library is refactored to be able to be used outside an xBlock, users of the library may not want all LTI services, right? edx-exams will not want Assignment and Grade Services, for example. I think having separate classes might allow us to gate what features are supported for an installation.

lti_consumer/models.py
MichaelRoytman: How do you recommend doing that? Right now, that kind of information is stored on the xBlock, and there isn't an analogous place when storing configuration in the database. Is the right approach to add a `proctoring_enabled` field to the model?

lti_consumer/lti_1p3/consumer.py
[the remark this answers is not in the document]
MichaelRoytman: Okay, I'll drop a comment about `redirect_uri`. For `login_hint`, perhaps we'll deal with that when we do the refactor of the library to be less dependent on xBlocks. I don't want to mess with it right now.

lti_consumer/migrations/0014_add_remaining_fields_to_lti_config.py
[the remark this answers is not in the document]
MichaelRoytman: Got it. In that case, we should probably just leave them alone and not delete anything. I modified the condition to only raise an exception if the `version` is 1.3, not 1.1. Configurations using LTI 1.1 will be unaffected. If someone tries to change a configuration using LTI 1.3, they'll just have to supply one of these values or reach out to support.

MichaelRoytman:
I'm closing this pull request, as this is not ready to be merged into the library. The library will require a refactor to decouple xBlocks and the LTI capabilities before we make additional proctoring changes to the library. `edx-edxams` will install from this commit for the LTI for proctoring POC until such time. Please do not delete this branch.

PR #254, opened 2022-06-03
[BB-5559] Decouple LTI 1.3 from LTI Consumer XBlock functionality

tecoholic:
### Description

The goal of the PR is to decouple the LTI 1.3 functionality from the XBlock and move the functionality to the Django Plugin, while keeping backward compatibility with the XBlock Handlers. The main consideration is that, this does not break any existing integrations. So all the endpoints and XBlock handlers are made backward compatible.

#### Changes introduced by the refactor

1. The functionality of LTI 1.3 Launch Handler is moved from the XBlock to the Django plugin
2. The functionality of the Access Token endpoint is moved from XBlock to the Django plugin
3. A [new URL format](https://github.com/openedx/xblock-lti-consumer/pull/254/files#diff-80779cdd3ff9ef517c54ff7142a76970028a4eaf8b155c013e2a4a3d54b015d0R40) using the `LtiConfiguration` ID is introduced for the Access Token endpoint and is used when a LTI Consumer is configured without a `location` allowing LTI integrations to be created without the XBlock context.
5. A [new URL format](https://github.com/openedx/xblock-lti-consumer/pull/254/files#diff-80779cdd3ff9ef517c54ff7142a76970028a4eaf8b155c013e2a4a3d54b015d0R24) using the `LtiConfiguration` ID is introduced for the Keyset Endpoint and is used with the `location` of the XBlock is not available in the configuration.

### Testing instructions

1. Setup an LTI Consumer block (using test tools like [Saltire](https://saltire.lti.app/) or [IMS Global Reference Tool](https://lti-ri.imsglobal.org/))
2. Verify the LTI launches are working as expected.
3. Install the PR branch in the test system `pip install git+https://github.com/open-craft/xblock-lti-consumer.git@tecoholic/BB-5559-lti1p3-refactor`. If you are using devstack, install it in both lms and studio containers.
4. Restart the services as necessary.
5. Ensure that all the LTI Blocks are working as before.


openedx-webhooks:
Thanks for the pull request, @tecoholic! I've created [OSPR-6754](https://openedx.atlassian.net/browse/OSPR-6754) to keep track of it in JIRA, where we prioritize reviews. Please note that it may take us up to several weeks or months to complete a review and merge your PR.

Feel free to add as much of the following information to the ticket as you can:
- supporting documentation

Please let us know once your PR is ready for our review and all tests are green.

giovannicimolin:
@tecoholic 
>@giovannicimolin Once we move the Access Token logic to the Django view, do we need to keep this XBlock Handler here? I do not find any internal reference to this handler except from the Django view. As that is no longer required, we can completely get rid of this handler here right?
I wanted to confirm this with you before I started rewriting the related unittests. Kindly let me know if there is any reason to keep this.
Don't mind the redirect stub I have put in here - its just a placeholder.

Unfortunately, we can't, this is used by external tools - and there's already a lot of them configured on edX.org. 
We need to keep backwards capability here (just as a passthrough view, minimal code).

Studio should only display the new URL from now onwards.

giovannicimolin (changes requested):
@tecoholic Woah, great progress here! 

You'll also need to update the URLs on https://github.com/openedx/xblock-lti-consumer/blob/master/lti_consumer/utils.py#L56 to point to the new endpoints and update references to those.

lti_consumer/plugin/views.py
```diff
@@ -117,41 +126,200 @@ def public_keyset_endpoint(request, usage_id=None):
...
-    This is basically a passthrough function that uses the
-    OIDC response parameter `login_hint` to locate the block
-    and run the proper handler.
+    This uses the location key from the "login_hint" query parameter
+    to identify the LtiConfiguration and its consumer to generate the
+    LTI 1.3 Launch Form.
     """
+    usage_id = request.GET.get('login_hint')
+    if not usage_id:
+        return render(request, 'html/lti_1p3_launch_error.html', status=400)
```
Agrendalath, line 145: Nit: let's use `HTTP_400_BAD_REQUEST`.

lti_consumer/plugin/views.py
```diff
@@ -117,41 +126,200 @@ def public_keyset_endpoint(request, usage_id=None):
...
+
     try:
-        usage_key_str = request.GET.get('login_hint')
-        usage_key = UsageKey.from_string(usage_key_str)
+        usage_key = UsageKey.from_string(usage_id)
+    except InvalidKeyError:
+        return render(request, 'html/lti_1p3_launch_error.html', status=404)
```
Agrendalath, line 150: Nit: let's use `HTTP_404_NOT_FOUND`. For consistency, we could convert all status codes to constants within this diff.

lti_consumer/lti_xblock.py
```diff
@@ -1100,172 +1099,70 @@ def lti_launch_handler(self, request, suffix=''):  # pylint: disable=unused-argu
...
         References:
             Sucess: https://tools.ietf.org/html/rfc6749#section-4.4.3
             Failure: https://tools.ietf.org/html/rfc6749#section-5.2
         """
-        if self.lti_version != "lti_1p3":
-            return Response(status=404)
-        if request.method != "POST":
-            return Response(status=405)
+        # Update the DB with the LTI 1.3 config and fetch the config
+        conf = self._save_lti_1p3_config_to_db()
```
Agrendalath, line 1162: All other invocations of `_save_lti_1p3_config_to_db` are gated with the `self.lti_version == "lti_1p3"` check. Don't we need to check it here too?
Or maybe it would be better to check it directly in `_save_lti_1p3_config_to_db`?

lti_consumer/lti_xblock.py
```diff
@@ -1100,172 +1099,70 @@ def lti_launch_handler(self, request, suffix=''):  # pylint: disable=unused-argu
...
+
+    def _save_lti_1p3_config_to_db(self):
+        """
+        Save LTI 1.3 config to the database.
+        """
+        from .api import _get_lti_config  # pylint: disable=import-outside-toplevel
+        conf = _get_lti_config(block=self)
+        # Transfer the config to DB so that the django view can initialize the LTI1P3Consumer
+        # without having to load the XBlock
+        conf.lti_config = self._lti_1p3_config_as_dict()
```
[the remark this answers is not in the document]
tecoholic: This was on oversight on my part. This will be replaced with `config.lti_config.update(self._lti_1p3_config_as_dict())` to preserve existing information.

lti_consumer/plugin/compat.py
```diff
@@ -227,3 +228,21 @@ def clean_course_id(model_form: ModelForm) -> CourseKey:
...
     return lms_clean_course_id(model_form)
+
+
+def get_user_role(user: User, course_key: CourseKey) -> str:
+    """
+    Import the get_user_role from LMS and return the value.
+    """
+    # pylint: disable=import-outside-toplevel
+    from lms.djangoapps.courseware.access import get_user_role
+    return get_user_role(user, course_key)
```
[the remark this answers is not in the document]
tecoholic: Yes. It matches the `user` service provided by the LMS Module renderer [here](https://github.com/openedx/edx-platform/blob/master/lms/djangoapps/courseware/module_render.py#L526)

tecoholic:
**NOTE:** The `lti_config_id` in the URLs are [supposed to be UUID](https://github.com/openedx/xblock-lti-consumer/blob/master/lti_consumer/models.py#L94) but it got [mis-implemented here, I think](https://github.com/openedx/xblock-lti-consumer/commit/ea85ee14ecd3f36477182b3f868c02aba4dd9125#diff-ca60187b5387933d55e3e6828611d72dd18b76c30aeda38ff1c955c676ef7dd8R95). I guess this wasn't caught early on because the regex `([-\w]+)` applies to both UUID and integer ids. Maybe this was intentional. Anyway, this means, the new URLs can use UUIDs and prevent primary key from leaking as originally intended.

lti_consumer/lti_xblock.py
```diff
@@ -1100,172 +1099,70 @@ def lti_launch_handler(self, request, suffix=''):  # pylint: disable=unused-argu
...
+        request.user = self.runtime.service(self, 'user').get_current_user()
+        return launch_gate_endpoint(request, suffix)
+
+    def _save_lti_1p3_config_to_db(self):
+        """
+        Save LTI 1.3 config to the database.
+        """
+        from .api import _get_lti_config  # pylint: disable=import-outside-toplevel
+        conf = _get_lti_config(block=self)
+        # Transfer the config to DB so that the django view can initialize the LTI1P3Consumer
```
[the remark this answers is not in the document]
tecoholic: I have improved the [DB update logic](https://github.com/openedx/xblock-lti-consumer/pull/254/commits/c7a85450202c265ea41d12b1574444a86dc658d7#diff-6e408862179502a0ac809fab914e028b18311a17a3b0874469bd6416bc4a449fR114). It is done in the following scenarios:

1. Every time the [XBlock is updated](https://github.com/openedx/xblock-lti-consumer/pull/254/commits/c7a85450202c265ea41d12b1574444a86dc658d7#diff-cc618c2e63b8c2d293d245f9799381d9321e40fe2d05e9a4b2f80bb91995157dR1052) from Studio using the Edit button
2. In order to maintain backward compatibility with existing integrations, the `get_or_update_lti_config` will be called in every request passing through the XBlock Handlers (preflight_launch_handler or access_token_handler), but the update will only in the first call.

If someone manually changes values in `lti_config` attribute via the admin, and it has corresponding values in the XBlock, then the XBlock values will be writer OVER the MANUALly edited values. Is this something of a concern? I am assuming the maintainers are more likely to edit the XBlock in Studio to change LTI 1.3 configurations, than to tinker with values in the admin :thinking:

lti_consumer/lti_xblock.py
```diff
@@ -1100,172 +1099,70 @@ def lti_launch_handler(self, request, suffix=''):  # pylint: disable=unused-argu
...
-            template = loader.render_django_template('/templates/html/lti_1p3_permission_error.html', context)
-            return Response(template, status=403, content_type='text/html')
+        # Backward compatibility - copy the config to the DB
+        self._save_lti_1p3_config_to_db()
+
+        from .plugin.views import launch_gate_endpoint  # pylint: disable=import-outside-toplevel
+        request.user = self.runtime.service(self, 'user').get_current_user()
+        return launch_gate_endpoint(request, suffix)
+
+    def _save_lti_1p3_config_to_db(self):
```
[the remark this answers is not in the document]
tecoholic: Yes. That made more sense. I have now moved it to the `api.py` [here](https://github.com/openedx/xblock-lti-consumer/pull/254/commits/c7a85450202c265ea41d12b1574444a86dc658d7#diff-6e408862179502a0ac809fab914e028b18311a17a3b0874469bd6416bc4a449fR114)

lti_consumer/lti_xblock.py
```diff
@@ -1100,172 +1099,70 @@ def lti_launch_handler(self, request, suffix=''):  # pylint: disable=unused-argu
...
-                str(self.location),  # pylint: disable=no-member
-                self.external_user_id,
-                exc,
-            )
-            template = loader.render_django_template('/templates/html/lti_1p3_permission_error.html', context)
-            return Response(template, status=403, content_type='text/html')
+        # Backward compatibility - copy the config to the DB
+        self._save_lti_1p3_config_to_db()
+
+        from .plugin.views import launch_gate_endpoint  # pylint: disable=import-outside-toplevel
```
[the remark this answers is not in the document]
tecoholic: Same as above, Django views are available only after LMS or Studio has finished initializing.

lti_consumer/lti_xblock.py
```diff
@@ -1100,172 +1099,70 @@ def lti_launch_handler(self, request, suffix=''):  # pylint: disable=unused-argu
...
-            return Response(
-                json_body={"error": "invalid_client"},
-                status=400,
-            )
-        except UnsupportedGrantType:
-            return Response(
-                json_body={"error": "unsupported_grant_type"},
-                status=400,
-            )
+        from .plugin.views import access_token_endpoint  # pylint: disable=import-outside-toplevel
```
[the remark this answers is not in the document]
tecoholic: Same reasons as the above one.

lti_consumer/plugin/views.py
```diff
@@ -117,41 +126,200 @@ def public_keyset_endpoint(request, usage_id=None):
         response = JsonResponse(lti_config.lti_1p3_public_jwk)
         response['Content-Disposition'] = 'attachment; filename=keyset.json'
         return response
-    except (LtiError, InvalidKeyError, ObjectDoesNotExist) as exc:
+    except (LtiError, InvalidKeyError, ObjectDoesNotExist, ValueError) as exc:
```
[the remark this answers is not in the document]
tecoholic: This `ValueError` has been removed after the urls were modified to use UUIDs instead of primary keys.

lti_consumer/plugin/views.py
```diff
@@ -117,41 +126,200 @@ def public_keyset_endpoint(request, usage_id=None):
...
         log.info("Error while retrieving keyset for usage_id %r: %s", usage_id, exc)
```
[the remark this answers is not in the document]
tecoholic: Done. Added it to the log.

lti_consumer/plugin/views.py
tecoholic: Updated all instances where integers were used.

lti_consumer/lti_xblock.py
tecoholic: Makes sense. This has been changed to make the check happen in the save function.

tecoholic:
@Zacharis278 Hi! Looks like #260 is a very similar to this. :eyes: You are right, @giovannicimolin is going to be unavailable for the month. @Agrendalath is handling the reviews in the meantime. Maybe he can help.

lti_consumer/api.py
```diff
@@ -83,6 +83,63 @@ def _get_lti_config(config_id=None, block=None):
...
+        block_config = _lti_1p3_config_as_dict(block)
+        config_changed = any(
+            conf.lti_config.get(k, '') != block_config[k] for k in block_config
+        )
+
+        # Optimization - Update the DB only if values are different or config is still
+        # set to CONFIG_ON_XBLOCK
+        if config_changed or conf.config_store == conf.CONFIG_ON_XBLOCK:
+            conf.lti_config.update(block_config)
+            conf.config_store = conf.CONFIG_ON_DB
```
Agrendalath, line 138: Do we also need to set `config_type` to `'external'`? I see that `_get_lti_config` is checking this, not `_get_lti_config`.

lti_consumer/lti_xblock.py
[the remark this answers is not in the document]
Agrendalath: @tecoholic,

> If someone manually changes values in lti_config attribute via the admin, and it has corresponding values in the XBlock, then the XBlock values will be writer OVER the MANUALly edited values. Is this something of a concern? I am assuming the maintainers are more likely to edit the XBlock in Studio to change LTI 1.3 configurations, than to tinker with values in the admin thinking

Good point. Doesn't this make DB configuration unusable? If somebody changes this through the admin interface, then even updating this XBlock's Display Name is going to overwrite all changes, no? What if we added something like this, e.g. as a `post_save` receiver?
```python
@receiver(post_save, sender=LtiConfiguration)
def update_xblock_config(sender, instance, **kwargs):
    block = instance.block
    # Update configs.
    block.save()
```

lti_consumer/plugin/views.py
```diff
@@ -301,6 +316,33 @@ def access_token_endpoint(request, lti_config_id):
...
 
+@csrf_exempt
+@require_http_methods(["POST"])
+def access_token_endpoint_via_location(request, usage_id=None):
+    """
+    Access token endpoint that provides backwards compatibility to the LTI tools
+    that were configured using the the older version of the URL with usage_id in it
+    instead of the config id.
+
+    We maintain this extra view instead of fetching LTI Config using the usage_id is
```
Agrendalath, line 327:
```suggestion
    We maintain this extra view instead of fetching LTI Config using the usage_id
```
Nit.

lti_consumer/tests/unit/test_api.py
```diff
@@ -291,12 +291,12 @@ def test_retrieve_with_block(self):
...
                 'client_id': lti_config.lti_1p3_client_id,
                 'keyset_url': 'https://example.com/api/lti_consumer/v1/public_keysets/{}'.format(
-                    lti_config.location
+                    lti_config.config_id
                 ),
                 'deployment_id': '1',
                 'oidc_callback': 'https://example.com/api/lti_consumer/v1/launch/',
                 'token_url': 'https://example.com/api/lti_consumer/v1/token/{}'.format(
-                    lti_config.location
+                    lti_config.config_id
```
Agrendalath, line 299: Shouldn't we check both options (for backward compatibility)?

tecoholic: Good catch here. Yes. We will need to. In fact, I will have to make sure these urls for the existing blocks aren't persisted (if it is not happening already). Otherwise we run into the situation where XBlock will show an URL with the `config_id` in the UI while it might be configured with `location` in the tool. Functionally it wouldn't break anything, but will be a confusing situation if someone tries to change/debug the integration for some reason.

lti_consumer/lti_xblock.py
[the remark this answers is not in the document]
tecoholic: @Agrendalath I have implemented the hook in [3976c2b](https://github.com/openedx/xblock-lti-consumer/pull/254/commits/3976c2b1102878009a9b399b96ff89ad2954b40f) using the `pre_save` signal instead of the `post_save`, so that the side-effects can easily be managed.

Initially I tried with `post_save` and then patched all the test cases where this was causing a side-effect. But it was still causing a lot of test-case failures. The only way I could find to totally isolate the situation was to determine if the XBlock really needed updating, by comparing the previous value with the current value. But it's not possible from the `post_save` as the DB is already updated.

So I have used `pre_save` which allows us to fetch the XBlock and update it only when specific fields are changed. Right now I have limited it to the values stored in `lti_config` attribute. But after #260 is merged, I think this will change.

lti_consumer/signals.py
```diff
@@ -62,3 +62,42 @@ def publish_grade_on_score_update(sender, instance, **kwargs):  # pylint: disabl
...
+    if not instance.id:
+        # New entry is being created. Do Nothing
+        return
+
+    # We are concerned only with updating the LTI 1.3 values stored in lti_config (for now)
+    # TODO: Should we expand this criteria to include other attributes like version?
+    #
+    previous = LtiConfiguration.objects.get(pk=instance.id)
+    if previous.lti_config != instance.lti_config:
+        block = instance.block
```
giovannicimolin, line 99: @tecoholic What if this block is configured in the database only, and `instance.block` is `None`? This will fail saving changes. 
This case is specially important given we want to make LTI 1.3 launches independent from blocks - and we won't have `block` defined for all `LtiConfiguration` instances.

tecoholic: @giovannicimolin This Signal is for a way to update the XBlock's settings when the change occurs in the DB. If the block is `None` then we don't have the scenario where we need to keep the XBlock in sync. So performing no action at that point, I think is the valid action.

lti_consumer/api.py
tecoholic: Ah. I have not considered the external case. For now we can simply return the config without any changes in the `get_or_update_lti_config` (Line 91). Adding

```py
    if block.config_type == "external":
        return conf
```

tecoholic:
@Agrendalath @giovannicimolin I have rebased the PR after the merge of #260 and refactored the PR in response to the changes introduced by it. I need some clarity regarding the design decision to make to bring this PR to a close.

In order to maintain complete independence of the LTI Views from XBlock, I introduced the mechanism to _migrate the changes made in the XBlock (via Studio) to the Django Model_. This allows the Django view to respond to requests without relying on XBlock settings. Backward compatibility is also provided both in XBlock methods and in Django URLs. So old XBlocks will continue to function as it is without issues.

However, 2 things happened:
1. @Agrendalath [noted](https://github.com/openedx/xblock-lti-consumer/pull/254#discussion_r917853759) that we might need to verify that the URLs generated in `get_lti_1p3_launch_info` will have to be checked for both old format (XBlock usage id) and new format (LtiConfiguration id) 
    - This made me realize, that even though the LTI Tool integrations might function properly, anyone opening the Studio block to debug will encounter only the new URLs using LtiConfiguration ID as `get_lti_1p3_launch_info` generates these URLs on the fly. So there is no way for the migrated XBlocks to show the old URLs if someone opens a block after an upgrade.
2. PR 260 made storing the configuration in Django Model explicit by selecting "Database" (enabled via WaffleFlag).
    - I had implemented the config migration with the goal of moving _both future and existing_ LTI Blocks completely to the Django Model. That allows Django Views to generate responses without having to load the XBlock for every request and also doesn't bother the users with choosing between storing config on XBlock and Database. Functionally the user has no benefits either way.

What's the right next step? To account for the shortcoming flagged in point **1**, it looks like it is better to follow the approach set by PR 260 and keep all the existing XBlock configurations in XBlock and use the CONFIG_ON_DB option only for future XBlocks.

natabene:
Just checking what's going on here. How can we move this along?

tecoholic:
@giovannicimolin I am waiting on your review on this.

cc: @natabene

Zacharis278:
@tecoholic / @giovannicimolin does this just need approvals at this point? If we're good to go I can do some testing on my end and give a final 👍 as well.

giovannicimolin:
@Zacharis278 No, this still needs a full review and testing round. I'm planning to do it tomorrow before my EOD.  
Sorry for the delay here.

giovannicimolin:
@tecoholic I didn't get your last ping, sorry for the delay here! :disappointed:

lti_consumer/api.py
```diff
@@ -88,6 +88,47 @@ def _get_lti_config(config_id=None, block=None):
...
+
+        # Optimization - Update the DB only if values are different or config is still
+        # set to CONFIG_ON_XBLOCK
+        if config_changed or conf.config_store == conf.CONFIG_ON_XBLOCK:
+            for key in block.config_sync_params:
+                if key == "lti_1p3_enable_nrps":
+                    conf.lti_advantage_enable_nrps = block.lti_1p3_enable_nrps
+                else:
+                    setattr(conf, key, getattr(block, key))
+            conf.config_store = conf.CONFIG_ON_DB
```
Zacharis278, line 127: is the intention that the values will always sync but we won't actually change the config type until the waffle flag is rolled out? If so it's probably worth commenting how this is impacted by the 'database_config_enabled' feature flag in LtiConfiguration.clean(). It took me a bit to figure out how those two things were connected

Zacharis278, line 127: @tecoholic had some time today so I dug into this a bit and noticed some odd behavior here. In studio this value gets saved to the database as CONFIG_ON_DB some other part of the code seems to simultaneous be saving the xblock value which is still CONFIG_ON_XBLOCK and overriding this.

lti_consumer/api.py
```diff
@@ -88,6 +88,47 @@ def _get_lti_config(config_id=None, block=None):
...
+        the updated LtiConfiguration object
+    """
+    conf = _get_lti_config(block=block)
+
+    if block.config_type == "external":
+        return conf
+
+    if block.lti_version == "lti_1p3":
+        # Transfer the config to DB so that the django view can initialize the
+        # LTI1P3Consumer without having to load the XBlock
```
giovannicimolin, line 109: @tecoholic What's the reasoning behind ditching `CONFIG_ON_XBLOCK` in favor of `CONFIG_ON_DB` and copying over all the configuration? As far as I understand, this is not really needed to achieve the reusability we're looking for and adds extra complexities to syncing data between the block and the model.

@tecoholic @tecoholic @MichaelRoytman  Was there any discussion on the previous PRs about this?

giovannicimolin (changes requested):
@tecoholic It's working as expected on the devstack, but I need a few clarifications on the changes (plus I left some comments).

lti_consumer/api.py
giovannicimolin: @Zacharis278 I think this the exact scenario that I've described on this comment 5 minutes ago:

https://github.com/openedx/xblock-lti-consumer/pull/254/files#r936016036

lti_consumer/signals.py
```diff
@@ -62,3 +62,51 @@ def publish_grade_on_score_update(sender, instance, **kwargs):  # pylint: disabl
                 exc,
             )
             raise exc
+
+
+@receiver(pre_save, sender=LtiConfiguration, dispatch_uid='update_xblock_lti_configuration')
+def update_xblock_lti_configuration(sender, instance, **kwargs):  # pylint: disable=unused-argument
```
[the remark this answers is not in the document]
giovannicimolin: @MichaelRoytman ^

lti_consumer/api.py
giovannicimolin: CCing @Zacharis278 on this discussion too ^

Zacharis278: Something that was on my mind while looking for this was if we're moving from configuration in modulestore to the database it does seem like something that should have its own ADR to explain the reasoning/benefit.

Maybe it would be a good idea to split that into its own feature? What if we just synced the parameters in memory here and didn't save back to the DB yet?

tecoholic: @giovannicimolin @Zacharis278 Thank you for the comments. I have been thinking about this in my head for about a 2 weeks now. I think I might have over-engineered the "LTI Launch without XBlock" here. It happened when I was testing backward compatibility for the XBlock's handlers. The flow ends up something like this:

1. The XBlocks's handler is called
2. The XBlock handler forwards the request to the Django View 
3. The Django view loads the config from XBlock, thereby creating another instance of the XBlock withing it's context

This seemed too convoluted at that time and I couldn't reason out the context when Unit tests failed. So syncing the data to the DB, seemed like a clearer path -  *The Django View always loads the config from DB* - this seemed like a much simpler way to reason about.

But after the changes introduced in #260, it has become somewhat of a code smell. The implicit migration of data to the DB from XBlock is not aligned with the explicit WaffleFlag approach. I also have come to realize that this has just shifted the complexity & interdependency between XBlock & Django View to XBlock & Django Model. Which in itself might not be bad, but definitely seems like it needs some clarity along the lines of the @Zacharis278's questions.

I will drop this from the PR and let you know.

lti_consumer/signals.py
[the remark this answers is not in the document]
tecoholic: @giovannicimolin This scenario did arise in multiple unit-tests and I found the right conditions to keep this in check and get the tests to pass. But there might be a scenario that only happens in the real world when multiple requests are being served from multiple app-servers that could probably trigger such an issue. 

With that said, I think [this was over-engineered](https://github.com/openedx/xblock-lti-consumer/pull/254#discussion_r937315605) and would like to drop it from the PR. I will ping again once its done.

lti_consumer/lti_xblock.py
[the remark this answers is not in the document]
tecoholic: This part is now removed.

lti_consumer/api.py
tecoholic: This syncing is now removed.

lti_consumer/plugin/views.py
```diff
@@ -118,53 +128,204 @@ def public_keyset_endpoint(request, usage_id=None):
...
     Gate endpoint that triggers LTI launch endpoint XBlock handler
 
-    This is basically a passthrough function that uses the
-    OIDC response parameter `login_hint` to locate the block
-    and run the proper handler.
+    This uses the location key from the "login_hint" query parameter
+    to identify the LtiConfiguration and its consumer to generate the
+    LTI 1.3 Launch Form.
     """
+    usage_id = request.GET.get('login_hint')
```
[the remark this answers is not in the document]
tecoholic: @giovannicimolin I tried implementing using either the location OR the config uuid. But ran into issues down the line as the [launch context information](https://github.com/openedx/xblock-lti-consumer/blob/master/lti_consumer/lti_xblock.py#L1191-L1216) is all dependent on the course, which is being obtained from the UsageKey. So, until we figure out how to establish the context (user_id, user_role, context_title, content_types...etc.,) using the `config_id` value in `login_hint` might not be possible.

tecoholic:
@giovannicimolin I have removed the overengineered parts, incorporated your feedback about the logging, rebased the code to master. Kindly give this another go.

lti_consumer/plugin/views.py
```diff
@@ -118,53 +128,215 @@ def public_keyset_endpoint(request, usage_id=None):
...
+    try:
+        lti_config = LtiConfiguration.objects.get(
+            location=usage_key
+        )
+    except LtiConfiguration.DoesNotExist as exc:
+        log.error("Invalid usage_id '%s' for LTI 1.3 Launch callback", usage_id)
         raise Http404 from exc
 
+    if lti_config.version != LtiConfiguration.LTI_1P3:
+        log.error("The LTI Version of configuraiont %s is not LTI 1.3", lti_config)
```
MichaelRoytman, line 175: Nit.
```suggestion
        log.error("The LTI Version of configuration %s is not LTI 1.3", lti_config)
```

MichaelRoytman (commented):
I'll be building off this, so I'm reviewing to better understand your work. I left a few small thoughts.

lti_consumer/lti_xblock.py
```diff
@@ -710,6 +700,14 @@ def editable_fields(self):
         if not is_external_config_filter_enabled:
             noneditable_fields.append('external_config')
 
+        if self.lti_version == 'lti_1p3' and is_database_config_enabled and self.config_type == 'database':
```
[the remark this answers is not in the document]
tecoholic: @MichaelRoytman I believe this would be leftover code from #260. I am not sure why rebasing hasn't updated it. Maybe a conflict line. Let me update it to the latest code.

lti_consumer/lti_xblock.py
```diff
@@ -1169,186 +1167,51 @@ def lti_launch_handler(self, request, suffix=''):  # pylint: disable=unused-argu
...
-        try:
-            # Pass user data
-            lti_consumer.set_user_data(
-                user_id=self.external_user_id,
-                # Pass django user role to library
-                role=user_role
-            )
+        # Asserting that the consumer can be created. This makes sure that the LtiConfiguration
+        # object exists before calling the Django View
+        assert self._get_lti_consumer()
```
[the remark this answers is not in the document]
tecoholic: The [original code](https://github.com/openedx/xblock-lti-consumer/blob/master/lti_consumer/lti_xblock.py#L1188) doesn't have any error handling around this line. So I assume it is safe to call. We can add one here if it is required.

lti_consumer/plugin/views.py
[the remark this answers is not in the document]
giovannicimolin: @tecoholic Good point! 

This means we'll need to modify either the `login_hint` or the `lti_message_hint` to pass the id along with the appropriate context. Otherwise, we won't be able to make LTI 1.3 configurations really independent of blocks.

Let me think about a solution today - I'll try to comment back before my EOD.

giovannicimolin (changes requested):
Launches are working! @tecoholic Nice job! 

I left a few comments in the PR though, we can address the remaining issues next sprint with smaller PRs.

Zacharis278:
Hey guys, wanted to check in on where we stand with this PR. We've got some work lined up that would require doing a 1.3 launch without an xblock context so this would be great to have in master since there's some big changes here, even if we haven't totally solved the context/login_hint solution in `launch_gate_endpoint()`.

More than happy to collaborate on how to solve that in a future PR since it's key to our needs as well!

lti_consumer/plugin/views.py
```diff
@@ -102,9 +111,10 @@ def public_keyset_endpoint(request, usage_id=None):
     and run the proper handler.
     """
     try:
-        lti_config = LtiConfiguration.objects.get(
-            location=UsageKey.from_string(usage_id)
-        )
+        if usage_id:
+            lti_config = LtiConfiguration.objects.get(location=UsageKey.from_string(usage_id))
+        elif lti_config_id:
+            lti_config = LtiConfiguration.objects.get(config_id=lti_config_id)
```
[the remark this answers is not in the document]
tecoholic: But wouldn't the Django's URL resolver take care this path? The two paths that lead to this function are

```
 'lti_consumer/v1/public_keysets/<uuid:lti_config_id>'
 f'lti_consumer/v1/public_keysets/{settings.USAGE_ID_PATTERN}$'
```

lti_consumer/plugin/views.py
```diff
@@ -118,53 +128,215 @@ def public_keyset_endpoint(request, usage_id=None):
...
     try:
-        usage_key = UsageKey.from_string(usage_id)
+        if lti_config_id:
+            lti_config = LtiConfiguration.objects.get(config_id=lti_config_id)
+        else:
+            usage_key = UsageKey.from_string(usage_id)
+            lti_config = LtiConfiguration.objects.get(location=usage_key)
+    except LtiConfiguration.DoesNotExist as exc:
+        log.warning("Error getting the LTI configuration with id %r: %s", lti_config_id, exc)
+        raise Http404 from exc
```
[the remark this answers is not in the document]
tecoholic: Same question as above? The Djanog's URL resolver would reject calls without either `lti_config_id` or `usage_id`.

lti_consumer/lti_xblock.py
```diff
@@ -1169,186 +1167,51 @@ def lti_launch_handler(self, request, suffix=''):  # pylint: disable=unused-argu
         return Response(template, content_type='text/html')
 
     @XBlock.handler
-    def lti_1p3_launch_callback(self, request, suffix=''):  # pylint: disable=unused-argument
```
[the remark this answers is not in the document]
tecoholic: @giovannicimolin I have removed this handler and moved all the relevant tests to the Django View in c71a68a.

I have also removed the block you mentioned in [this comment](https://github.com/openedx/xblock-lti-consumer/pull/254#discussion_r940613322)

tecoholic:
@Zacharis278 I have made the changes to address the comments from @giovannicimolin. I think it is close to being merged.

giovannicimolin:
@Zacharis278 This is in the final stages of review. 
I've started an issue so we can all discuss the launch context problem and getting re-usability working.

giovannicimolin (approved):
:+1:

- [x] I tested this:
I set up the following tools and checked that they still work as expected:
- IMS Reference implementation (https://lti-ri.imsglobal.org/)
- H5P.com (https://opencraft.h5p.com/)
- Locally, with IMS's PHP tool reference implementation.
1. I checked that launches are working as expected :heavy_check_mark: 
2. I checked that Grades are being sent back to the LMS :heavy_check_mark: 
3. I checked that Deep linking configuration is still working :heavy_check_mark: 

- [x] I read through the code
- [ ] Includes documentation

@tecoholic Good to go, I've checked that everything keeps working with this new rework.

**Before this is merged, can you add a version bump?**
We should go to `__version__ = '4.4.0'` with this change.

tecoholic:
@giovannicimolin Great. Thank you for testing it so thoroughly. I have updated bumped the version to 4.4.0

Zacharis278 (approved):
Also run through a basic reference tool launch with this. Looks great!

openedx-webhooks:
@tecoholic 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/xblock-lti-consumer&entry.867055334=https://github.com/openedx/xblock-lti-consumer/pull/254&entry.1484655318=https://github.com/tecoholic&entry.752974735=2022-06-03+05:51&entry.1917517419=2022-08-17+18:20&entry.2133058324=Yes) so we can improve your experience in the future.

PR #259, opened 2022-06-23
feat: add event tracking to lti launch

Zacharis278:
Track lti launch request events with the LMS event tracking library. This should fail silently if used outside the LMS.

### [MST-1506](https://2u-internal.atlassian.net/browse/MST-1506?atlOrigin=eyJpIjoiOGZmOThmYzlhMjJjNDNiMmIzMzFlMjAzZTU5ZDc4NjciLCJwIjoiaiJ9)

### DIFF
--- a/lti_consumer/lti_1p1/contrib/django.py
+++ b/lti_consumer/lti_1p1/contrib/django.py
@@ -5,7 +5,7 @@ This module provides functionality for rendering an LTI embed without an XBlock.
 # See comment in docstring for explanation of the usage of ResourceLoader
 from xblockutils.resources import ResourceLoader
 
-from lti_consumer.utils import track_event
+from lti_consumer.track import track_event
 
 from ..consumer import LtiConsumer1p1
 
--- a/lti_consumer/lti_1p1/contrib/django.py
+++ b/lti_consumer/lti_1p1/contrib/django.py
@@ -5,6 +5,8 @@ This module provides functionality for rendering an LTI embed without an XBlock.
 # See comment in docstring for explanation of the usage of ResourceLoader
 from xblockutils.resources import ResourceLoader
 
+from lti_consumer.utils import track_event
+
 from ..consumer import LtiConsumer1p1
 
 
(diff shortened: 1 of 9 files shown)


Zacharis278:
@openedx/edx-infinity this also adds tracking to non-xblock launches for zoom and discussions. Any comments or concerns?

lti_consumer/lti_1p1/contrib/django.py
```diff
@@ -123,6 +125,14 @@ def lti_embed(
     }
     context.update({'lti_parameters': lti_parameters})
 
+    # emit tracking event
+    event = {
+        'lti_version': lti_parameters.get('lti_version'),
+        'user_roles': lti_parameters.get('roles'),
+        'launch_url': lti_consumer.lti_launch_url,
+    }
+    track_event('embed.launch_request', event)
```
ashultz0, line 134: right now you could lti.track_even with lots of different data - should track_event take a set of defined fields instead of just a freeform event?

lti_consumer/utils.py
```diff
@@ -164,3 +164,13 @@ def external_config_filter_enabled(course_key):
...
+
+
+def track_event(event_name, data):
+    """
+    Track analytics event. Fail silently if no tracking module present
+    """
+    tracker = get_event_tracker()
+    if tracker:
+        event_name = '.'.join(['edx', 'lti', event_name])
+        tracker.emit(event_name, data)
```
ashultz0, line 176: this feels like it deserves its own track.py or event.py file even if that's small rather than just bundling into utils

lti_consumer/lti_1p1/contrib/django.py
Zacharis278: I'm thinking freeform since some 'defined fields' such as course id / context get tacked on deeper in the library. Each event itself may be totally different in what we want to store.

asadazam93:
@Zacharis278 I have no concerns about this change. In fact the data might be useful for us.

PR #260, opened 2022-06-29
feat: Add core LTI 1.3  and LTI Advantage configuration to LTIConfiguration model and support XBlock using the database

michaelroytman:
This pull request adds additional core LTI 1.3 and LTI Advantage variables to the LTIConfiguration model. The additional core LTI 1.3 variables are `lti_1p3_oidc_url`, `lti_1p3_launch_url`, `lti_1p3_tool_public_key`, and `lti_1p3_tool_keyset_url`. The additional LTI Advantage variables are `lti_advantage_enable_nrps`, `lti_advantage_deep_linking_enabled`, `lti_advantage_deep_linking_launch_url`, and `lti_advantage_ags_mode`.

This commit also adds a configuration type to the `LtiConsumerXBlock` to support the storage of these LTI variables on the LTIConfiguration model (i.e. the database) instead of the xBlock itself.

Changes that allow the use of this configuration option are behind the `lti_consumer.enable_database_config` CourseWaffleFlag.

JIRA: [MST-1467](https://2u-internal.atlassian.net/browse/MST-1467)


lti_consumer/lti_xblock.py
```diff
@@ -1120,6 +1153,10 @@ def lti_1p3_launch_callback(self, request, suffix=''):  # pylint: disable=unused
         Returns:
             webob.response: HTML LTI launch form or error page if misconfigured
         """
+        # TODO: I need to use the LTIConfiguration model for the version value here, but the consumer model
```
MichaelRoytman, line 1156: Any ideas for this?

lti_consumer/models.py
```diff
@@ -150,6 +150,34 @@ class LtiConfiguration(models.Model):
...
+                  'One of either lti_1p3_tool_public_key or lti_1p3_tool_keyset_url must not be blank.'
+    )
+
+    lti_1p3_tool_keyset_url = models.CharField(
+        max_length=255,
+        blank=True,
+        help_text='This is the Tool\'s JWK (JSON Web Key) Keyset (JWKS) URL. This should be provided by the LTI Tool.'
+                  ' One of either lti_1p3_tool_public_key or lti_1p3_tool_keyset_url must not be blank.'
+    )
+
```
MichaelRoytman, line 180: Note: I did not add the `lti_1p3_tool_key_mode`, which is a field on the xBlock, to the model. Its only use is to toggle which field is displayed in the Studio view and to validate that `lti_1p3_tool_keyset_url` and `lti_1p3_tool_public_key` have valid values  and it doesn't seem to be used otherwise. I didn't feel it was necessary to include this in the model, since I can raise an exception in the `clean` method.

lti_consumer/api.py
```diff
@@ -67,12 +82,25 @@ def _get_lti_config(config_id=None, block=None):
             lti_config = _get_or_create_local_lti_config(
                 config.get("version"),
                 block.location,
-                block.external_config
+                external_id=block.external_config,
+                config_type="external"
+            )
+        elif block.config_type == "database":
+            lti_config = _get_or_create_local_lti_config(
+                # TODO: I'm not satisfied with this. I want the version to come from the LTIConfiguration model, not
```
MichaelRoytman, line 90: Any thoughts?

lti_consumer/api.py
```diff
@@ -34,15 +38,26 @@ def _get_or_create_local_lti_config(lti_version, block_location, external_id=Non
...
-        lti_config.external_id = external_id
+    config_store = None
+    if config_type == "external" or external_id is not None:
+        config_store = LtiConfiguration.CONFIG_EXTERNAL
+    elif config_type == "database":
+        config_store = LtiConfiguration.CONFIG_ON_DB
     else:
-        lti_config.config_store = LtiConfiguration.CONFIG_ON_XBLOCK
-        lti_config.external_id = None
+        config_store = LtiConfiguration.CONFIG_ON_XBLOCK
```
ashultz0, line 47: config_type new is the default but it ends up with config store xblock - could we just call config_type as config_store instead and the default is "xblock" not "new"?

it would also be nice to have this be else if type = "new" (or "xblock" if you like the above) and the final default else be else: throw exception "wtf is this that is not a type"

lti_consumer/api.py
```diff
@@ -34,15 +38,26 @@ def _get_or_create_local_lti_config(lti_version, block_location, external_id=Non
...
+
+    lti_config.config_store = config_store
+    lti_config.external_id = external_id
+
+    # If the LtiConfiguration version on the model does not match what is on the configuration used by the integration
+    # (i.e xBlock or external configuration), then update the model to match. This may occur if the LTI
+    # configuration is stored on the xBlock (i.e. "new"), and the xBlock field has changed since the LtiConfiguration
+    # was last accessed.
+    # If the LTI configuration is stored on the LtiConfiguration model (i.e. "database"), then the lti_version function
+    # parameter can be ignored. There's no need to update the version on the model, because the model is the source of
```
ashultz0, line 57: what does it mean if the database says version 1.3 but the incoming change says version 1.1? Doesn't that mean the request is possibly incoherent?

lti_consumer/api.py
ashultz0: this is a version of my question above so you're already thinking of it

_get_or_create doesn't do that much work aside from choosing store type and you already switch on db/block/external here. Would it make sense to have _get_or_create_from_db and _get_or_create_external and _get_or_create_on_xblock which take different parameters? Each of those would be tiny.

lti_consumer/lti_xblock.py
```diff
@@ -131,6 +132,26 @@ def parse_handler_suffix(suffix):
     raise LtiError(msg)
 
 
+def config_type_values_provider(block):
+    """
+    Return a list of possible values for the config_type xBlock field.
```
ashultz0, line 137: maybe call this allowed_config_types(block) or possible_config_types(block) that obviously returns a set of types rather than using the jargony "provider" which might return some weird model?

lti_consumer/lti_xblock.py
```diff
@@ -922,6 +958,28 @@ def is_past_due(self):
...
 
+    def _get_lti_advantage_deep_linking_enabled(self, consumer):
+        """
+        Return whether the LTI 1.3 Advantage Deep Linking is enabled. Use the LTI consumer to determine whether deep
+        linking is enabled instead of directly accessing the xBlock fields, the LTI Configuration model, or the
+        external configuration.
+        """
+        try:
+            deep_linking_enabled = consumer.lti_dl
+            return deep_linking_enabled is not None
```
ashultz0, line 969: it can never be none because lti_dl can't return none, only an exception - lti_dl behavior is already confusing in this PR

lti_consumer/lti_xblock.py
```diff
@@ -922,6 +958,28 @@ def is_past_due(self):
...
+        except LtiAdvantageServiceNotSetUp:
+            return False
+
+    def _get_lti_advantage_deep_linking_launch_url(self, consumer):
+        """
+        Return the LTI 1.3 Advantage Deep Linking launch URL. Use the LTI consumer to get the deep linking launch URL
+        instead of directly accessing the xBlock fields, the LTI Configuration model, or the external configuration.
+        """
+        try:
+            return consumer.lti_dl.deep_linking_launch_url
```
ashultz0, line 979: this will return a URL even if lti_dl is false, is that legit?

lti_consumer/lti_xblock.py
ashultz0: I think it's ok to have a version attribute on the consumer class - using the name is just a brittle horrible version of doing that anyway.

You could have some sort of lookup table that maintains a mapping of the class itself to the version but again that's just a fancy version of having each class have to provide a version attribute. If in future we make LtiConsumerMagic1p3 that inherits from LtiConsumer1p3 it is a good feature that it would inherit version = lti_1p3 

If we make LtiConsumer1p3p5 we could override the version.

Or conversely you could make the consumer have a supports_version(string) function that lets you know that the lti1p3 supports 1.3 but not 1.3.5 and lti1p3p5 supports both, but that is maybe more fancy than we need now

lti_consumer/lti_xblock.py
```diff
@@ -1506,17 +1572,23 @@ def _get_context_for_template(self):
...
         launch_url = self.launch_url
+        lti_1p3_launch_url = self.lti_1p3_launch_url.strip()
+
+        lti_consumer = None
+        if self.config_type in ("external", "database"):
+            lti_consumer = self._get_lti_consumer()
 
-        if self.config_type == "external":
-            launch_url = self._get_lti_consumer().lti_launch_url
+        # The lti_launch_url property only exists on the LtiConsumer1p1.
```
ashultz0, line 1589: if config is external here we will set launch_url no matter what the version is in the boolean logic below - either this comment should change or the below should be config in (external, database) AND version = 1.1

lti_consumer/lti_xblock.py
```diff
@@ -1506,17 +1572,23 @@ def _get_context_for_template(self):
...
+        if self.config_type == "external" or (self.config_type == "database" and self.lti_version == "lti_1p1"):
+            launch_url = lti_consumer.lti_launch_url
 
         # As we currently support only LTI 1.1 for external configuration, let's simply
         # use the same logic as the LTI 1.1 for getting the launch handler.
```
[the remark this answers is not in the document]
ashultz0: the code below looks pretty suspect in that it will do some stuff with version and external just like above

maybe we can compact this together and do the boolean logic less often and set more variables in each branch

lti_consumer/static/js/xblock_studio_view.js
```diff
@@ -98,14 +98,15 @@ function LtiConsumerXBlockInitStudio(runtime, element) {
...
+     *  database - Do not show the LTI 1.1/1.3 config fields
      *  external - Show only the External Config ID field
      */
     function toggleLtiConfigType() {
         const configType = $(element).find('#xb-field-edit-config_type').val();
         const configFields = lti1P1FieldList.concat(lti1P3FieldList, ['lti_version']);
 
-        if (configType === "external") {
-            // hide the lti_verison and all the LTI 1.1 and LTI 1.3 fields
+        if ((configType === "external") || (configType === "database")) {
```
ashultz0, line 108: you use configtype in (external, database) above and it's easier to read :)

lti_consumer/api.py
Zacharis278: Definitely a fan of this idea. I think passing around the value of config_type tends to get a bit messy because again it's nicer to have things come from the model and not a string off the xblock.

lti_consumer/models.py
Zacharis278: makes sense to me 👍

Zacharis278:
@Agrendalath / @tecoholic I think the additions here and in #254 are mostly compatible since we're making somewhat adjacent changes. The biggest conflicting decision I see is we've added the LTI 1.3 fields explicitly to the model instead of just storing it all in the 'lti_config' JSON blob.

Would like to get your feedback if you see anything in here that is problematic for you since we'll need to merge these feature relatively soon.

lti_consumer/api.py
Zacharis278: does this actually get solved by the calls to `get_or_update_config` in #254? This kinda feels like an unexpected side-effect to "get or create" but I'm not sure where it belongs.

lti_consumer/lti_xblock.py
Zacharis278: something we could look at here. You can get the consumer class through the LTIConfiguration model. At first glance this actually seems like a preferable way to do it since you get the lti config in scope for the version and the helper _get_lti_consumer just calls that anyway in a roundabout manner 🤔

tecoholic:
@Zacharis278 Thanks for comparing the changes. I reviewed the changes here as well. Here are a couple of things that stand out for me:

1. Storing the LTI 1.3 fields in the `lti_config` JSON field vs individual columns on the model - I was mostly working with the perspective of getting this information from `external` sources and having them as a dict made sense. Although it might not be the best in general - editing values from Admin interface for example will be hairy. So this could be something that #254 can definitely make use of.
2. The complexity around accessing LTI version seems to surface in a couple of locations and this PR considers the `LtiConfiguration` Model to be an independent source of truth. #254 takes a more user oriented approach where it considers the XBlock to be the ultimate source of truth. So for example, if a course creator changes a LTI 1.3 attribute on Studio, the DB would be updated with the values from the XBlock. This allows an XBlock to change tools, versions ..etc., without needing admin intervention (at least, that's the idea).
3. #254 takes the view that DB be the only config storage going forward, as it also moves the LTI 1.3 endpoints away from XBlock into the Django Plugin. This again takes a user viewpoint, as in the distinction between storing the configuration in XBlock (MongoDB) and Database (MySQL) shouldn't be the concern of the course creator. To that effect, 254 [copies the values on to the DB](https://github.com/openedx/xblock-lti-consumer/pull/254/files#diff-cc618c2e63b8c2d293d245f9799381d9321e40fe2d05e9a4b2f80bb91995157dR1045) every time an XBlock is updated, and to provide backward compatibility for XBlocks which are currently in use (config on xblock), it copies it over whenever [the handler is directly called](https://github.com/openedx/xblock-lti-consumer/pull/254/files#diff-cc618c2e63b8c2d293d245f9799381d9321e40fe2d05e9a4b2f80bb91995157dR1141) as well.

If you think the approach in 2 & 3 makes sense, I think it would drop the version complexity for this PR. Since we plan to support `external` sources as a possible source of truth, relying on DB as an independent source of truth will be a difficult design to maintain. By allowing all 3 XBlock, DB and External be sources of truth for their own usecases, we could support existing XBlock, and new non-XBlock LTI integrations as well.

cc: @Agrendalath

Zacharis278:
The the ultimate goal here is to launch 1.3 tools that don't have any xblock context. Those tools would be set up by another application outside of studio which would save values to the LTIConfiguration model. (we do this today with [1.1 already](https://github.com/openedx/edx-platform/blob/9795c7f5aa22bd3aff607229b5ef7f2135c01acf/openedx/core/djangoapps/course_live/views.py#L28)) So the DB config option introduced in studio+admin here is really just a matter of making this functional/testable right now. With the introduction of 254 I don't think it's necessary that this continue to exist as a separate option, we can simply merge this with the default behavior introduced there. The only sticking point is making sure the default behavior can get everything it needs based on a configuration id if no xblock is provided.

To that end I think I'm an agreement on 2 and 3 above. 

Does an eventual end state where we can build this consumer with or without an xblock (provided the config id is passed instead) make sense?

I could go either way on which of these to merge first, so my instinct is whatever is ready the soonest. If it is this one, we can simply TODO some of the lti_version weirdness since it's functional but not ideal and fix that later.

tecoholic:
@Zacharis278 

> Does an eventual end state where we can build the consumer with or without an xblock (provided the config id is passed instead) make sense?

100% on the XBlock free LTI Launch. 254 is aimed at making that possible by moving all the endpoints from XBlock handlers into the Django plugin and making the XBlock depend on Django's models and views.

I think 260 is something that can be used by 254. So, I think we should merge 260 first once it attains a functional and stable state and them refactor 254 to adapt to the changes of 260.

Zacharis278:
@tecoholic sounds great, we talked through this late yesterday and came to the same conclusion. We'll work on getting this into a mergeable state.

lti_consumer/lti_xblock.py
MichaelRoytman: Thanks for the thoughts! Using `block.lti_version` means I don't have to deal with this issue!

lti_consumer/lti_xblock.py
MichaelRoytman: You're right. We don't support the external config for LTI 1.3, so I must have been relying on that, but I don't like that implicit condition. I'll make it explicit. Thanks!

lti_consumer/static/js/xblock_studio_view.js
MichaelRoytman: This is Javascript, so I can't do that here. This is the cleanest way I know to accomplish this.

tecoholic:
@MichaelRoytman 
> Our thought was it would be set on the LtiConfiguration model. However, this will prevent anyone from setting the config_type to "database" on the XBlock. The "database" type is meaningful, so I have decided to keep it as an option.

The approach in #254 is to migrate all the XBlocks to `CONFIG_ON_DATABASE` without any user interaction. Since the HTTP handlers are all moved to the Django Views, there is a hard requirement to migrate all the configuration from XBlocks to the DB. 

The migration happens whenever the XBlock is edited from Studio (automatically, without user input) and also whenever the LTI pre-launch handler or the access token handler is called. So, it feels like this option would never be used in reality. Maybe I could be wrong, I might be missing a scenario. Since we already have it built, I think we can include it now and remove it later, if there really is no use?

Zacharis278 (approved):
this looks good to me, I'd say we can move forward

MichaelRoytman:
@tecoholic I'm just pinging to let you know I merged this in. Thank you for your feedback.

@giovannicimolin I wanted to ping you about this as well. I would have requested your feedback, but I knew you were on PTO, so if you have any feedback after-the-fact, I'm happy to address it.

tecoholic:
@MichaelRoytman Thank you. I will rebase 254 on to this and make necessary changes.

PR #266, opened 2022-07-21
fix: fix hidden lti_version and 1.3 fields in edit view and incorrect menu behavior in Javascript

MichaelRoytman:
This commit fixes three bugs.

1. The first bug is that the lti_version field is inappropriately hidden in the Studio author view edit menu when the selected config_type is database.

2. The second bug is that the editable_fields property of the LtiConsumerXBlock is inappropriately excluding LTI 1.3 fields when the config_type is database. The editable_fields property should include LTI 1.3 fields even when the config_type is database, because the Javascript defined in xblock_studio_view.js may want to show these fields if the user selects a different config_type in the menu. We want to support a dynamic edit menu, so these fields must be considered editable by the XBlock in order for the Javascript to be able to manipulate them.

3. The third bug is in inconsistent rendering of the Studio author view edit menu. Depending on the order in which a user selects lti_version, config_type, or lti_1p3_tool_key_mode, different sets of fields are displayed, due to the overlapping sets of rules that govern what fields should be hidden or shown for a given field selection. This commit corrects this inconsistent rendering by first showing all fields and then gradually hiding fields depending on the sets of rules, for each change to the fields.


schenedx (approved):
I have no issue with the change here.
However, does this present more fields to the Instructor or LTI setup person? If so, we need to give heads to support and update docs so Instructors are not confused by the additional fields unhidden?

Zacharis278 (approved):
lgtm

MichaelRoytman:
@schenedx Thanks for noting that. This doesn't uncover any more fields. The changes to `editable_fields` fixes a bug that was actually not allowing the edit menu to show fields that are _supposed_ to be visible.

 For example, without these changes, if the current `config_type` on the XBlock were to be `database`, then `editable_fields` would _not_ include the LTI 1.3 fields. If the user were to change the `config_type` in the edit menu to `new` (which should show the LTI 1.3 fields, because they are no longer being stored on the database), then the LTI 1.3 fields would be missing. This is because the `config_type` hasn't been committed to the data, so the LTI 1.3 fields wouldn't be in `editable_fields`. This change makes the Javascript responsible for hiding and showing fields. No new fields are exposed that weren't already exposed.

This change is also behind a waffle flag and may be significantly changed by [BB-55999. ](https://github.com/openedx/xblock-lti-consumer/pull/254), so I think the risk is very low. If and when we decide to release the database configuration, we will let support know and update any docs.

PR #270, opened 2022-08-02
fix: missing exception details in logs

MichaelRoytman:
The error handler in `LtiConsumerXBlock.lti_1p3_launch_callback` logs a warning when a select set of exceptions are handled. That log does not contain useful information about the nature of the exception, because the exceptions were not being instantiated with error messages. The try...catch is a large block that contains code that can raise a multitude of errors, so these changes will enable better debugging.

This commit:
* adds helpful messages to the raised exceptions.
* adds the `exc_info=True` argument to include the stack trace of the handled exception.
* adds `ValueError` and `TypeError` to the list of handled exceptions, because the code can raise exceptions of these types.

This is a sample log before.

```
2022-08-01 21:26:51,992 WARNING 5544 [lti_consumer.lti_xblock] [user 3] [ip 172.28.0.1] lti_xblock.py:1278 - Error preparing LTI 1.3 launch for block 'block-v1:edX+DemoX+Demo_Course+type@lti_consumer+block@b98a9382452540a686b6f9de5e619cf1': 
```

This is a sample log after.

```
2022-08-02 13:17:27,727 WARNING 5974 [lti_consumer.lti_xblock] [user 3] [ip 172.28.0.1] lti_xblock.py:1277 - Error preparing LTI 1.3 launch for block 'block-v1:edX+DemoX+Demo_Course+type@lti_consumer+block@b98a9382452540a686b6f9de5e619cf1': The content_type is not supported by LTI Deep Linking.
Traceback (most recent call last):
  File "/edx/src/xblock-lti-consumer/lti_consumer/lti_xblock.py", line 1233, in lti_1p3_launch_callback
    lti_consumer.lti_dl.get_lti_deep_linking_launch_claim(accept_types='gibberish')
  File "/edx/src/xblock-lti-consumer/lti_consumer/lti_1p3/deep_linking.py", line 45, in get_lti_deep_linking_launch_claim
    raise exceptions.LtiDeepLinkingContentTypeNotSupported('The content_type is not supported by LTI '
lti_consumer.lti_1p3.exceptions.LtiDeepLinkingContentTypeNotSupported: The content_type is not supported by LTI Deep Linking.
```

### DIFF
--- a/lti_consumer/__init__.py
+++ b/lti_consumer/__init__.py
@@ -4,4 +4,4 @@ Runtime will load the XBlock class from here.
 from .apps import LTIConsumerApp
 from .lti_xblock import LtiConsumerXBlock
 
-__version__ = '4.3.2'
+__version__ = '4.3.3'
--- a/lti_consumer/lti_1p3/consumer.py
+++ b/lti_consumer/lti_1p3/consumer.py
@@ -363,7 +363,7 @@ class LtiConsumer1p3:
         # Check if all required claims are present
         for required_claim in LTI_1P3_ACCESS_TOKEN_REQUIRED_CLAIMS:
             if required_claim not in token_request_data.keys():
-                raise exceptions.MissingRequiredClaim()
+                raise exceptions.MissingRequiredClaim(f'The required claim {required_claim} is missing from the JWT.')
 
         # Check that grant type is `client_credentials`
         if token_request_data['grant_type'] != 'client_credentials':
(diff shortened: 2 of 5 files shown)


schenedx (approved):
👍 Nice

Zacharis278 (commented):
it seems like each of these custom exceptions should have its own default message so we don't have to type it out in each instance. Is there a good way to do that?

giovannicimolin:
:+1:

I've just suggested this on a separate PR before bumping into these changes.

Zacharis278 (approved):
👍

PR #278, opened 2022-08-19
Handle LtiError Error During LTI 1.1 Launch for Unauthenticated User

MichaelRoytman:
[[MST-1585]](https://2u-internal.atlassian.net/browse/MST-1585)

**Description**

In the LTI 1.1 launch handler, we set the user context, including the `user_id`. We do this by calling to the LMS's `DjangoXBlockUserService` to get information about the user. Sometimes, the user is unauthenticated. Sometimes, this is because the user is a web crawler. Other times, the user is a real user, but we do not know why the user is unauthenticated. We have some theories, but we have been unable to confirm them. Regardless, we should not surface a 500 error to the user.

This pull request adds handling for the `LtiError` that is raised when a user is unauthenticated during an LTI 1.1 launch. It catches the `LtiError` and renders an error page. The error page that was used for LTI 1.3 launches, formerly named `lti_1p3_launch_error.html`, has been renamed to `lti_launch_error.html` to reflect the fact that it is used for both LTI 1.1 and 1.3 launches. It was modified to remove the reference to the version of LTI used by the XBlock; these details are unnecessary for a learner, and removing them allows us to reuse a single template for both LTI versions.

This pull request also reverts a previous commit that added additional logging to help investigate the error.

The below photos are from devstack, so the UI looks a little different because debug mode is on.

**Before**

**Modal Launch**

Screen Shot 2022-08-19 at 8 32 21 AM

**New Window Launch**

Screen Shot 2022-08-19 at 8 36 48 AM

**After**

**Modal Launch**

Screen Shot 2022-08-19 at 8 38 46 AM

**New Window Launch**

Screen Shot 2022-08-19 at 8 37 24 AM

### DIFF
--- a/lti_consumer/lti_xblock.py
+++ b/lti_consumer/lti_xblock.py
@@ -745,7 +745,11 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
         """
         Get system user role and convert it to LTI role.
         """
-        role = self.runtime.service(self, 'user').get_current_user().opt_attrs.get('edx-platform.user_role', 'student')
+        user = self.runtime.service(self, 'user').get_current_user()
+        if not user.opt_attrs["edx-platform.is_authenticated"]:
+            raise LtiError(self.ugettext("Could not get user data for current request"))
+
+        role = user.opt_attrs.get('edx-platform.user_role', 'student')
         return ROLE_MAP.get(role, 'Student,Learner')
 
     @property
@@ -978,6 +982,10 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
         Extract and return real user data from the runtime
         """
         user = self.runtime.service(self, 'user').get_current_user()
+
+        if not user.opt_attrs["edx-platform.is_authenticated"]:
+            raise LtiError(self.ugettext("Could not get user data for current request"))
+
         user_data = {
             'user_email': None,
             'user_username': None,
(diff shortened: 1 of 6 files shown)


lti_consumer/lti_xblock.py
```diff
@@ -1111,13 +1091,22 @@ def lti_launch_handler(self, request, suffix=''):  # pylint: disable=unused-argu
...
-            self.user_id,
-            self.role,
-            result_sourcedid=self.lis_result_sourcedid,
-            person_sourcedid=username,
-            person_contact_email_primary=email
-        )
+        # Occassionally, users try to do an LTI launch while they are unauthenticated. It is not known why this occurs.
+        # Sometimes, it is due to a web crawlers; other times, it is due to actual users of the platform. Regardless,
+        # return a 400 response with an appropriate error template.
+        try:
```
ashultz0, line 1097: this try feels a little weird to me because what can throw the error is self.user_id, not set_user_data which cannot throw any errors

also why doesn't extract_real_user_data above also fail?

maybe a little more restructuring to make this function well ordered:
first get the consumer
then get the various user fields inside a try block indicating "there might not be a user"
then set the user data off those fields which can't fail

MichaelRoytman: I figured that every call to the user service should care whether the user is authenticated, so I added that logic.

ashultz0: thanks, this is clearer

ashultz0 (approved):

PR #281, opened 2022-09-12
Add ADR for decoupling LTI launches from the XBlock

MichaelRoytman:
This commit adds an ADR describing our strategy for decoupling LTI launches from the XBlock and from the edX platform.

### DIFF
--- a/docs/decisions/0007-lti-launch-decoupling.rst
+++ b/docs/decisions/0007-lti-launch-decoupling.rst
@@ -1,0 +1,170 @@
+0007 Decouple LTI Launches from XBlock and edX Platform
+#######################################################
+
+Status
+******
+**Draft (=> Accepted)**
+
+Context
+*******
+
+**Definitions**
+
+In this document, the terms "launch context" and "launch information" are used. Launch context describes the
+technological environment or application in which an LTI launch occurs. For example, the launch context may be an
+XBlock, a course, a Django application, an independently deployable application (IDA), etc. Launch information describes
+information included in or necessary for an LTI launch. Launch information is relative to the launch context. For
+example, a user identifier or a user role is launch information. Launch information may refer to data that are used as
+query or form parameters of a request, as claims in a JSON web token (JWT), or as data needed by the library to
+function.
+
+**Problems**
+
+Open edX would like to support the following LTI launch use cases:
+
+* content launches from an XBlock launch context
+* content launches from multiple XBlocks, reusing LTIConfigurations between them (i.e. course launch context)
+* content launches from outside an XBlock launch context in the platform (e.g. LTI course tabs, LTI 1.3 embedded
+  discussions)
+* content launches from an independently deployable application (i.e. IDA or Django application context)
+
+This library implements both LTI 1.1 and LTI 1.3 launches.
+
+**LTI 1.1**
+
+Currently, LTI 1.1 launches are already decoupled from the XBlock and the edx-platform. The `lti_embed
+<https://github.com/openedx/xblock-lti-consumer/blob/master/lti_consumer/lti_1p1/contrib/django.py#L13>`_ function
+accepts a series of arguments that define necessary launch information and returns an HTML template with Javascript that
+initiates an LTI 1.1 launch. No additional work is required to decouple LTI 1.1 launches, but this document will
+describe changes necessary to bring LTI 1.1 launches in alignment with the below strategy for LTI 1.3 launches.
+
+**LTI 1.3**
+
+LTI 1.3 launches rely on the ``location`` field of the ``LtiConfiguration`` model. The ``location`` field is
+a ``UsageKeyField`` and represents the location of the XBlock associated with the ``LtiConfiguration`` instance. This
+association was made because, originally, the XBlock was the source of truth for LTI configuration information and was
+the only launch context in which LTI launches occurred.
+
+The dependence on the ``location`` field prevents LTI 1.3 launches from being possible outside the launch context of an
+XBlock. This prevents Open edX from using LTI more widely across the platform and across the ecosystem, because its use
+is restricted to specific launch contexts in which ``location`` is meaningful.
+
+For example, because ``LTIConfiguration`` instances are associated with a particular ``location``, it is not possible
+for a ``LtiConfiguration`` to be reused across different XBlocks (i.e. a course launch context), even if the
+``LtiConfiguration`` would be identical amongst the different XBlocks.
+
+In an LTI 1.3 launch, the ``location`` field is used to determine the
+``https://purl.imsglobal.org/spec/lti/claim/roles`` and ``https://purl.imsglobal.org/spec/lti/claim/context`` claims of
+the LTI 1.3 launch. In an LTI 1.3 launch, the ``https://purl.imsglobal.org/spec/lti/claim/roles`` claim is required, and
+the ``https://purl.imsglobal.org/spec/lti/claim/context`` claim is optional. This couples the LTI 1.3 launches to the
+XBlock through the ``location`` field.
+
+Furthermore, LTI 1.3 launches are also dependent on functions imported from edx-platform. These functions are used `to
+get the requesting user's role
+<https://github.com/openedx/xblock-lti-consumer/blob/master/lti_consumer/plugin/views.py#L182>`_, `to get the requesting
+user's external ID <https://github.com/openedx/xblock-lti-consumer/blob/master/lti_consumer/plugin/views.py#L183>`_, `to
+get the course key from the location field
+<https://github.com/openedx/xblock-lti-consumer/blob/master/lti_consumer/plugin/views.py#L181>`_, etc. This prevents LTI
+1.3 launches from running outside of the platform.
+
+https://github.com/openedx/xblock-lti-consumer/pull/254 decoupled the LTI 1.3 launch handler endpoint
+``launch_gate_endpoint``, the LTI 1.3 keyset endpoint ``public_keyset_endpoint``, and the LTI 1.3 access token endpoint
+``access_token_endpoint`` from the XBlock by moving these endpoints into the Django plugin. This was the first step in
+decoupling the XBlock from the LTI 1.3 launch. This decision describes how to fully decouple the LTI 1.3 launch from the
+XBlock and how to modify the LTI 1.1 launch to follow this pattern.
+
+Decision
+********
+
+**LTI 1.3**
+
+* For backwards compatability, we will not remove references to the ``location`` field from the
+  ``public_keyset_endpoint`` and ``access_token_endpoint`` endpoint. This is because LTI 1.3 integrations may exist in
+  the platform and in other installations that use URLs that contain the
+  ``usage_key`` in them.
+* We will change the URLs that are displayed in Studio for the ``public_keyset_endpoint`` and
+  ``access_token_endpoint` endpoints. This will prevent additional LTI 1.3 configurations that use the ``usage_key``. This
+  makes us more able to deprecate URLs that use ``usage_key`` and to remove references to ``location``.
+* The responsibility of determining the launch information included in an LTI launch (e.g. roles and context claims)
+  will be that of the launch context in which the LTI 1.3 launch occurs (e.g. XBlock, IDA).
+* For LTI 1.3 launches, launch information will be encoded in an instance of a launch data class called
+  Lti1p3LaunchData.
+* The library will expose a Python API function that will return the URL that initiates the LTI 1.3 launch - the third
+  party initiated login URL. This Python API function may be `get_lti_1p3_content_url
+  <https://github.com/openedx/xblock-lti-consumer/blob/master/lti_consumer/api.py#L186>`_.
+* Each launch context will supply the launch information to the LTI 1.3 launch by including an instance of
+  Lti1p3LaunchData as an argument to the aforementioned function. Each launch context will call this function to fetch
+  the URL to trigger an LTI 1.3 launch.
+* In order for this launch information to be available throughout the LTI 1.3 launch, the library will cache the the
+  launch information provided by the launch context in the third-party initiated login request and include the cache_key
+  in the ``lti_message_hint`` query or form parameter. This allows the library to read the launch_data out of the cache
+  when necessary by using the ``lti_message_hint``.
+* The third-party initiated login request ``login_hint`` query or form parameter will be used to include the user
+  identifier. The ``login_hint`` will not be used to identify the ``LtiConfiguration`` instance. Information necessary
+  to identify the ``LtiConfiguration`` instance will be included in the launch information instead.
+* We will not describe how to decouple LTI Advantage Services in this document. This may be done at a later time.
+
+**LTI 1.1**
+
+* For LTI 1.1 launches, launch information will be encoded in an instance of a launch data class called
+  Lti1p1LaunchData.
+* In order to have a similar interface to LTI 1.3, we will modify the lti_embed function to use an instance of the
+  Lti1p1LaunchData class.
+* We will modify the ``LtiConsumerXBlock`` to use the `lti_embed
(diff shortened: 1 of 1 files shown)


docs/decisions/0007-lti-1p3-decoupling.rst
```diff
@@ -0,0 +1,115 @@
...
+* The third-party initiated login request ``login_hint`` query or form parameter will be used to include the user
+  identifier. The ``login_hint`` will not be used to identify the ``LtiConfiguration`` instance. Information necessary
+  to identify the ``LtiConfiguration`` instance will be included in the ``lti_message_hint`` query or form parameter
+  instead.
+* We will not describe how to decouple LTI Advantage Services in this document. This may be done at a later time.
+
+Consequences
+************
+
+* LTI 1.3 launches will be decoupled from the XBlock and from the edX platform. This will allow Open edX to better
```
MichaelRoytman, line 81: I haven't been able to find a definitive answer to this question, but I'm concerned about the URL length for GET requests if we stuff all this information into the query parameter. We may need to enforce that contexts send POSTs to the LTI 1.3 launch initiation URL instead of GETs, which is a consequence. [Here](https://stackoverflow.com/a/417184) is some advice on the topic.

ashultz0: this makes sense, although I worry that we will run into an external tool which demands GET.

docs/decisions/0007-lti-1p3-decoupling.rst
```diff
@@ -0,0 +1,115 @@
...
+#########################################################
+
+Status
+******
+**Draft (=> Accepted)**
+
+Context
+*******
+
+In this document, the terms context and launch information are important. Context describes the technological
```
schenedx, line 11: When you used words `context` and `Launch Information`, I believe these are code variable names, right? Can you make that explicit? I confused `context` with the title of this section.

schenedx (approved):
I think the decision makes a lot of sense. Even if there are some fine tuning needed associated with putting URL into the function as inputs, in the future, I think the spirit of this ADR is still important to convey.

docs/decisions/0007-lti-1p3-decoupling.rst
```diff
@@ -0,0 +1,115 @@
...
+
+Furthermore, LTI 1.3 launches are also dependent on functions imported from edx-platform. These functions are used `to
+get the requesting user's role
+<https://github.com/openedx/xblock-lti-consumer/blob/master/lti_consumer/plugin/views.py#L182>`_, `to get the requesting
+user's external ID <https://github.com/openedx/xblock-lti-consumer/blob/master/lti_consumer/plugin/views.py#L183>`_, `to
+get the course key from the location field
+<https://github.com/openedx/xblock-lti-consumer/blob/master/lti_consumer/plugin/views.py#L181>`_, etc. This prevents LTI
+1.3 launches from running outside of the platform.
+
+Open edX would like to support the following LTI use cases:
```
Zacharis278, line 44: nice 👍 

I think this really clearly states why we are doing this even if the audience doesn't have a deep knowledge of LTI yet. A suggestion tho, maybe we lead with this earlier in this section before getting into technical details?

docs/decisions/0007-lti-1p3-decoupling.rst
```diff
@@ -0,0 +1,115 @@
...
+  to identify the ``LtiConfiguration`` instance will be included in the ``lti_message_hint`` query or form parameter
+  instead.
+* We will not describe how to decouple LTI Advantage Services in this document. This may be done at a later time.
+
+Consequences
+************
+
+* LTI 1.3 launches will be decoupled from the XBlock and from the edX platform. This will allow Open edX to better
+  leverage LTI across the platform and across the ecosystem.
+* LTI Advantage Services will not be decoupled from the XBlock and from the edX platform. This means that LTI Advantage
```
Zacharis278, line 83: a side consequence we may want to mention is how this decision impacts the LTIConfiguration data model. Because we haven't tackled these advantage services yet location still needs to exist as a field right?

docs/decisions/0007-lti-1p3-decoupling.rst
```diff
@@ -0,0 +1,115 @@
+0007 Decouple LTI 1.3 Launch from XBlock and edX Platform
```
Zacharis278, line 1: Does this decision really apply to only LTI 1.3? It seems like we should deciding we want these features for any launch, it's just the 1.3 implementation that is most impacted by this because of the extra steps involved.

ashultz0: boot camp LTI tools turn out to be currently LTI 1.1 and probably need to be decoupled from the blocks

Zacharis278: best news I've heard all day 😂  (that these tools aren't 1.3/advantage).

You can kinda do a 'decoupled' 1.1 launch already but IMO we should commit to a similar interface for both even if 1.1 is simple enough to hack on from the outside.

jristau1984: Just FYI - We were told yesterday that Bootcamps will be updating their integrations to LTI 1.3 before the beginning of the calendar year 2023.

ashultz0: I think if it's just another couple of tickets to do 1.1 as well that would insulate us from the 1.3 transition slipping and be useful for the community

docs/decisions/0007-lti-1p3-decoupling.rst
giovannicimolin: @MichaelRoytman We can't choose what method tools will use to perform the LTI launch, so we can go with the worst-case scenario here and model our parameters to fit that size.

Instead of using a JWT we can try using something smaller with small parameter names?

docs/decisions/0007-lti-1p3-decoupling.rst
```diff
@@ -0,0 +1,115 @@
...
+``launch_gate_endpoint``, the LTI 1.3 keyset endpoint ``public_keyset_endpoint``, and the LTI 1.3 access token endpoint
+``access_token_endpoint`` from the XBlock by moving these endpoints into the Django plugin. This was the first step in
+decoupling the XBlock from the LTI 1.3 launch. This decision describes how to fully decouple the LTI 1.3 launch from the
+XBlock.
+
+Decision
+********
+
+* We will remove references to the ``location`` field from all views that implement the generic LTI 1.3 launch. This
+  includes the ``launch_gate_endpoint``, ``public_keyset_endpoint``, and ``access_token_endpoint`` views.
```
giovannicimolin, line 61: @MichaelRoytman @Zacharis278 The `access_token_endpoint` needs contextual information in order to select which scopes and services it'll allow the tool to use. [Check the access_token method here](https://github.com/openedx/xblock-lti-consumer/blob/master/lti_consumer/lti_1p3/consumer.py#L343).

For this, we should take the simple approach and store the LTI Advantage service configuration directly in the LTIConfiguration model and let all instances re-using that configuration have access to the same set of services (instead of storing LTI Adv configuration for each placement).

What do you say?

docs/decisions/0007-lti-1p3-decoupling.rst
giovannicimolin: I think we can find easy replacements for the location field for Deep Linking and Assignments and Grades services. 
These are the methods that can potentially break LTI Advantage when removing the the reference:
1. https://github.com/openedx/xblock-lti-consumer/blob/master/lti_consumer/signals.py#L16 - Links grades back to LMS (we should probably split this into (1) a signal emitted by the pure LTI function and (2) a receiver on the block side to react and link the grade).
2. Permissions and authentication for REST API endpoints: https://github.com/openedx/xblock-lti-consumer/tree/0a1f26234f9515524d117f5032ad3d8ff2bfb871/lti_consumer/lti_1p3/extensions/rest_framework
3. Declarative LTI AGS mode: https://github.com/openedx/xblock-lti-consumer/blob/0a1f26234f9515524d117f5032ad3d8ff2bfb871/lti_consumer/models.py#L420
4. This `UsageKey` reference: https://github.com/openedx/xblock-lti-consumer/blob/0a1f26234f9515524d117f5032ad3d8ff2bfb871/lti_consumer/models.py#L596

giovannicimolin (approved):
:+1: for these changes, great work!

There's still a few things we need to figure out, so I left comments on the ADR, but we can address those as we implement this on the codebase.

docs/decisions/0007-lti-1p3-decoupling.rst
Zacharis278: I hadn't considered how large the parameter count could theoretically grow here. However, most fields could be looked back up again based on the configuration so any 'launch fields' (such as the attempt id for proctoring) should be pretty limited right?

If we're still uncomfortable with how this would scale I might suggest an alternative. We persist a server side 'launch' object and instead of encoding it in the ``lti_message_hint`` and we simply send along a lookup key instead. This would be pretty straightforward to do with a simple cache. It isn't too different from the suggested approach so we could always see if we run into trouble and pivot later.

giovannicimolin: @Zacharis278 :+1: for storing a server-side object with all the launch parameters and just passing a token/identifier through the launch message.

docs/decisions/0007-lti-1p3-decoupling.rst
MichaelRoytman: That sounds good. I'll update the ADR to mention that we'll modify the LTI 1.1 embed function to get configuration data in the same way.

docs/decisions/0007-lti-1p3-decoupling.rst
MichaelRoytman: @giovannicimolin 

> The access_token_endpoint needs contextual information in order to select which scopes and services it'll allow the tool to use. [Check the access_token method here](https://github.com/openedx/xblock-lti-consumer/blob/master/lti_consumer/lti_1p3/consumer.py#L343).

What additional contextual information does this endpoint need? I see that it needs a `UsageKey` or a `config_id` to fetch the corresponding `LtiConfiguration` instance, but is there any additional context that's not already stored in the `LtiConfiguration`? I'm not seeing other necessary information. Can you point me at it? It looks like the library grants all scopes that the consumer/tool requests [here](https://github.com/openedx/xblock-lti-consumer/blob/master/lti_consumer/lti_1p3/consumer.py#L386), without referring to any contextual information. I am not suggesting that we remove the use of the `LtiConfiguration` model. Rather, I'm suggesting that we refer to `LtiConfigurations` by their `config_id` instead of their `location`.

> For this, we should take the simple approach and store the LTI Advantage service configuration directly in the LTIConfiguration model and let all instances re-using that configuration have access to the same set of services (instead of storing LTI Adv configuration for each placement).

This is referring to the case where multiple placements of an LTI link will share a single `LtiConfiguration`, right?

If multiple LTI link placements share a single `LtiConfiguration`, so long as `get_lms_lti_access_token_link(location)` is called with the same `config_id` when generating the access token to provide to the tool, the same `LtiConfiguration` instance will be used to generate the access token for the tool, so that should address what you're saying.

Eventually, we'd like for `get_lms_lti_access_token_link` to only accept a `config_id`.


I am not very familiar with LTI Advantage services just yet, so please let me know if I'm misunderstanding something.

docs/decisions/0007-lti-1p3-decoupling.rst
MichaelRoytman: What are you referring to as the "launch model"? Do you mean the set of launch data that is defined in the context and passed by the context to the library's API methods?

What is the potential schema of an `LtiPlacement`? I understand it would store a `resource_link_id`, right? I suppose I'm trying to understand the difference between what is in `launch_data` and what would be in `LtiPlacement`.

Just taking a cursory look through the various claims, I could see `resource_link_id` and `launch_presentation` as placement specific values. What about `context`? We could share `LtiConfigurations` across `contexts` (e.g. courses), so this could be considered a placement-specific value. So, where would we draw the line between what's in `launch_data` and what's in an `LtiPlacement`?

I'm not as familiar with deep linking yet. What deep linking configuration could go in `LtiPlacement`?

docs/decisions/0007-lti-1p3-decoupling.rst
giovannicimolin: @MichaelRoytman 

>What additional contextual information does this endpoint need? I see that it needs a UsageKey or a config_id to fetch the corresponding LtiConfiguration instance, but is there any additional context that's not already stored in the LtiConfiguration? I'm not seeing other necessary information. Can you point me at it? 

You're right! It currently grants all available scopes, but there's a `TODO` in the code [here](https://github.com/openedx/xblock-lti-consumer/blob/master/lti_consumer/lti_1p3/consumer.py#L424) that mentions limiting scopes.

>This is referring to the case where multiple placements of an LTI link will share a single LtiConfiguration, right?

Yes, and you are right about this as well. The LTI configuration should handle the advantage configuration, while the tool just needs to use the right `ltiResourceLink` to get and set grades for each placement.

giovannicimolin: My original concerns were not with the token, but with the access levels that tools will have to learner data (this is the question I asked myself: does a tool configured to work platform-wide have access to **all users** through NRPS? I don't think it should). 

But I think this will be a different problem to tackle in a separate ADR. Lets keep this one about decoupling LTI 1.3 and handle the permission issues separately. Right now our main use case (XBlock) won't cause issues.

docs/decisions/0007-lti-1p3-decoupling.rst
giovannicimolin: @Zacharis278 I'm a bit late to reply, so I'm not sure this answer will be up-to-date. Feel free to ignore if so.

>Because we haven't tackled these advantage services yet location still needs to exist as a field right?

IIRC Not necessarily - as long as tools know which context they are being launched to, they should be able to send back unique identifiers. This only affects LTI advantage services, since they can store data that is placement specific.

E.g.: **AGS** stores grades for each placement, so it needs to know what is the location for each placement (in the Xblock launch case).
**DL** configures different placements to have different contents show up - so it needs to store a unique placement ID somewhere.

giovannicimolin: _Note:_ about my previous comment with the code links - those are just places I know to be breaking points, no need to do anything related to them here in the ADR, but good to link to who will be implementing the code changes.

docs/decisions/0007-lti-1p3-decoupling.rst
giovannicimolin: >Just taking a cursory look through the various claims, I could see resource_link_id and launch_presentation as placement specific values. What about context? We could share LtiConfigurations across contexts (e.g. courses), so this could be considered a placement-specific value. So, where would we draw the line between what's in launch_data and what's in an LtiPlacement?

I think that context should come from the placement, no? Each placement would represent a different XBlock being used somewhere in the platform and should yield different contexts.

>So, where would we draw the line between what's in launch_data and what's in an LtiPlacement?

Yeah, this is still a bit confusing - I haven't thought through this all year, just wanted to expose one of the options I originally came up with for consideration.

>I'm not as familiar with deep linking yet. What deep linking configuration could go in LtiPlacement?

None, but each `LtiDlContentItem` would have a Foreign Key to a placement instead of to the `LtiConfig`. Currently there's nothing differentiating between LTI DL items on each placement, meaning they won't work properly if we try to re-use configuration (e.g. with the current setup, any LTI DL configuration that is done for a block will be replicated to any other block using the same configuration, because `LtiDlContentItem` is tied to `LtiConfig`). 

Let me know if that is not clear - I can elaborate.

mphilbrick211:
Hi @Zacharis278! Looks like this is ready to merge!

MichaelRoytman:
@mphilbrick211 It's not ready to merge. I have a few changes I need to include. I'll get those in this week.

Zacharis278 (approved):
👍

PR #288, opened 2022-09-21
feat: decouple LTI 1.3 launch from the XBlockLtiConsumer

MichaelRoytman:
### Decouple LTI 1.3 Launch from LtiConsumerXBlock

Purpose
-------

The purpose of these changes is to decouple the LTI 1.3 launch from the LtiConsumerXBlock. It is in accordance with the ADR "0007 Decouple LTI 1.3 Launch from XBlock and edX Platform", which is currently under review. The pull request for the ADR is here: https://github.com/openedx/xblock-lti-consumer/pull/281.

The general premise of these changes is to shift the responsibility of defining key launch claims to users of the library. Such claims include user ID, user role, resource link ID, etc. Prior to this change, this context was defined directly in the launch view by referencing XBlock fields and functions, thereby tying the LTI 1.3 launch to the XBlock. By shifting the responsibility out of the view, we will be able to genericize the launch and make it functional in more contexts than just the XBlock and the XBlock runtime.

In short, the key launch claims are encoded in an instance of a data class Lti1p3LaunchData. Users of the library will instantiate this class with necessary launch data to it and pass the instance to various methods of the Python API to communicate the data to the library. Please see the aforementioned ADR for more details about this decoupling strategy.

Note that the majority of these changes affect only the basic LTI 1.3 launch. There have largely been no changes to LTI 1.3 Advantage Services. The one exception is the Deep Linking content launch endpoint. This is because this launch is implemented in the basic LTI 1.3 launch, and it was necessary to make the same changes to the deep linking content launch to ensure that it works properly. Otherwise, LTI 1.3 Advantage Services are out of scope of these changes. 

Change Summary for Developers
-----------------------------

Below is a summary of changes contained in this pull request.

* added an Lti1p3LaunchData data class
* added caching for Lti1p3LaunchData to limit data sent in request query or form parameters
* modified Python API methods to take Lti1p3LaunchData as a required argument
  * get_lti_1p3_launch_info
  * get_lti_1p3_launch_start_url
  * get_lti_1p3_content_url
* replaced references to LtiConsumerXBlock.location with Lti1p3LaunchData.config_id
* removed definition of key LTI 1.3 claims from the launch_gate_endpoint and instantiated Lti1p3LaunchData from within the LtiConsumerXBlock instead
* added a required launch_data_key request query parameter to the deep_linking_content_endpoint and refactored associated templates and template tags to pass this parameter in the request to the view

Change Summary for Course Staff and Instructors
-----------------------------------------------

The only changes relevant for course staff and instructors is that the access token and keyset URLs displayed in Studio have changed in format.

The old format was:

Access Token URL: `https://courses.edx.org/api/lti_consumer/v1/token/block-v1:edX+999+2022Q3+type@lti_consumer+block@714c10a5e4df452da9d058788acb56be`
Keyset URL: `https://courses.edx.org/api/lti_consumer/v1/public_keysets/block-v1:edX+999+2022Q3+type@lti_consumer+block@714c10a5e4df452da9d058788acb56be`

The new format is:

Access Token URL: `https://courses.edx.org/api/lti_consumer/v1/token/c3f6af60-dbf2-4f85-8974-4ff870068d43`
Keyset URL: `https://courses.edx.org/api/lti_consumer/v1/public_keysets/c3f6af60-dbf2-4f85-8974-4ff870068d43`

The difference is in the slug at the end of the URL. In the old format, the slug was the UsageKey of the XBlock associated with the LTI integration. In the new format, the slug is the config_id of the LtiConfiguration associated with the LTI integration. This is an iterative step toward decoupling the access_token_endpoint and the public_keyset_endpoint views from the XBlock location field. The XBlock location field appears as the usage_key parameter to both views. We cannot simply remove the usage_key parameter from the views, because existing LTI 1.3 integrations may have been created using the old format, and we need to maintain backwards compatibility. This change, however, prevents new integrations from being created that are coupled to the XBlock. In the future, we may address integrations that use the old format to fully decouple the XBlock from the views.

Testing
-------

Unit tests were added for all changes.

In addition, manual testing was performed using the instructions in the documents listed below.

* https://github.com/openedx/xblock-lti-consumer#lti-13
* https://openedx.atlassian.net/wiki/spaces/COMM/pages/1858601008/How+to+run+the+LTI+Validation+test

Resources
---------
JIRA: MST-1603: https://2u-internal.atlassian.net/browse/MST-1603

### DIFF
--- a/lti_consumer/api.py
+++ b/lti_consumer/api.py
@@ -9,9 +9,11 @@ import json
 
 from opaque_keys.edx.keys import CourseKey
 
+from lti_consumer.lti_1p3.constants import LTI_1P3_ROLE_MAP
 from .exceptions import LtiError
 from .models import CourseAllowPIISharingInLTIFlag, LtiConfiguration, LtiDlContentItem
 from .utils import (
+    get_lti_1p3_context_types_claim,
     get_lti_deeplinking_content_url,
     get_lms_lti_keyset_link,
     get_lms_lti_launch_link,
@@ -24,7 +26,7 @@ def _get_or_create_local_lti_config(lti_version, block_location,
                                     config_store=LtiConfiguration.CONFIG_ON_XBLOCK, external_id=None):
     """
     Retrieve the LtiConfiguration for the block described by block_location, if one exists. If one does not exist,
-    create an LtiConfiguration with the LtiConfiguration.CONFIG_ON_DB config_store.
+    create an LtiConfiguration with the LtiConfiguration.CONFIG_ON_XBLOCK config_store.
 
     Treat the lti_version argument as the source of truth for LtiConfiguration.version and override the
     LtiConfiguration.version with lti_version. This allows, for example, for
@@ -100,7 +102,11 @@ def get_lti_consumer(config_id=None, block=None):
     return _get_lti_config(config_id, block).get_lti_consumer()
 
 
-def get_lti_1p3_launch_info(config_id=None, block=None):
+def get_lti_1p3_launch_info(
+    launch_data,
+    config_id=None,
+    block=None,
+):
     """
     Retrieves the Client ID, Keyset URL and other urls used to configure a LTI tool.
     """
@@ -116,9 +122,9 @@ def get_lti_1p3_launch_info(config_id=None, block=None):
     deep_linking_enabled = lti_consumer.lti_dl_enabled()
 
     if deep_linking_enabled:
+        launch_data.message_type = "LtiDeepLinkingRequest"
         deep_linking_launch_url = lti_consumer.prepare_preflight_url(
-            hint=lti_config.location,
-            lti_hint="deep_linking_launch"
+            launch_data,
         )
 
         # Retrieve LTI Content Items (if any was set up)
@@ -129,22 +135,28 @@ def get_lti_1p3_launch_info(config_id=None, block=None):
         if dl_content_items.exists():
             deep_linking_content_items = [item.attributes for item in dl_content_items]
 
-    link_location = lti_config.location if lti_config.location else lti_config.config_id
+    config_id = lti_config.config_id
 
     # Return LTI launch information for end user configuration
     return {
         'client_id': lti_config.lti_1p3_client_id,
-        'keyset_url': get_lms_lti_keyset_link(link_location),
+        'keyset_url': get_lms_lti_keyset_link(config_id),
         'deployment_id': '1',
         'oidc_callback': get_lms_lti_launch_link(),
-        'token_url': get_lms_lti_access_token_link(link_location),
+        'token_url': get_lms_lti_access_token_link(config_id),
         'deep_linking_launch_url': deep_linking_launch_url,
         'deep_linking_content_items':
             json.dumps(deep_linking_content_items, indent=4) if deep_linking_content_items else None,
     }
 
 
-def get_lti_1p3_launch_start_url(config_id=None, block=None, deep_link_launch=False, dl_content_id=None, hint=""):
+def get_lti_1p3_launch_start_url(
+    launch_data,
+    config_id=None,
+    block=None,
+    deep_link_launch=False,
+    dl_content_id=None,
+):
     """
     Computes and retrieves the LTI URL that starts the OIDC flow.
     """
@@ -152,25 +164,25 @@ def get_lti_1p3_launch_start_url(config_id=None, block=None, deep_link_launch=Fa
     lti_config = _get_lti_config(config_id, block)
     lti_consumer = lti_config.get_lti_consumer()
 
-    # Change LTI hint depending on LTI launch type
-    lti_hint = ""
+    # Include a message hint in the launch_data depending on LTI launch type
     # Case 1: Performs Deep Linking configuration flow. Triggered by staff users to
     # configure tool options and select content to be presented.
     if deep_link_launch:
-        lti_hint = "deep_linking_launch"
+        launch_data.message_type = "LtiDeepLinkingRequest"
     # Case 2: Perform a LTI Launch for `ltiResourceLink` content types, since they
     # need to use the launch mechanism from the callback view.
     elif dl_content_id:
-        lti_hint = f"deep_linking_content_launch:{dl_content_id}"
+        launch_data.deep_linking_content_item_id = dl_content_id
 
     # Prepare and return OIDC flow start url
-    return lti_consumer.prepare_preflight_url(
-        hint=hint,
-        lti_hint=lti_hint
-    )
+    return lti_consumer.prepare_preflight_url(launch_data)
 
 
-def get_lti_1p3_content_url(config_id=None, block=None, hint=""):
+def get_lti_1p3_content_url(
+    launch_data,
+    config_id=None,
+    block=None,
+):
     """
     Computes and returns which URL the LTI consumer should launch to.
 
@@ -189,20 +201,20 @@ def get_lti_1p3_content_url(config_id=None, block=None, hint=""):
 
     # If there's no content items, return normal LTI launch URL
     if not content_items.count():
-        return get_lti_1p3_launch_start_url(config_id, block, hint=hint)
+        return get_lti_1p3_launch_start_url(launch_data, config_id=config_id, block=block)
 
     # If there's a single `ltiResourceLink` content, return the launch
-    # url for that specif deep link
+    # url for that specific deep link
     if content_items.count() == 1 and content_items.get().content_type == LtiDlContentItem.LTI_RESOURCE_LINK:
         return get_lti_1p3_launch_start_url(
-            config_id,
-            block,
+            launch_data,
+            config_id=config_id,
+            block=block,
             dl_content_id=content_items.get().id,
-            hint=hint,
         )
 
     # If there's more than one content item, return content presentation URL
-    return get_lti_deeplinking_content_url(lti_config.id)
+    return get_lti_deeplinking_content_url(lti_config.id, launch_data)
 
 
 def get_deep_linking_data(deep_linking_id, config_id=None, block=None):
@@ -232,3 +244,40 @@ def get_lti_pii_sharing_state_for_course(course_key: CourseKey) -> bool:
         bool: The state of PII sharing for this course for LTI.
     """
     return CourseAllowPIISharingInLTIFlag.current(course_key).enabled
+
+
+def validate_lti_1p3_launch_data(launch_data):
+    """
+    Validate that the data in Lti1p3LaunchData are valid and raise an LtiMessageHintValidationFailure exception if they
+    are not.
+
+    The initializer of the Lti1p3LaunchData takes care of ensuring that required data is provided to the class. This
+    utility method verifies that other requirements of the data are met.
+    """
+    validation_messages = []
+
+    # The context claim is an object that composes properties about the context. The claim itself is optional, but if it
+    # is provided, then the id property is required. Ensure that if any other of the other optional properties are
+    # provided that the id property is also provided.
+    if ((launch_data.context_type or launch_data.context_title or launch_data.context_label) and not
+            launch_data.context_id):
+        validation_messages.append(
+            "The context_id attribute is required in the launch data if any optional context properties are provided."
+        )
+
+    if launch_data.user_role not in LTI_1P3_ROLE_MAP:
+        validation_messages.append(f"The user_role attribute {launch_data.user_role} is not a valid user_role.")
+
+    context_type = launch_data.context_type
+    if context_type:
+        try:
+            get_lti_1p3_context_types_claim(context_type)
+        except ValueError:
+            validation_messages.append(
+                f"The context_type attribute {context_type} in the launch data is not a valid context_type."
+            )
+
+    if validation_messages:
+        return False, validation_messages
+    else:
+        return True, []
--- a/lti_consumer/lti_1p3/consumer.py
+++ b/lti_consumer/lti_1p3/consumer.py
@@ -1,7 +1,10 @@
 """
 LTI 1.3 Consumer implementation
 """
+import logging
 from urllib.parse import urlencode
+
+from lti_consumer.utils import cache_lti_1p3_launch_data, get_data_from_cache
 
 from . import constants, exceptions
 from .constants import (
@@ -15,6 +18,8 @@ from .key_handlers import ToolKeyHandler, PlatformKeyHandler
 from .ags import LtiAgs
 from .deep_linking import LtiDeepLinking
 from .nprs import LtiNrps
+
+log = logging.getLogger(__name__)
 
 
 class LtiConsumer1p3:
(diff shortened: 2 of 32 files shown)


lti_consumer/api.py
```diff
@@ -180,29 +190,65 @@ def get_lti_1p3_content_url(config_id=None, block=None, hint=""):
         b. Deep Linking is configured, but a single ltiResourceLink was selected.
     2. The Deep Linking content presentation URL if there's more than one
        Lti DL content in the database.
+
+    The launch_information parameter is a dictionary that contains data necessary and related to an LTI 1.3 launch.
```
MichaelRoytman, line 194: Once we agree on these, I'd like to put this comment in a more centralized place. Or, at least, one each API method that accepts this dictionary (e.g. `get_lti_1p3_launch_start_url` and `get_lti_1p3_launch_info`).

lti_consumer/plugin/views.py
```diff
@@ -438,10 +449,24 @@ def deep_linking_response_endpoint(request, lti_config_id=None):
 
 @require_http_methods(['GET'])
 @xframe_options_sameorigin
-def deep_linking_content_endpoint(request, lti_config_id=None):
+def deep_linking_content_endpoint(request, lti_config_id):
```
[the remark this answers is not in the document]
MichaelRoytman: This does feel a little code smelly to me to have to pass this `launch_data` parameter around so much.

lti_consumer/api.py
```diff
@@ -180,29 +190,65 @@ def get_lti_1p3_content_url(config_id=None, block=None, hint=""):
...
     """
+    # TODO: In order to clean up the Python API, we could like to remove the block parameter to this function and only
+    #       support the config_id parameter. This decouples the XBlock from the LTI launch more completely. However,
+    #       there is a difference in the way the LtiConfiguration._block attribute is set. When block is passed to
+    #       _get_lti_config, the block instance is set as LtiConfiguration._block. When config_id is passed to
+    #       _get_lti_config, the block is loaded from the modulestore via the load_block_as_anonymous_user compatibility
+    #       function. The anonymous user is bound to the block, and, due to the way the XBlock runtime works, this
+    #       affects the instance of the XBlock in the caller. More specifically, XBlock fields that have the
+    #       Scope.user_state scope raise an error after _get_lti_config is called.
+    #       See JIRA:MST-1645.
```
ashultz0, line 226: Sadly we can't remove the block load right now because the lti config is not guaranteed to have enough data to actually do LTI, it could have to look up that data in the block. There are also a couple of advantage places where the code gets a config via the ID which is the only thing it got in the request... and then uses the block out of the config because it needs block info.

I think we can actually make block load anonymous only if it has to be with something like this: https://github.com/openedx/xblock-lti-consumer/commit/1212c2f8899bfa62cb68d75e72d4cfb2dd5e5e20 which should prevent the anonymous user from polluting a request that has a real user. I haven't actually tested that yet.

lti_consumer/api.py
```diff
@@ -180,29 +190,65 @@ def get_lti_1p3_content_url(config_id=None, block=None, hint=""):
...
+    The launch_information parameter is a dictionary that contains data necessary and related to an LTI 1.3 launch.
+    It provides a mechanism for the context to provide contextually defined or stored launch information to the library.
+    The schema of launch_information may change as we support additional claims or LTI specifications.
+
+    Below are the currently support keys for launch_information.
+
+    config_id (required; string): the config_id field of an LtiConfiguration to use for the launch
+    user_role (required; string): the user's role as one of the keys in LTI_1P3_ROLE_MAP: staff, instructor, student,
+        or guest.
+    user_id (required; string, int): the user's unique identifier
```
ashultz0, line 203: specifically the LTI-facing identifier right, not our own ID?

lti_consumer/api.py
```diff
@@ -180,29 +190,65 @@ def get_lti_1p3_content_url(config_id=None, block=None, hint=""):
...
     2. The Deep Linking content presentation URL if there's more than one
        Lti DL content in the database.
+
+    The launch_information parameter is a dictionary that contains data necessary and related to an LTI 1.3 launch.
+    It provides a mechanism for the context to provide contextually defined or stored launch information to the library.
+    The schema of launch_information may change as we support additional claims or LTI specifications.
+
+    Below are the currently support keys for launch_information.
+
+    config_id (required; string): the config_id field of an LtiConfiguration to use for the launch
```
ashultz0, line 200: in most of the calls you leave config id / block id in place which makes this one redundant. From messing around on 1645 I think we probably should leave it out of the launch info and replace the config/block optional pair with just passing in a config directly.

So most calls look like lti_whatever_thing(lti_config, launch_info)

Currently everything would get the config by just doing get_config(optional id, optional block). But we can break those up later so that we can get different config satisfying objects from different paths, some bound to blocks, some not, some able to load their own blocks.

lti_consumer/lti_xblock.py
```diff
@@ -1405,6 +1414,59 @@ def _get_lti_launch_url(self, consumer):
...
+
+        return {
+            "config_id": str(lti_config.config_id),
+            "message_type": "LtiResourceLinkRequest",
+            "user_id": external_user_id,
+            "user_role": user_role,
+            "context": context_information,
+            "launch_presentation": launch_presentation_information,
+            "resource_link": resource_link_information,
+        }
```
ashultz0, line 1448: we don't really need anything other than a dictionary but if we made this a python object it would be more obvious when you put in weird stuff. And it would be a place to put the documentation in the big comment above that describes the launch structure.

lti_consumer/plugin/views.py
```diff
@@ -177,50 +175,60 @@ def launch_gate_endpoint(request, suffix=None):  # pylint: disable=unused-argume
...
-        # more specific in the future for supporting other tools like discussions, etc.
-        lti_consumer.set_context_claim(
-            str(course_key),
-            context_types=[LTI_1P3_CONTEXT_TYPE.course_offering],
-            context_title=context_title,
-            context_label=str(course_key)
+        lti_consumer = lti_config.get_lti_consumer()
+        # Decode the JWT and verify the signature before proceeding to verify that the lti_message_hint
+        # JWT came from the platform.
+        lti_message_hint_data = lti_consumer.decode_token(lti_message_hint)
```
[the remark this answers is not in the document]
ashultz0: I think technically the JWT key handling stuff knows which part of the private key to use when verifying the signature because they JWT carries a key ID field that says "signed with this key". We have both the public and private parts of the key available so when the JWT library knows it was signed with key foo (private side) it knows to verify with key foo (public side)

lti_consumer/api.py
```diff
@@ -100,7 +100,11 @@ def get_lti_consumer(config_id=None, block=None):
     return _get_lti_config(config_id, block).get_lti_consumer()
 
 
-def get_lti_1p3_launch_info(config_id=None, block=None):
+def get_lti_1p3_launch_info(
```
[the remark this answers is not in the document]
ashultz0: or at least don't version and release the library without both of them

in practice is anyone calling these APIs yet outside of the xblock which is also in here?

lti_consumer/api.py
```diff
@@ -180,29 +190,65 @@ def get_lti_1p3_content_url(config_id=None, block=None, hint=""):
...
+    It provides a mechanism for the context to provide contextually defined or stored launch information to the library.
+    The schema of launch_data may change as we support additional claims or LTI specifications.
+
+    Below are the currently support keys for launch_data.
+
+    config_id (required; string): the config_id field of an LtiConfiguration to use for the launch
+    user_role (required; string): the user's role as one of the keys in LTI_1P3_ROLE_MAP: staff, instructor, student,
+        or guest.
+    user_id (required; string, int): the user's unique identifier
+    resource_link_id (required; string): a unique identifier that is guaranteed to be unique within the platform for
```
giovannicimolin, line 204: @MichaelRoytman Just a note here: `resource_link_id` should be unique **per placement**, not per platform. This means that each place that has an LTI 1.3 integration needs to have a unique ID - in the previous implementation this was easily solvable by using the XBlock ID, but I'm not sure how are we going to do this now. My guess is that for XBlock launches, we should keep using the XBlock ID (it also makes linking grades easier). What do you think?

giovannicimolin:
@MichaelRoytman Is this branch working on the devstack? I can do a test run if so. :grin:

lti_consumer/lti_xblock.py
Zacharis278: +1 for making this an object. We could even use that class to hide away how we encode/decode/validate this which might help keep the consumer class cleaner.

lti_consumer/api.py
MichaelRoytman: Thank you for the clarification! I'll update the comment.

Are you comfortable with allowing the context to supply the `resource_link_id`? I ask because you mentioned having an `LtiPlacement` model in your [feedback](https://github.com/openedx/xblock-lti-consumer/pull/281/files#r978036669) in the draft ADR. I assume that the library would get the `resource_link_id` from the `LtiPlacement` model instead of getting it as a part of `launch_data`.

lti_consumer/api.py
ashultz0: from discussion in slack the config ID in the launch data is probably the one to keep, I'll look into removing the other one while trying to remove the block argument in my own refactor

MichaelRoytman:
@giovannicimolin I have tested XBlock and database based basic LTI 1.3 launches with success. I am trying to test deep linking using the 1EdTech test suite, but I cannot get it to work. I keep going in circles and getting errors in the test suite. If you're able to, can you show me how you test deep linking?

Also, this is ready for a rereview!

lti_consumer/api.py
schenedx: Another #NIT thing. Our team's MST jira board is not publicly visible. If you want to reference an MST ticket, you'd have to think about how to expose the content of that ticket to the open source readers of this code. The easiest way is don't reference the ticket, but to copy and paste the content of the ticket to a github issue, then reference github issue. But you'd have to manage the state of that github issue.

schenedx (commented):
Shouldn't this change result into a major version bump on the library? Is that being taken care of by the commit message with `feat` prefix?

lti_consumer/tests/test_utils.py
```diff
@@ -0,0 +1,96 @@
+"""
```
MichaelRoytman, line 1: This was a move from `lti_consumer/tests/unit/test_utils.py` to `lti_consumer/test/test_utils.py`. This was to accommodate a file in `lti_consumer/test/unit` named `test_utils.py`, which contains actual tests of this module of testing utilities. This should be displayed as a renamed file, but it's not.

lti_consumer/lti_1p3/consumer.py
```diff
@@ -90,20 +93,30 @@ def _get_user_roles(role):
...
-            hint="hint",
-            lti_hint="lti_hint"
+            launch_data,
     ):
         """
         Generates OIDC url with parameters
         """
+        user_id = launch_data.user_id
+
+        # Set the launch_data in the cache. We store the launch_data in the cache instead of sending a representation of
```
Zacharis278, line 103: kind of a NIT? IMO I don't feel there's a lot of value in explaining a solution we don't implement with the query param. I think it would be more useful to just explain why we store this (to reference later) instead.

lti_consumer/lti_1p3/consumer.py
```diff
@@ -90,20 +93,30 @@ def _get_user_roles(role):
...
+        # the launch_data (e.g. a JWT) as a query or form parameter because we cannot guarantee that the length of the
+        # GET query parameter will not exceed some browser maximum length and we need to support GET and POST requests
+        # back from tools.
+        # TODO: What is the expiration value by default? Will it provide enough time for the learner to get back?
```
Zacharis278, line 107: I think we should explicitly set this instead of relying on some internal default just so it's more visible. We can set this pretty short if we wanted but even 60s is a lot of buffer for an auth callback.

lti_consumer/lti_xblock.py
```diff
@@ -1405,6 +1416,58 @@ def _get_lti_launch_url(self, consumer):
...
+
+        launch_presentation_information = {
+            "document_target": "iframe",
+        }
+
+        # Runtime import since this will only run in the
+        # Open edX LMS/Studio environments.
+        # pylint: disable=import-outside-toplevel
+        from lti_consumer.api import _get_lti_config
```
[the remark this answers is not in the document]
Zacharis278: For the sake of accounting for 'things we need to fix' would it be worth putting a comment here?

lti_consumer/lti_1p3/consumer.py
MichaelRoytman: That's fair. I'm not concerned as much about the authentication callback for learners on the LMS where the LTI launch is in an iframe, because the LTI launch occurs automatically after the page is rendered.

However, because the launch data is cached when the XBlock renders and the preflight URL is generated, this poses a problem for delayed LTI launches. This includes learner LMS LTI launches that are configured to launch in a modal or new window and Studio deep linking LTI launches for course staff (initiated by clicking a URL). I'm concerned that the data will not be cached long enough these LTI launches if we choose a 60 second timeout. Course staff or learners may leave pages open for longer than this and reasonably expect them to be operational.

The default timeout is 5 minutes. This is the same timeout for the JWT ID token. I don't see a harm in caching these values for this long to make launches more user friendly. If you'd like, I can include `django.core.cache.backends.base.DEFAULT_TIMEOUT` as an an argument to `TieredCache.set_all_tiers` to make clear the timeout.

What do you think?

I'm also thinking about providing more specific instructions on the error page in cases where the cache has expired, as opposed to our generic, "contact the course team" message. If it's as simple as reloading the LMS or Studio page, we should instruct them to do so instead of sending them to support.

Zacharis278: oh interesting I hadn't considered the fact that some of these preflight url's just hang around before getting consumed. Maybe that's something we could improve on (but later) I'd think a simple message to reload would be sufficient.

In any case I also don't see any issues with holding these for 5mins given the scale of use on edx.org (assuming all of that moved to 1.3 some day). But, why not just define our own timeout value to pass in instead, or is that default a platform setting we want to be configurable?

Zacharis278:
I think this looks great. I also ran through a quick manual test with the basic launch with no issues. I'm gonna take another look at this next week with fresh eyes since there's a few spots I skipped over quickly.

lti_consumer/lti_1p3/consumer.py
MichaelRoytman: I didn't see a need to specify a value that's also the default, but I understand that it'll be more readable to explicitly specify the timeout we want. I'll add that in. 

I'm going to add in some logging to capture cache misses. It will be good to keep an eye on how frequently those occur in case 5 minutes is not long enough.

lti_consumer/api.py
MichaelRoytman: It can be whatever identifier the user of the library will use to uniquely identify the user. Ideally, we wouldn't use our internal database identifier, yes, but each user may define that value independently.

Zacharis278 (approved):
let's ship this! 🎉

ashultz0 (approved):

PR #294, opened 2022-10-19
API gets config from launch data via config_id

ashultz0:
Includes an improvement to anonymous loading because we're going to be using block load all the time for configs stored on the xblock. Basically if the block has already been loaded and bound to a user be happy and leave it. And if the block needs to be loaded and you have a user, load as that user. Only fall through to anonymous user when you have no other option (for example when being called by the LTI provider directly which is not a user).

There are a few cleanup diffs, the only interesting one is that codecov complained about compat so I added pragmas to document our lack of compat tests. The amount of mocking required would make those basically worthless.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -15,13 +15,25 @@ Please See the [releases tab](https://github.com/edx/xblock-lti-consumer/release
 
 Unreleased
 ~~~~~~~~~~
+6.0.0 - 2022-10-24
+------------------
+BREAKING CHANGE:
+
+Please note that additional breaking changes will be forthcoming in future versions of this library.
+
+* Modified Python API methods to use config_id (the UUID field) exclusively rather than config.id or block.
+
+  * For the functions changed in 5.0.0 the config_id is available in the launch_data.
+  * Other functions had config.id changed to config_id and block removed as an argument.
+  * The new function config_id_for_block gets that config ID if all you have is a block.
+
 5.0.1 - 2022-10-17
 ------------------
 * Fixed a bug that prevented LTI 1.3 launches from occurring in the browser due to Django's clickjacking protection.
 
   * Added the xframe_options_exempt view decorator to launch_gate_endpoint to allow loading response in an <iframe> tags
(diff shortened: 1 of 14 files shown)


ashultz0:
Intial 1.3 smoke testing is successful. I put some logging into the load_as_user function to see who it is being on its way through the system

on block load it goes through get_lti_1p3_launch_start_url out of the template and recognizes the user so it does not reload the block

after poking GET on the LTI testing tool it makes two requests that load as anonymous user and succeeds: launch_gate_endpoint and access_token_endpoint

giovannicimolin:
@ashultz0 Nice work here so far! The rework looks good.

>after poking GET on the LTI testing tool it makes two requests that load as anonymous user and succeeds: launch_gate_endpoint and access_token_endpoint

I did not expect this, I thought that after #288 we wouldn't need to load the block since the information would be stored on the model? 

Or this is a step that we still need to do as an outcome of ADR https://github.com/openedx/xblock-lti-consumer/pull/281?

giovannicimolin:
@ashultz0 Got it! So there are still some steps to be done to get rid of block dependencies - thanks for the clarification.

Let me know when you need a review/test here, but so far looks like a good cleanup. :grin:

ashultz0:
I repeated my 1.3 launch test and as expected when the LTI configuration is set to CONFIG_IN_DB it does not have to load the block.

ashultz0:
@giovannicimolin I got through my LTI testing as config_on_db to validate that it does what I expect and some other cleanup, so it's ready for final review :)

Zacharis278 (approved):
Hurray for making this just a bit more simple to understand

giovannicimolin (approved):
:+1:

- [x] I tested this:
1. Using LTI 1.3 launches for the IMS reference tool as well as to H5P. :heavy_check_mark: 
2. Using LTI 1.1 launching the SaLTIre test tool (https://saltire.lti.app/tool). :heavy_check_mark: 
- [x] I read through the code
- [ ] I checked for accessibility issues
- [x] Includes documentation

@ashultz0 Nice work! I done some basic testing with LTI 1.1 and 1.3 launches.

PR #297, opened 2022-10-25
Proctoring Service Implementation In-Browser Proctoring Launch

MichaelRoytman:
**Description:**

This pull request introduces support for the LTI 1.3 Proctoring Services specification. It includes support for the in-browser proctoring launch flow. It does not currently include support for the proctoring services.

The previous iteration of these changes can be found [here](https://github.com/openedx/xblock-lti-consumer/pull/297). This pull request takes its place.

**Changes:**

Here is a digest of important changes.

* An `Lti1p3ProctoringLaunchData` data class is introduced. It should be included as an attribute of the `Lti1p3LaunchData` data class to provide necessary proctoring data for a launch.
* An `LtiProctoringConsumer` class is introduced. This class is used to generate LTI proctoring launch requests and to decode and validate the JWT send back by the Tool with the `LtiStartAssessment` message.
* An `lti_1p3_proctoring_enabled` BooleanField is added to the `LtiConfiguration` model. This field controls whether proctoring is enabled for a particular LTI integration.
* The `launch_gate_endpoint` has been modified to support `LtiStartProctoring` and `LtiEndAssessment` LTI launch messages.
* A `start_proctoring_assessment_endpoint` is added to support `LtiStartAssessment` messages from the Tool.
* An `LTI_1P3_PROCTORING_ASSESSMENT_STARTED` signal is introduced. This signal is emitted when the `LtiStartAssessment` message is sent from the Tool to inform users of the library that the `LtiStartAssessment` message has been received.

**Testing Instructions:**

You will need a user of this new code. You can the edx-exams repository to do so. You can use this pull request ().

* In your local checkout of edx-exams, set the following settings in `edx_exams/settings/private.py`.

  * ROOT_URL = <NGROK_URL>
  * LMS_ROOT_URL = <NGROK_URL>
  * SESSION_COOKIE_SAMESITE = "None"
  * SESSION_COOKIE_SECURE = True
  * ALLOWED_HOSTS = ['*']

* The requirements files should be pointing at the correct branch of xblock-lti-consumer and should install it correctly. If there are problems, check out the xblock-lti-consumer library locally and pip install it. If the proposed Proctoring Services implementation is still on a branch, check out the branch an install it.

* Log in to the edx-exams service using the ngrok URL - `<NGROK_URL>/admin`. The reason it's important to log in via the ngrok URL is to ensure that the session cookie is set on the proper domain so that the cookie is included by the browser when the Tool makes requests to the NGROK_URL. If you log in through localhost, then when the Tool makes requests to the NGROK_URL, the user will be the AnonymousUser. This will cause the flow to fail.
* Set up the [Proctoring Certification Suite 1.0](https://certification.imsglobal.org/certification/proctoring/launch.html).
  * Create a new `LtiConfiguration` instance in the edx-exams admin panel.

    * Set the "Config store" to "Configuration Stored on this model".
  
  * The process of setting this testing suite up is very similar to the [existing instructions](https://github.com/openedx/xblock-lti-consumer#lti-13) for testing LTI 1.3 launches in the xblock-lti-consumer README. Use that as a reference. Instead of setting the data on the XBlock, set it on the LtiConfiguration in the admin panel. Instead of setting the data in the LTI 1.3 launch testing tool, set it in the Proctoring Certification Suite 1.0.

* In your browser, hit the following URL to start the LTI proctoring flow: `<NGROK_URL>/lti/start_proctoring/1`. 1 represents the database ID for the `LtiConfiguration`, so set it appropriately if you have more than one `LtiConfiguration` you're testing with.

**Quality Assurance:**

This code has been tested manually with a `CONFIG_EXTERNAL` configuration using the aforementioned edx-exams code.

Here are screenshots from the Proctoring Services test suite published by 1EdTech.

Successful LTI Launch with "LtiStartProctoring" message:

image

There are a number of tests under the "Start Assessment" section. They're shown below.

image

All tests intended to test failures show this page.

image

The last test, "Valid LTI Message", shows this page.

image


lti_consumer/plugin/views.py
```diff
@@ -704,3 +756,104 @@ def list(self, *args, **kwargs):
...
+    iss = jwt_payload.get('iss')
+
+    try:
+        lti_config = LtiConfiguration.objects.get(lti_1p3_client_id=iss)
+    except LtiConfiguration.DoesNotExist as exc:
+        log.error("Invalid iss claim '%s' for LTI 1.3 Proctoring Services start_proctoring_assessment_endpoint"
+                  " callback", iss)
+        raise Http404 from exc
+
+    if not lti_config.lti_1p3_proctoring_enabled:
```
MichaelRoytman, line 795: I _think_ we're intended to interface through the consumer instead of accessing these fields directly (hence the `dl`, `nrps`, and `ags` attributes of the `LtiAdvantageConsumer`). I could use the `proctoring_data` attribute (doesn't roundabout) or check `isinstance(LtiProctoringConsumer, lti_consumer`).

lti_consumer/plugin/views.py
```diff
@@ -704,3 +749,102 @@ def list(self, *args, **kwargs):
...
+    # the Assessment Platform MUST send an End Assessment Message at the end of the user's
+    # proctored exam.
+    end_assessment_return = proctoring_response.get('end_assessment_return')
+    if end_assessment_return:
+        end_assessment_return_key = get_cache_key(app="lti", key="end_assessment_return", user_id=request.user.id)
+        # We convert the boolean to an int because memcached will return an int even if a boolean is stored. This
+        # ensures a consistent return value.
+        end_assessment_return_value = int(end_assessment_return)
+
+        TieredCache.set_all_tiers(end_assessment_return_key, end_assessment_return_value, django_cache_timeout=36000)
```
MichaelRoytman, line 841: 10 hours is a little dramatic. However, I want to be sure that this data lives long enough that the learner can complete their exam.

We may want to store this data in `edx-exams` instead for each attempt, if this feels hacky.

lti_consumer/plugin/views.py
```diff
@@ -277,7 +329,7 @@ def launch_gate_endpoint(request, suffix=None):  # pylint: disable=unused-argume
             'user_roles': user_role,
             'launch_url': context['launch_url']
         }
-        track_event('xblock.launch_request', event)
+        track_event('launch_request', event)
```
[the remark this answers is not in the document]
Zacharis278: actually we own this and the reason to have xblock here is to differentiate xblock launches from other types. Soooo this tracking code might need to be moved elsewhere now that the assumption this code is only ever run as an xblock launch is no longer true.

lti_consumer/lti_1p3/consumer.py
```diff
@@ -704,3 +708,174 @@ def enable_nrps(self, context_memberships_url):
...
+        does not validate their values. Use validate_token_contents to validate values. Returns a dictionary
+        representation of key proctoring claims in the Tool JWT token.
+
+        Arguments:
+            * token (string): a JWT
+        """
+        # Decode token and check expiration.
+        proctoring_response = self.tool_jwt.validate_and_decode(token)
+
+        claim_key = "https://purl.imsglobal.org/spec/lti-ap/claim/session_data"
```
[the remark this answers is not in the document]
ashultz0: maybe a separate claim validation method? so you don't have to mix together "this is a signed token" and "this token is not expired" with "this token has 16 special claims"

lti_consumer/lti_xblock.py
```diff
@@ -786,6 +786,18 @@ def lti_provider_key_secret(self):
...
+        """
+        user_id = self.runtime.service(self, 'user').get_current_user().opt_attrs.get(
+            'edx-platform.user_id', None)
+
+        if user_id is None:
+            raise LtiError(self.ugettext("Could not get user id for current request"))
+        return user_id
+
+    @property
+    def anonymous_user_id(self):
```
ashultz0, line 800: this is nice and specific :)

lti_consumer/plugin/views.py
```diff
@@ -233,14 +249,20 @@ def launch_gate_endpoint(request, suffix=None):  # pylint: disable=unused-argume
...
+                                                    'The required parameter redirect_uri is missing.')
+
+        context.update({'launch_url': redirect_uri})
         # Modify LTI Launch URL depending on launch type.
         # Deep Linking Launch - Configuration flow launched by
         # course creators to set up content.
         deep_linking_content_item_id = launch_data.deep_linking_content_item_id
 
-        if lti_consumer.dl and launch_data.message_type == 'LtiDeepLinkingRequest':
+        if getattr(lti_consumer, 'dl', None) and launch_data.message_type == 'LtiDeepLinkingRequest':
```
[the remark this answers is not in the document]
ashultz0: could we just switch the order so that it doesn't try to dl unless it's that message type? and the same below

lti_consumer/plugin/views.py
ashultz0: it's probably fine. A nice trick for timeouts like this is to write it as 60*60*10 so the 10 hours is immediately obvious

lti_consumer/lti_1p3/consumer.py
```diff
@@ -704,3 +708,178 @@ def enable_nrps(self, context_memberships_url):
...
+    ):
+        """
+        Builds and return LTI launch message for proctoring.
+
+        Overrides LtiConsumer1p3's method to include proctoring specific launch claims. Leverages
+        the set_extra_claim method to include these additional claims in the LTI launch message.
+        """
+        lti_message_hint = preflight_response.get('lti_message_hint')
+        launch_data = get_data_from_cache(lti_message_hint)
+
```
ashultz0, line 814: codecov is right, this seems pretty testable and good to test :)

lti_consumer/lti_1p3/consumer.py
```diff
@@ -704,3 +708,178 @@ def enable_nrps(self, context_memberships_url):
...
+        proctoring_response = self.tool_jwt.validate_and_decode(token)
+
+        claim_key = "https://purl.imsglobal.org/spec/lti-ap/claim/session_data"
+        session_data = proctoring_response.get(claim_key)
+        if session_data is None:
+            raise MissingRequiredClaim(f"Token is missing required {claim_key} claim.")
+
+        claim_key = "https://purl.imsglobal.org/spec/lti/claim/resource_link"
+        resource_link = proctoring_response.get(claim_key)
+        if resource_link is None or not resource_link:
```
ashultz0, line 845: can collapse to "if not resource_link" since None is falsey

lti_consumer/lti_1p3/consumer.py
```diff
@@ -193,25 +195,27 @@ def set_resource_link_claim(
...
+            document_target="iframe",
+            return_url=None,
     ):
         """
         Optional: Set launch presentation claims
 
         http://www.imsglobal.org/spec/lti/v1p3/#launch-presentation-claim
         """
-        if document_target not in ['iframe', 'frame', 'window']:
+        if document_target and document_target not in ['iframe', 'frame', 'window']:
```
ashultz0, line 206: this means it's ok to have document_target = None or ''?

lti_consumer/data.py
```diff
@@ -13,34 +36,50 @@ class Lti1p3LaunchData:
     launch data between apps. Applications using this library should use the Lti1p3LaunchData class to supply
     contextually defined or stored launch data to the generic LTI 1.3 launch.
 
-    * user_id (required): the user's unique identifier
-    * user_role (required): the user's role as one of the keys in LTI_1P3_ROLE_MAP: staff, instructor, student, or
+    * user_id (required): A unique identifier for the user that is requesting the LTI 1.3 launch. If the optional
```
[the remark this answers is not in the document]
varshamenon4: @MichaelRoytman So to clarify, the `user_id` that you've removed and the one you've added are different, right?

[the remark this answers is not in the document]
varshamenon4: Actually, I think your comment below answers my question!

> See my comment about `user_id` in data.py. I need the actual user database identifier, so I had to rename `user_id` to `anonymous_user_id` and then implement `user_id` to get the database identifier.

lti_consumer/api.py
```diff
@@ -277,7 +279,33 @@ def validate_lti_1p3_launch_data(launch_data):
...
     else:
         return True, []
+
+
+def get_end_assessment_return(user_id):
+    """
+    Returns the end_assessment_return value stored in the cache. This can be used by applications to determine whether
+    to invoke an LtiEndAssessment LTI launch.
+    """
+    end_assessment_return_key = get_cache_key(app="lti", key="end_assessment_return", user_id=user_id)
```
varshamenon4, line 308: Nice! In general, I really like how descriptive, yet concise your variable/method names are.

lti_consumer/data.py
```diff
@@ -3,7 +3,30 @@
...
+
+    * attempt_number (required): The number of the user's attempt in the assessment. The attempt number
+        should be an integer that starts with 1 and that is incremented by 1 for each of user's subsequent attempts at
+        the assessment.
+    * start_assessment_url (conditionally required): The Platform URL that the Tool will send the start
+        assessment message to after it has completed the proctoring setup and verification. This attribute is required
+        if the message_type attribute of the Lti1p3LaunchData instance is "LtiStartProctoring". It is optional and
+        unused otherwise.
+    """
+    attempt_number = field()
```
varshamenon4, line 28: TIL about `@define` and `field()` usage... this is neat!

lti_consumer/lti_1p3/consumer.py
varshamenon4: I think the first condition (`if document_target`) should check that document_target is not None right?

varshamenon4: Oh wait I think I understand what you're saying @ashultz0. I guess, should there be a check for `if not document_target`?

lti_consumer/lti_1p3/consumer.py
```diff
@@ -704,3 +708,178 @@ def enable_nrps(self, context_memberships_url):
...
+            if key in LTI_PROCTORING_DATA_KEYS:
+                self.proctoring_data[key] = value
+
+    def _get_base_claims(self):
+        """
+        Returns claims common to all LTI Proctoring Services LTI launch messages, to be used when creating LTI launch
+        messages.
+        """
+        proctoring_claims = {
+            "https://purl.imsglobal.org/spec/lti-ap/claim/attempt_number": self.proctoring_data.get("attempt_number"),
```
varshamenon4, line 770: Question: so does the url show the value (i.e. attempt number)?

lti_consumer/lti_1p3/consumer.py
```diff
@@ -704,3 +708,178 @@ def enable_nrps(self, context_memberships_url):
...
+        })
+
+        return proctoring_claims
+
+    def generate_launch_request(
+        self,
+        preflight_response,
+    ):
+        """
+        Builds and return LTI launch message for proctoring.
```
varshamenon4, line 807: Question: Pretty basic, but could you clarify what the LTI launch message is for?

lti_consumer/lti_xblock.py
varshamenon4: +1

lti_consumer/plugin/views.py
```diff
@@ -704,3 +756,104 @@ def list(self, *args, **kwargs):
...
+@require_http_methods(['POST'])
+def start_proctoring_assessment_endpoint(request):
+    """
+    Receives the Proctoring Tool's message to start the assessment. Emits a signal informing interested parties that
+    the assessment should be started.
+
+    Form Parameters:
+    * JWT (REQUIRED): a signed JWT containing the LTI message
+    """
+    # DECISION: Instead of relying on a config_id URL parameter, let's use the iss claim in signed JWT sent by the
```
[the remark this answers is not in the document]
varshamenon4: Nice call!

varshamenon4 (commented):
Looks great overall! Just some questions and comments. I really appreciate how you document things and also how you clearly name things. Nice!

lti_consumer/data.py
```diff
@@ -13,34 +36,51 @@ class Lti1p3LaunchData:
...
+    * context_id (conditionally required): The id property of the context claim. It is the stable, unique identifier for
+        the context. It is required if any of the context properties are provided.
+    * context_type (optional): The type property of the context claim. It is a list of some combination of the following
+        valid context_types: group, course_offering, course_section, or course_template.
+    * context_title (optional): The title proerty of the context claim. It is a short, descriptive name for the context.
+    * context_label (optional): The label property of the context claim. It is a full, descriptive name for the context.
+    * deep_linking_context_item_id (optional): The database id of the LtiDlContentItem that should be used for the LTI
+        1.3 Deep Linking launch. It is used when the LTI 1.3 Deep Linking launch is a regular LTI resource link
+        request using a content item that was configured via a previous LTI 1.3 Deep Linking request.
+    * Lti1p3ProctoringLaunchData (conditionally required): An instance of the Lti1p3ProctoringLaunchData that contains
```
Zacharis278, line 66: nit: I think you mean to put the key `proctoring_launch_data` here right?

lti_consumer/data.py
```diff
@@ -13,34 +36,51 @@ class Lti1p3LaunchData:
...
+        data necessary and related to an LTI 1.3 proctoring launch. It is required if the message_type attribute is
+        "LtiStartProctoring" or "LtiEndAssessment".
     """
     user_id = field()
     user_role = field()
     config_id = field()
     resource_link_id = field()
+    external_user_id = field(default=None)
     launch_presentation_document_target = field(default="iframe")
+    launch_presentation_return_url = field(default=None)
```
Zacharis278, line 76: looks like this param is missing from the docstring

lti_consumer/lti_1p3/consumer.py
```diff
@@ -704,3 +708,178 @@ def enable_nrps(self, context_memberships_url):
...
+        """
+        Validates the contents of a Tool JWT token. Raises InvalidClaimValue appropriately.
+
+        Arguments:
+            * token_contents (dict): a dictionary representation of the Tool JWT claims, as returned by
+              check_and_decode_token.
+        """
+        # TODO: We only HAVE to validate the session_data claim. Other claims must be validated, but they should be
+        #       validated by the ToolKeyHandler class.
+        # TODO: An Assessment Platform MAY reject the Start Assessment message if a required identity claim is missing
```
Zacharis278, line 870: this sounds like something we can revisit once we start to look at integration with name affirmation / verification. It may make sense to be less restrictive at the start.

lti_consumer/plugin/views.py
Zacharis278: no strong opinions here but if we need to build the consumer anyway below I suppose it would be cleaner to just check the type of that consumer object.

lti_consumer/models.py
```diff
@@ -487,8 +500,13 @@ def _get_lti_1p3_consumer(self):
         Uses the `config_store` variable to determine where to
         look for the configuration and instance the class.
         """
+        consumer_class = LtiAdvantageConsumer
+        # LTI Proctoring Services is not currently supported for CONFIG_ON_XBLOCK or CONFIG_EXTERNAL.
+        if self.lti_1p3_proctoring_enabled and self.config_store == self.CONFIG_ON_DB:
+            consumer_class = LtiProctoringConsumer
```
giovannicimolin, line 506: @MichaelRoytman These lines mean that if `lti_1p3_proctoring_enabled` is true, then none of the advantage services will be available to that consumer, right? If we're working with a globally re-usable configuration then this adds a constraint: proctoring configs and other types of LTI configs must always be separate.

We don't have a use case for this right now, so this shouldn't block this ticket, but can you add a comment mentioning this new constraint with some context?

giovannicimolin: _I have one idea in mind_: instead of using inheritance in the `LtiConsumer`, we could use composition, and add the services incrementally to the main class only if they are enabled in the configuration.

lti_consumer/plugin/views.py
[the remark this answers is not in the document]
giovannicimolin: Yup, @ashultz0 suggestions seems simpler.

Also, with these complications arising, I'd be in favor of switching from inheritance to composition in the `LtiConsumer` classes - only configuring the LTI services if they are in the LTI config and throwing errors if the clients try to launch invalid LTI setups.

giovannicimolin (commented):
@MichaelRoytman I don't have access to IMS's certification suite, so I'll leave the testing to you and @Zacharis278. 

I did a review pass and left some comments. Something that bothers me a bit is the added complexity that having two different consumer classes is added. Maybe we should rework this and work with composition, only setting up each service if it is defined in the LTI configuration?

I don't think this needs to be added here since it will require some rework, but worth keeping in mind and discussing on a GH issue.

Zacharis278:
@giovannicimolin I'm liking the suggestion not to have multiple child classes of LtiConsumer. I do think this would be a good candidate for a GH issue.

Zacharis278:
Overall comment after reviewing both this and https://github.com/edx/edx-exams/pull/32. This looks great, is the next step just to add some further unit testing around the new consumer type?

MichaelRoytman (commented):
Just leaving some responses to feedback...

MichaelRoytman:
@giovannicimolin I created https://github.com/openedx/xblock-lti-consumer/issues/303 to track the LTI consumer class refactoring.

giovannicimolin (approved):
:+1:

- [x] I tested this: (I tested LMS for side-effects of this change) 
1. Done some limited testing to ensure functionality continues working on LMS and Studio.
2. Checked H5P deep-linked launch and a normal launch using IMS-RI. :heavy_check_mark: 
image
image
4.  Checked that the IMS certification tests passed (screenshot on PR) :heavy_check_mark: 
- [x] I read through the code and left comments
- [ ] Includes documentation: not yet :warning: 

@MichaelRoytman Looks good, and it doesn't seem to affect normal launches! :grin: 
Left a comment about the caching mechanism, but it's not a blocker for this PR.

lti_consumer/__init__.py
```diff
@@ -4,4 +4,4 @@
 from .apps import LTIConsumerApp
 from .lti_xblock import LtiConsumerXBlock
 
-__version__ = '6.0.0'
+__version__ = '6.1.0'
```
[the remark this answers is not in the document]
ashultz0: yeah I just made things a little more difficult but checking the conflicts they're quite minor

lti_consumer/api.py
```diff
@@ -278,7 +280,33 @@ def validate_lti_1p3_launch_data(launch_data):
                 f"The context_type attribute {context_type} in the launch data is not a valid context_type."
             )
 
+    proctoring_launch_data = launch_data.proctoring_launch_data
+    if (launch_data.message_type in ["LtiStartProctoring", "LtiEndAssessment"] and not
+            launch_data.proctoring_launch_data):
```
ashultz0, line 285: you clearly meant to also make this just proctoring_launch_data for brevity and missed it

lti_consumer/plugin/views.py
```diff
@@ -139,13 +120,18 @@ def public_keyset_endpoint(request, usage_id=None, lti_config_id=None):
...
+    Launch Flow, respectively. The authentication response contains the LTI message and is the LTI launch.
+
+    Returns a response containing an auto-submitting form that directs the browser to make a POST to the Tool.
 
-    This uses the config_id key of the "lti_message_hint" query parameter
-    to identify the LtiConfiguration and its consumer to generate the
-    LTI 1.3 Launch Form.
+    Query Parameters:
+    * lti_message_hint (REQUIRED): a value used as a cache key to retrieved a cached instance of Lti1p3LaunchData
+    * login_hint (REQUIRED): an identifier for the user that initiated the launch; it is stable and unique to the issuer
```
ashultz0, line 134: thanks for this

lti_consumer/api.py
```diff
@@ -278,7 +280,33 @@ def validate_lti_1p3_launch_data(launch_data):
...
     else:
         return True, []
+
+
+def get_end_assessment_return(user_id):
+    """
+    Returns the end_assessment_return value stored in the cache. This can be used by applications to determine whether
+    to invoke an LtiEndAssessment LTI launch.
+    """
+    end_assessment_return_key = get_cache_key(app="lti", key="end_assessment_return", user_id=user_id)
```
ashultz0, line 309: does this key mean that a user can only end one assessment at a time? Admittedly a ridiculous corner case.

ashultz0 (approved):
I found a little minor stuff you might want to change but overall 👍 
let's ship it and iterate rather than further worrying at it

lti_consumer/api.py
MichaelRoytman: Thank you for catching this! It's not intended. I meant to include the `resource_link_id` in the cache key, which will avoid this corner case.

PR #300, opened 2022-11-04
docs: emphasize that for dev the xblock must be installed in two containers

ashultz0:


MichaelRoytman (approved):
Thank you for improving the documentation!

schenedx (approved):
👍

PR #301, opened 2022-11-07
fix: remove general use of load_block_as_user

ashultz0:
For the config model we do not need to go as far as binding the block to the user and already get enough data out of the modulestore to satisfy the storage on the xblock case. Add a new function to get that much xblock only.

For the limited cases where we are using the block more directly as a block we maintain the old function.

Also includes a fix to test_views that was closing the wrong level mock and leaving an open patch into other tests.

The behavior we are fixing is described in https://2u-internal.atlassian.net/browse/MST-1697 - basically, the weird compat.load_block_as_user squirrels its way into LMS specific block loading and binding code and that does not work in Studio. Before version 6, we got away with it because we mostly asked for the config by saying i_can_has_config(block) with a preloaded block and that would be set into the config so you would never try to load it. Version 6 made the only way to get the config i_can_has_config(config_id) which works fine but then config.block triggers the offending code.

Rather than fix config.block as we try to move away from being tightly tied to the block, replace it with a less complete load for the config-on-DB data store case and greatly limit the use of the fancier version.

### DIFF
--- a/lti_consumer/models.py
+++ b/lti_consumer/models.py
@@ -227,17 +227,14 @@ class LtiConfiguration(models.Model):
                   'grades.'
     )
 
-    # Empty variable that'll hold the block once it's retrieved
-    # from the modulestore or preloaded
-    _block = None
-
     def clean(self):
         if self.config_store == self.CONFIG_ON_XBLOCK and self.location is None:
             raise ValidationError({
                 "config_store": _("LTI Configuration stores on XBlock needs a block location set."),
             })
         if self.version == self.LTI_1P3 and self.config_store == self.CONFIG_ON_DB:
-            if not database_config_enabled(self.block.location.course_key):
+            block = compat.load_enough_xblock(self.location)
+            if not database_config_enabled(block.location.course_key):
                 raise ValidationError({
                     "config_store": _("LTI Configuration stores on database is not enabled."),
                 })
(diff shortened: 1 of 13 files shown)


ashultz0:
some covered code was removed so a slight reduction in covered lines is expected

ashultz0:
removed some covered code so a slight reduction in coverage is expected and ok

tested basic 1.1 and 1.3 using ngrok and LTI testing tools, I'm going to give it another round of testing tomorrow

schenedx (approved):
LGTM

lti_consumer/models.py
```diff
@@ -227,17 +227,14 @@ class LtiConfiguration(models.Model):
...
-    _block = None
-
     def clean(self):
         if self.config_store == self.CONFIG_ON_XBLOCK and self.location is None:
             raise ValidationError({
                 "config_store": _("LTI Configuration stores on XBlock needs a block location set."),
             })
         if self.version == self.LTI_1P3 and self.config_store == self.CONFIG_ON_DB:
-            if not database_config_enabled(self.block.location.course_key):
+            block = compat.load_enough_xblock(self.location)
```
Zacharis278, line 236: @ashultz0 so this line will cause an exception if this is run outside of edx-platform like in exams. I think maybe a try/catch would be enough to fix it since we don't care about the xmodule import in compat failing outside platform.

ashultz0: @Zacharis278 actually I'm realizing that the only thing it does with the block here is load it by the location and then check the block's location's course key - I should be able to just check the location's course key without loading the block right?

PR #302, opened 2022-11-08
299 validate redirect uris

geoff-va:
* Adds `lti_1p3_redirect_uris` to XBlock
* Adds `lti_1p3_redirect_uris` to LTIConfiguration model
* Asserts `redirect_uri` from preflight response in valid redirect uris during preflight validation
* To minimize disruption for previous versions, uses current launch url and deep linking url as default registered `redirect_uris` if `redirect_uris` is not explicitly set


geoff-va:
@giovannicimolin Haven't written tests yet or checked for other things I may need to change to finish this, but wanted to get the core concept up here for anyone else to see.

giovannicimolin:
@geoff-va Thanks for the pull request! 

I'll start reviewing this week and focus on these points:
- Impact on user setup (which is already complicated - maybe we should go with sane defaults and allow editing if necessary?)
- Changes or issues with current integrations (I can test H5P, IMS Reference implementation - someone at edX can run the certification suite for launches/deep links and IMS-RI with deep linking to make sure current launches aren't affected).

@Zacharis278 or @ashultz0 Can you help me test this one?

lti_consumer/models.py
```diff
@@ -538,6 +550,42 @@ def get_lti_consumer(self):
...
+
+        return self._get_lti_1p3_redirect_uris_with_defaults(
+            redirect_uris,
+            launch_url,
+            deep_link_launch_url
+        )
+
+    def _get_lti_1p3_redirect_uris_with_defaults(self, redirect_uris, launch_url, deep_link_url):
+        """
+        Return provided redirect_uris if set, else use launch/deep_link as defaults
```
ashultz0, line 578: Is there any harm in always including the launch URL and deep link URL if one exists, since they are already there as trusted URLs?

lti_consumer/lti_xblock.py
```diff
@@ -314,6 +314,12 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
             "prior to doing the launch request."
         ),
     )
+    lti_1p3_redirect_uris = List(
+        display_name=_("Registered Redirect URIs"),
+        # TODO: Further clarify this help
+        help=_("Valid URLs the Tool may redirect you to. This may or may not be the same as the launch url."),
```
MichaelRoytman, line 320: I know you have a TODO on this, so perhaps you already have this in mind, but thought I'd mention it anyway. I think this help text is a little confusing as is. The redirect URIs represents URIs that the platform is allowed to redirect to on the tool's side, as specified by the tool during the preregistration phase. It's the platform that issues a redirect to one of these URIs, not the tool. The tool instructs the platform which of these URIs to use in the authentication request.

Regarding, "This may or may not be the same as the launch url.", is it your experience that tools don't explicitly specify redirect URIs and rely on the person setting up the integration to know that this is the launch URL otherwise? I haven't set up live LTI 1.3 tools yet - only the 1EdTech testing suite, which explicitly provides the redirect URIs. I just worry that course team authors or other users will just stick the launch URL in here.

lti_consumer/models.py
MichaelRoytman: Do we want to have defaults? The exchange of registered redirect URIs is a required part of the registration of a tool, right? Why are we making it an optional part of the configuration? Is it your experience that tools don't reliably and explicitly provide this information during registration?

MichaelRoytman (commented):
Thank you for the pull request! I left a few questions.

lti_consumer/models.py
geoff-va: Yes that’s exactly what I do now! if the redirect Uris are not specifically set then it uses the launch url and deep linking url as the defaults. Once you set anything in the redirect Uris field though, you would then need to include all possible redirect Uris. 

I did it this way so we it won’t affect the current setup of previous deployments.

The general use case we’re supporting here is that if the launch url and target link uri are the same then they only need to set the target link uri and/or deep linking uri. But there are tools where the redirect uri (the place we return the token to) is not the same as the target link uri (the final place the tool will redirect to once it redirects to the tools redirect uri). 

The valid redirect Uris should technically only ever be set to places the token should be returned to. No need to include the target link uri in this list because it wouldn’t know what to do with a token anyway. 

Hope that makes sense. Lots of redirect Uris and terminology here!

lti_consumer/lti_xblock.py
geoff-va: Correct, it was confusing which is why I left the TODO! I figured it would be something we needed to discuss. But yes your description is otherwise accurate.

> Is it your experience that tools don't explicitly specify redirect URIs and rely on the person setting up the integration to know that this is the launch URL otherwise?

No, the tool must always specify the allowed redirect uri's as that is part of the OIDC spec which is what LTI 1.3 security is modeled off. The point is more that the `target_link_uri` (what we call launch url) and the `redirect_uri` may not always be the same. As you stated the `redirect_uri` is where the token is POSTed to and must be registered on the platform, while the `target_link_uri` is where the tool would redirect you to after that step in order to display the final content.

lti_consumer/models.py
geoff-va: @MichaelRoytman I only implemented the defaults as a measure to account for what would otherwise be a breaking change for existing deployments. I'm fine with removing them as long as everyone understands that after upgrading to this version, their integration would cease to function until they added the appropriate urls to the new redirect uri's field.

lti_consumer/models.py
```diff
@@ -187,6 +188,18 @@ class LtiConfiguration(models.Model):
                   'Tool. One of either lti_1p3_tool_public_key or lti_1p3_tool_keyset_url must not be blank.'
     )
 
+    lti_1p3_redirect_uris = models.JSONField(
+        "LTI 1.3 Redirect URIs",
+        default=list,
+        blank=True,
+        # TODO: Further clarify this help
+        help_text="Valid urls the Tool may request us to redirect the id token to. The redirect uris "
```
geoff-va, line 196: @MichaelRoytman 

Do you think this help text is more clear or helps illustrate the intent?

geoff-va:
Hey @giovannicimolin 

I think I've got the core implementation all done and I've now added tests. I was able to touch test this with the reference implementation a bit.

**Basic Launch**
- If `redirect_uri's` is `[]`, launch succeeds (uses launch url as default)
- If `redirect_uri's` is `["<launch_url>"]`, launch succeeds
- if `redirect_uri's` is `["<not-launch-url>"]`, launch fails

**Deep Link Launch**
I was working on testing deep linking launch and had a couple of observations:
- It looks like we pass the [regular launch url](https://github.com/openedx/xblock-lti-consumer/blob/master/lti_consumer/lti_1p3/consumer.py#L123) as the `target_link_uri` even in the case of a deep linking launch
    - This causes the redirect uri that's returned to be the standard launch url, even in the case of a deep link launch (at least with the reference implementation)
- If we perform a deep link launch, we [set the redirect uri to the deep link launch uri](https://github.com/openedx/xblock-lti-consumer/blob/master/lti_consumer/plugin/views.py#L233-L239), regardless of what is returned as the redirect uri from the tool.

In the case of the reference implementation, if we set the `target_link_uri` to the deep linking launch url, then the tool sends the deep linking url back as the redirect uri, where we can properly validate it against the redirect uri's.

I'm not sure that the spec is especially clear here (unless I've missed it), but should we be sending the deep linking url as the target link uri during preflight (and id token) and then allow the tool to dictate the `redirect_uri`, assuming it's registered with us? I think this gives the tool our intent during preflight via the target link uri and then allows it to adjust the redirect uri if necessary?

giovannicimolin:
@geoff-va Thanks for the detailed comments. I'll try to review this on Monday - I don't have time available today.

@MichaelRoytman Can you do a testing round on the IMS global validation tool?

mphilbrick211:
@giovannicimolin @MichaelRoytman hi there! Just checking in on this to see if it still needs tests and review?

mphilbrick211:
@giovannicimolin and @MichaelRoytman - friendly follow-up on this!

giovannicimolin:
@mphilbrick211 @geoff-va Sorry for missing the pings! 
I'm not a core contributor anymore, so I don't have the capacity available to review this in the coming weeks.

@Zacharis278 or @MichaelRoytman Is this something you can take up?

Zacharis278:
@mphilbrick211 I'll get a task on our board to spend some time on this

mphilbrick211:
> @mphilbrick211 I'll get a task on our board to spend some time on this

Thanks, @Zacharis278!

Zacharis278:
@mphilbrick211 looks like this is falling approximately into our next sprint which is the second half of February. Is this blocking anything?

mphilbrick211:
> @mphilbrick211 looks like this is falling approximately into our next sprint which is the second half of February. Is this blocking anything?

Thanks @Zacharis278! No, I don't believe this is blocking anything.

Zacharis278:
@geoff-va I'm doing some testing on this now against the reference tool. If everything looks good I can approve to merge. In the mean time could you rebase this and resolve the conflicts? Thanks.

geoff-va:
@Zacharis278 For the changelog, do you want me to add the changes to the unreleased section and allow you guys to determine new version number and dates, etc?

geoff-va:
One other question - Would this be considered a possible breaking change? Since the `redirect_uri` was never validated at all in the past, it's entirely possible it's not the launch url or the deep linking url and an integration would cease to work until the redirect uris are added.

Was considering the following changelog entry:
```
BREAKING CHANGE:

* Validates Tool provided ``redirect_uri`` against stored ``lti_1p3_redirect_uris`` field per
  `LTI Specification <https://www.imsglobal.org/spec/security/v1p0/#step-3-authentication-response>`_
* ``lti_1p3_redirect_uris`` defaults to ``{lti_1p3_launch_url, lti_1p3_redirect_uris}`` when empty to provide
  minimal distruption to existing integrations
* **NOTE:** Since the redirect URI was never validated in the past, there is always a chance it is something
  other than the launch url/deep linking url, so you may have to explicitly set it as appropriate.
```

Zacharis278:
@geoff-va I think to be on the safe side we can consider this a breaking change and bump up to the next major version (9.0.0)

Zacharis278:
@geoff-va are you able to make the update to `changelog` and `__init__.py` for version 9.0.0?

geoff-va:
@Zacharis278 done!

geoff-va:
Also @Zacharis278  did you have any input about the deep linking comments I made [here](https://github.com/openedx/xblock-lti-consumer/pull/302#issuecomment-1327030753)? It seems like there could be some additional changes we may want to implement as part of this PR around what `target_link_uri` is sent and what we use as the `redirect_uri` during deep linking?

geoff-va:
Yea totally reasonable - Will create a new ticket for that. Thanks!

Zacharis278 (approved):
👍

openedx-webhooks:
@geoff-va 🎉 Your pull request was merged!  Please take a moment  to [answer a two question survey](https://docs.google.com/forms/d/e/1FAIpQLSceJOyGJ6JOzfy6lyR3T7EW_71OWUnNQXp68Fymsk3MkNoSDg/viewform?usp=pp_url&entry.1671973413=openedx/xblock-lti-consumer&entry.867055334=https://github.com/openedx/xblock-lti-consumer/pull/302&entry.1484655318=https://github.com/geoff-va&entry.752974735=2022-11-08+21:58&entry.1917517419=2023-03-06+16:46&entry.2133058324=Yes) so we can improve your experience in the future.

PR #280, opened 2022-09-07
Fix github url strings (org edx -> openedx)

sarina:
## This PR was autogenerated

This pr replaces the old GitHub organization, github.com/edx, with the new GitHub organization, github.com/openedx.

Tagging @openedx/tcril-engineering for review, but others are welcome to provide review.

Ref: https://github.com/openedx/tcril-engineering/issues/42

PR #308, opened 2022-11-21
Revert "Fix github url strings (org edx -> openedx)"

sarina:
Reverts openedx/xblock-lti-consumer#280

Didn't squash my commits


sarina:
nevermind... the badly worded commits will stay in history... i'm so sorry i'm an idiot 😭

PR #312, opened 2022-12-01
fix: remove lms specific waffle check

Zacharis278:
I'm running into two problems with this check when this code is run in edx-exams. Location isn't required on this model so location.course_key may fail but more importantly `database_config_enabled` can only run in the LMS. Checking if location exists first would technically fix this since a location implies this is running in the LMS (for now) but I'm coming to the conclusion that this check should probably not exist here.

All other instances of `database_config_enabled` occur within the lti_xblock.py which is totally fine to call into the LMS compatibility layer but baking this function into the data model feels like something that will catch us again later. The waffle flag already gates the UI behavior, do we really need this check on the model also?

Let know if I'm missing some good arguments for keeping this in, otherwise I propose we remove it.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,10 @@ Please See the [releases tab](https://github.com/openedx/xblock-lti-consumer/rel
 Unreleased
 ~~~~~~~~~~
 
+7.0.3 - 2022-12-02
+------------------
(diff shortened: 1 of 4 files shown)


ashultz0 (approved):

schenedx (approved):
👍

MichaelRoytman (approved):
This makes sense to me. I agree that the waffle flag should not be in the data model for the reasons you outlined.

PR #317, opened 2022-12-15
Fix PII Sharing Behavior and Enable PII Sharing in LTI 1.3 Launches and Fix LTI 1.3 Modal Launches

MichaelRoytman:
### Description

This pull request fixes a few issues and bugs with sharing personally identifiable information (PII) in LTI launches as well as one bug with the `modal` LTI launch in LTI 1.3.

The follow issues and bugs were resolved in this pull request:
* `username` and `email` are not sent in LTI 1.3 launches
* the PII sharing LTI XBlock fields always appear, regardless of the existence and value of a `CourseAllowPIISharingInLTIFlag`
* `inline` LTI launches did not trigger a PII consent modal
* `modal` LTI launches did not trigger a PII consent modal
* the `modal` LTI launch for LTI 1.3 failed due to differences in the form of the launch URL provided to the Javascript via the XBlock code
* the values returned by `lti_access_to_learners_editable` were not correct, and the values were not being recomputed on each request

Please see the `CHANGELOG` for a more detailed description of the bugs and their fixes.

### UI/UX Changes

The PII consent modal was replaced with an inline consent dialog for all LTI launch types and for both LTI 1.1 and LTI 1.3.

**LTI 1.1**

MST-1736-Demo-LTI-PII-Consent-Dialog-LTI-1 1

**LTI 1.3**

MST-1736-Demo-LTI-PII-Consent-Dialog-LTI-1 3-Inline
MST-1736-Demo-LTI-PII-Consent-Dialog-LTI-1 3-Modal
MST-1736-Demo-LTI-PII-Consent-Dialog-LTI-1 3-New-Window

### Manual Testing

These changes were tested on all three LTI launch types (i.e. `inline`, `modal` and `new_window`) across both LTI 1.1 and LTI 1.3.

* I tested that the `CourseAllowPIISharingInLTIFlag` controls the appearance of the PII sharing LTI XBlock fields.
* I tested that the consent dialog appears for all three LTI launch types and works as expected.
* I tested that the consent dialog appears only when any or all of the PII sharing LTI XBlock fields is set to True.
* I tested that the backward compatibility code in `lti_access_to_learners_editable` works correctly. Instances of `CourseAllowPIISharingInLTIFlag` are created successfully when appropriate.
* I tested that `lti_access_to_learners_editable` returns the correct value.
* I tested that `username` and `email` are transmitted in LTI 1.3 launches when appropraite.

### DIFF
--- a/lti_consumer/lti_xblock.py
+++ b/lti_consumer/lti_xblock.py
@@ -1590,6 +1590,7 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
             'modal_horizontal_offset': self._get_modal_position_offset(self.modal_width),
             'modal_width': self.modal_width,
             'accept_grades_past_due': self.accept_grades_past_due,
+            'lti_version': self.lti_version,
         }
 
     def _get_modal_position_offset(self, viewport_percentage):
--- a/lti_consumer/lti_xblock.py
+++ b/lti_consumer/lti_xblock.py
@@ -673,6 +673,26 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
             log.exception('Something went wrong in reading the LTI XBlock configuration.')
             raise
 
+    def get_pii_sharing_enabled(self):
+        """
+        Returns whether PII can be transmitted via this XBlock. This controls both whether the PII sharing XBlock
+        fields ask_to_send_username and ask_to_send_email are displayed in Studio and whether these data are shared
+        in LTI launches, regardless of the values of the settings on the XBlock.
+        """
+        config_service = self.runtime.service(self, 'lti-configuration')
+        if config_service:
+            is_already_sharing_learner_info = self.ask_to_send_email or self.ask_to_send_username
+
+            return config_service.configuration.lti_access_to_learners_editable(
+                self.scope_ids.usage_id.context_key,
+                is_already_sharing_learner_info,
+            )
+
+        # TODO: The LTI configuration service is currently only available from the studio_view. This means that
+        #       the CourseAllowPIISharingInLTIFlag does not control PII sharing in the author_view or student_view,
+        #       because the service is not defined in those contexts.
+        return True
+
     @property
     def editable_fields(self):
         """
@@ -710,14 +730,9 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
 
         # update the editable fields if this XBlock is configured to not to allow the
         # editing of 'ask_to_send_username' and 'ask_to_send_email'.
-        config_service = self.runtime.service(self, 'lti-configuration')
-        if config_service:
-            is_already_sharing_learner_info = self.ask_to_send_email or self.ask_to_send_username
-            if not config_service.configuration.lti_access_to_learners_editable(
-                    self.scope_ids.usage_id.context_key,
-                    is_already_sharing_learner_info,
-            ):
-                noneditable_fields.extend(['ask_to_send_username', 'ask_to_send_email'])
+        pii_sharing_enabled = self.get_pii_sharing_enabled()
+        if not pii_sharing_enabled:
+            noneditable_fields.extend(['ask_to_send_username', 'ask_to_send_email'])
 
         editable_fields = tuple(
             field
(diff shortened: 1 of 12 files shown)


lti_consumer/lti_xblock.py
```diff
@@ -673,6 +673,26 @@ def get_parameter_processors(self):
...
+        config_service = self.runtime.service(self, 'lti-configuration')
+        if config_service:
+            is_already_sharing_learner_info = self.ask_to_send_email or self.ask_to_send_username
+
+            return config_service.configuration.lti_access_to_learners_editable(
+                self.scope_ids.usage_id.context_key,
+                is_already_sharing_learner_info,
+            )
+
+        # TODO: The LTI configuration service is currently only available from the studio_view. This means that
```
MichaelRoytman, line 691: I'll follow up on this, but I didn't want to muck around in the runtimes on such short notice and without knowing what I'm doing.

The effect of this is we lose the kill switch functionality of the `CourseAllowPIISharingInLTIFlag`, but that is already the case.

lti_consumer/models.py
```diff
@@ -778,7 +777,6 @@ class CourseAllowPIISharingInLTIFlag(ConfigurationModel):
     course_id = CourseKeyField(max_length=255, db_index=True)
 
     @classmethod
-    @request_cached
```
MichaelRoytman, line 781: The caching was not working correctly. It appeared that the cached value was not being recomputed on a per-request basis, which meant that the cached value was not successfully being updated. Also, the initial cached value was computed incorrectly. The effect was that the PII sharing XBlock fields were always being displayed in Studio and ``CourseAllowPIISharingInLTIFlag`` instances were not being created for courses that had PII sharing XBlocks that did not already have ``CourseAllowPIISharingInLTIFlag`` instances.

This fixes the backwards compatibility of the ``CourseAllowPIISharingInLTIFlag``. This means that ``CourseAllowPIISharingInLTIFlag`` will now be created for all courses that contain an LTI XBlock that shares username and/or email for which a ``CourseAllowPIISharingInLTIFlag`` does not already exist.

Time was spent on determining why the caching was failing, but it was challenging. If the issue is in the ``request_cached`` decorator, then the impact of changing it is large - it's used elsewhere in the platform. The need for a fix was urgent, and the performance enhancement of leaving the caching in doesn't justify delaying a fix.

lti_consumer/static/js/xblock_lti_consumer.js
```diff
@@ -92,68 +105,97 @@ function LtiConsumerXBlock(runtime, element) {
             }
         });
 
+        function confirmDialog(message, triggerElement, showCancelButton) {
+            var def = $.Deferred();
+            // Hide the button that triggered the event, i.e. the launch button.
+            triggerElement.hide();
+
+            $('<div id="dialog-container"></div>').insertAfter(triggerElement) // TODO: this will need some cute styling. It looks like trash but it works.
```
MichaelRoytman, line 113: `TODO: this will need some cute styling. It looks like trash but it works.`

I didn't write this comment, but it's still ugly, so I guess I'll leave it in...

lti_consumer/static/js/xblock_lti_consumer.js
```diff
@@ -92,68 +105,97 @@ function LtiConsumerXBlock(runtime, element) {
...
-                    modal: true,
-                    title: 'Confirm',
-                    zIndex: 10000,
-                    autoOpen: true,
-                    width: 'auto',
-                    resizable: false,
-                    dialogClass: 'confirm-dialog',
-                    buttons: {
-                      OK: function() {
-                        $('body').append('<h1>Confirm Dialog Result: <i>Yes</i></h1>');
```
MichaelRoytman, line 124: I don't know what the purpose of these headers was. Anyone feel strongly I should leave them in? They don't appear in the LMS, but I could see them at the bottom of the page in Studio.

lti_consumer/models.py
Zacharis278: This is only called by studio anyway right? So the performance impact here doesn't seem all that concerning.

Zacharis278 (approved):
ran through the changes and this all looks good to me 👍

lti_consumer/models.py
MichaelRoytman: Right now, yes. Once we add the LTI configuration service to the right places, it will be called from the LMS as well.

lti_consumer/static/js/xblock_lti_consumer.js
MichaelRoytman: I was wondering if it was an accessibility thing. Thanks for finding that. I'll add it back in.

PR #319, opened 2022-12-20
feat: support separate different base urls for UI flow and API callbacks

Zacharis278:
### [MST-1732](https://2u-internal.atlassian.net/browse/MST-1732)

Allows independent configuration of the base URL used for LTI API requests and LTI browser flow. This primarily aids local development because we no longer have to tunnel the entire LMS in order to test against the IMS tools. You can use your normal browser steps to login to the LMS and access content without hacking up your settings which can break other parts of the platform.

Testing LTI launches in another service or IDA is particularly problematic right now because we need to tunnel multiple services (LMS + IDA).

This change would also support cases where a service may have a separate URL scheme for backend requests like the OAuth flow versus frontend redirects, which need to make use of specific browser cookies.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -203,8 +203,8 @@ Instructions:
     Client ID: f0532860-cb34-47a9-b16c-53deb077d4de
     Deployment ID: 1
     # Note that these are LMS URLS
-    Keyset URL: http://1234.ngrok.io/api/lti_consumer/v1/public_keysets/block-v1:OpenCraft+LTI101+2020_T2+type@lti_consumer+block@efc55c7abb87430883433bfafb83f054
-    Access Token URL: http://1234.ngrok.io/api/lti_consumer/v1/token/block-v1:OpenCraft+LTI101+2020_T2+type@lti_consumer+block@efc55c7abb87430883433bfafb83f054
+    Keyset URL: http://1234.ngrok.io/api/lti_consumer/v1/public_keysets/88e45ecbd-7cce-4fa0-9537-23e9f7288ad9
+    Access Token URL: http://1234.ngrok.io/api/lti_consumer/v1/token/8e45ecbd-7cce-4fa0-9537-23e9f7288ad9
     OIDC Callback URL: http://localhost:18000/api/lti_consumer/v1/launch/
 
 
--- a/README.rst
+++ b/README.rst
@@ -184,39 +184,36 @@ needs to know the LMS's one.
 
 Instructions:
 
-1. Set up a local tunnel tunneling the LMS (using `ngrok` or a similar tool) to get a URL accessible from the internet.
-2. Create a new course, and add the `lti_consumer` block to the advanced modules list.
-3. In the course, create a new unit and add the LTI block.
+#. Set up a local tunnel tunneling the LMS (using `ngrok` or a similar tool) to get a URL accessible from the internet.
+#. Add the following settings to `edx-platform/lms/envs/private.py` and `edx-platform/cms/envs/private.py`:
+
+    * LTI_VIEW_BASE_URL="http://localhost:18000"
+    * LTI_API_BASE_URL="http://<your_ngrok>.ngrok.io"
+
+#. Create a new course, and add the `lti_consumer` block to the advanced modules list.
+#. In the course, create a new unit and add the LTI block.
 
    * Set ``LTI Version`` to ``LTI 1.3``.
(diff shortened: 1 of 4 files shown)


lti_consumer/utils.py
```diff
@@ -26,22 +26,36 @@ def _(text):
...
     and strip the base domain name.
     """
-    if hasattr(settings, 'LTI_LMS_BASE_URL_OVERRIDE'):
-        return settings.LTI_LMS_BASE_URL_OVERRIDE
+    if hasattr(settings, 'LTI_API_BASE_URL'):
+        return settings.LTI_API_BASE_URL
+    elif hasattr(settings, 'LTI_VIEW_BASE_URL'):
+        return settings.LTI_VIEW_BASE_URL
     else:
+        # TODO: we should move away from supporting this setting as it conflicts with usage
```
Zacharis278, line 44: Since this lib is still installed as an xblock I didn't want to go removing support for this just yet. That said, we do need a new setting because we can't hijack the use of LMS_ROOT_URL if this is running outside the LMS. This should point at the API for the current runtime and LMS_ROOT_URL is likely already pointing at the LMS.

ashultz0: I wouldn't be surprised if we left this in forever - it's just too useful for the default LTI running out of the LMS case. We only need it to be different in dev and test and new IDAs

lti_consumer/utils.py
```diff
@@ -26,22 +26,36 @@ def _(text):
...
     TODO: This needs to be improved and account for Open edX sites and
     organizations.
     One possible improvement is to use `contentstore.get_lms_link_for_item`
     and strip the base domain name.
     """
-    if hasattr(settings, 'LTI_LMS_BASE_URL_OVERRIDE'):
-        return settings.LTI_LMS_BASE_URL_OVERRIDE
+    if hasattr(settings, 'LTI_API_BASE_URL'):
+        return settings.LTI_API_BASE_URL
+    elif hasattr(settings, 'LTI_VIEW_BASE_URL'):
```
ashultz0, line 41: it's interesting that API defaults to VIEW if available but not the other way around

ashultz0 (approved):

lti_consumer/utils.py
Zacharis278: I was thinking of this as view url is required and api url is an optional override but the naming I chose doesn't really make that clear.

lti_consumer/utils.py
Zacharis278: my gut is this can't go anywhere unless we break out the entire xblock part into it's own codebase. So agreed, not any time soon.

Zacharis278:
@MichaelRoytman you have any concerns with this one?

README.rst
```diff
@@ -184,55 +184,45 @@ needs to know the LMS's one.
...
+#. In Studio, you'll see a few parameters being displayed in the preview:
 
 .. code::
 
     Client ID: f0532860-cb34-47a9-b16c-53deb077d4de
     Deployment ID: 1
     # Note that these are LMS URLS
-    Keyset URL: http://localhost:18000/api/lti_consumer/v1/public_keysets/block-v1:OpenCraft+LTI101+2020_T2+type@lti_consumer+block@efc55c7abb87430883433bfafb83f054
-    Access Token URL: http://localhost:18000/api/lti_consumer/v1/token/block-v1:OpenCraft+LTI101+2020_T2+type@lti_consumer+block@efc55c7abb87430883433bfafb83f054
+    Keyset URL: http://1234.ngrok.io/api/lti_consumer/v1/public_keysets/block-v1:OpenCraft+LTI101+2020_T2+type@lti_consumer+block@efc55c7abb87430883433bfafb83f054
```
MichaelRoytman, line 206: Since you're changing the documentation, would you mind updating these URLs? I noticed they are outdated. They now include the external UUID of the corresponding `LtiConfiguration` instead of the XBlock location. Something like this `/api/lti_consumer/v1/public_keysets/8e45ecbd-7cce-4fa0-9537-23e9f7288ad9`.

I can change it if it's easier for you - just thought it would be easier to roll into your pull request than having a standalone documentation change!

MichaelRoytman (approved):
Thank you for your patience. This looks good to me. I left a small requested change to the documentation, but I'm happy to do it myself - just let me know.

If you can hit the coverage target, that would be great.

README.rst
Zacharis278: yeah, easy add

PR #325, opened 2023-01-24
Fix bug in rending buttons and message in PII sharing consent dialog.

MichaelRoytman:
This commit fixes a bug in the PII sharing consent dialog.

The bug resulted in bizarre behavior when there were more than one LTI component in a unit. For example, if there were two LTI inline launches in a unit, two "OK" button would appear in a single component, instead of in their respective components. Another example is that clicking the "View resource in a [modal|new] window" buttons under two LTI components resulted in the "OK" and "Cancel" buttons as well as the PII sharing prompt appearing in a single component, instead of in their respective components.

This is because the dialog-container div that is dynamically created in the Javascript was not scoped to the LTI component, so there was a div with a id of "dialog-container" for each component configured to share PII. When dynamically inserting and removing buttons and the PII sharing prompt, the Javascript would simply find the first div with the dialog-container ID and operate on it, instead of the div appropriate to the component the user is interacting with.

### Before

**Two Inline Launches**
image

**Two Non-Inline Launches**
image

### After

**Two Inline Launches**
image

**Two Non-Inline Launches**
image

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,12 @@ Please See the [releases tab](https://github.com/openedx/xblock-lti-consumer/rel
 Unreleased
 ~~~~~~~~~~
 
+7.2.3 - 2023-01-24
+------------------
+* This release fixes a bug in the way that the PII sharing consent dialog renders. The bug resulted in the "OK" and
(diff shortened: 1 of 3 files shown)


alangsto (approved):
LGTM 👍

Zacharis278 (approved):

PR #328, opened 2023-01-30
feanil/make upgrade

feanil:
- chore: Run `make upgrade`.
- fix: Deal with backward incompatible changes in bleach.

### DIFF
--- a/lti_consumer/templatetags/lti_sanitize.py
+++ b/lti_consumer/templatetags/lti_sanitize.py
@@ -13,7 +13,7 @@ def lti_sanitize(html):
     """
     Sanitize a html fragment with bleach.
     """
-    allowed_tags = bleach.sanitizer.ALLOWED_TAGS + ['img']
+    allowed_tags = bleach.sanitizer.ALLOWED_TAGS | {'img'}
     allowed_attributes = dict(bleach.sanitizer.ALLOWED_ATTRIBUTES, **{'img': ['src', 'alt']})
     sanitized_html = bleach.clean(html, tags=allowed_tags, attributes=allowed_attributes)
     return mark_safe(sanitized_html)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,12 @@ Please See the [releases tab](https://github.com/openedx/xblock-lti-consumer/rel
 Unreleased
 ~~~~~~~~~~
 
+8.0.0 - 2023-01-31
+------------------
+* Update to work with bleachk>=6.0.0 and make that an explicit requirement in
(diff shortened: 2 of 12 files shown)


idegtiarov (approved):
LGTM

lti_consumer/templatetags/lti_sanitize.py
```diff
@@ -13,7 +13,7 @@ def lti_sanitize(html):
     """
     Sanitize a html fragment with bleach.
     """
-    allowed_tags = bleach.sanitizer.ALLOWED_TAGS + ['img']
+    allowed_tags = bleach.sanitizer.ALLOWED_TAGS | {'img'}
```
Agrendalath, line 16: Won't this break in `edx-platform`? According to [the changelog](https://bleach.readthedocs.io/en/latest/changes.html#version-6-0-0-january-23rd-2023), the list -> dict change was introduced in `bleach==6.0.0`.
`edx-platform` is still using [bleach[CSS]==5.0.1](https://github.com/openedx/edx-platform/blob/e576478/requirements/edx/base.txt#L61) because it is pinned [here](https://github.com/openedx/edx-platform/blob/e576478/requirements/constraints.txt#L79). If we use the newer version in the CI, then we won't catch any incompatibilities.

Agrendalath (changes requested):

Agrendalath (approved):
The changelog entry is missing, but other than that it looks good to me.

:+1:

* [x] I tested this: checked that the CI is passing with updated dependencies and read bleach changelog for the new major release
* [x] I read through the code
* [x] I checked for accessibility issues: n/a
* [x] Includes documentation: n/a
* [x] I made sure any change in configuration variables is reflected in the corresponding client's `configuration-secure` repository: n/a

lti_consumer/templatetags/lti_sanitize.py
feanil: Done, thanks for the review!

Agrendalath: @feanil, please see [my suggestion from the Drag and Drop v2 XBlock](https://github.com/openedx/xblock-drag-and-drop-v2/pull/308#discussion_r1092974520) before merging this.

feanil: @Agrendalath sorry I missed this comment, let me know if you think it's worth it to make this backwards compatible, I'm happy to put up or review a PR to that effect.

Agrendalath: @feanil, no worries. Since the major version is already released, let's keep it as is.

PR #359, opened 2023-04-18
feat: add logging for access token view and key handlers

MichaelRoytman:
JIRA: [MST-1853](https://2u-internal.atlassian.net/browse/MST-1853)

This commit adds logging for error cases in the access_token_endpoint view and the key handler classes. This will make it easier to debug issues with requesting or using LTI 1.3 access tokens.


alangsto (approved):
LGTM 👍

PR #365, opened 2023-05-01
feat: add platform name to 1.1 and 1.3 claims

alangsto:
## [MST-1880](https://2u-internal.atlassian.net/browse/MST-1880)

Some LTI tools require a platform name in order for the LTI integration to be successful. The name is sent using the `tool_consumer_info_product_family_code` parameter for LTI 1.1, and the platform instance claim for LTI 1.3.

### DIFF
--- a/lti_consumer/lti_1p3/consumer.py
+++ b/lti_consumer/lti_1p3/consumer.py
@@ -3,6 +3,9 @@ LTI 1.3 Consumer implementation
 """
 import logging
 from urllib.parse import urlencode
+import uuid
+
+from django.conf import settings
 
 from lti_consumer.lti_1p3.exceptions import InvalidClaimValue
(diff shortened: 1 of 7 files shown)


Zacharis278 (approved):
This looks like the correct place to include this to me. 👍

lti_consumer/lti_1p3/consumer.py
```diff
@@ -349,6 +352,16 @@ def get_lti_launch_message(
             if self.lti_claim_context:
                 lti_message.update(self.lti_claim_context)
 
+            # Platform instance claim
+            # The GUID must be consistant across platform deployments, so we have opted to generate a UUID
```
MichaelRoytman, line 356: Teeny tiny nit!

```suggestion
            # The GUID must be consistent across platform deployments, so we have opted to generate a UUID
```

MichaelRoytman (approved):
This looks great to me! I just left one small nit.

PR #366, opened 2023-05-02
feat: support assessment control claims

Zacharis278:
### [MST-1846](https://2u-internal.atlassian.net/browse/MST-1846)

Adds optional support for the proctoring assessment control service (ACS)

Example of how to use this in edx-exams: https://github.com/edx/edx-exams/pull/122

### DIFF
--- a/lti_consumer/data.py
+++ b/lti_consumer/data.py
@@ -5,6 +5,8 @@ by users of this library.
 
 from attrs import define, field, validators
 
+from lti_consumer.lti_1p3.constants import LTI_PROCTORING_ASSESSMENT_CONTROL_ACTIONS
+
 
 @define
 class Lti1p3ProctoringLaunchData:
@@ -27,6 +29,14 @@ class Lti1p3ProctoringLaunchData:
     """
     attempt_number = field()
     start_assessment_url = field(default=None)
+    assessment_control_url = field(default=None)
+    assessment_control_actions = field(
+        default=[],
+        validator=[validators.deep_iterable(
+            member_validator=validators.in_(LTI_PROCTORING_ASSESSMENT_CONTROL_ACTIONS),
+            iterable_validator=validators.instance_of(list),
+        )],
+    )
 
 
 @define
--- a/lti_consumer/data.py
+++ b/lti_consumer/data.py
@@ -26,6 +26,8 @@ class Lti1p3ProctoringLaunchData:
         assessment message to after it has completed the proctoring setup and verification. This attribute is required
         if the message_type attribute of the Lti1p3LaunchData instance is "LtiStartProctoring". It is optional and
         unused otherwise.
+    * assessment_control_url (optional): The Platform URL that the Tool will send assessment control messages to.
+    * assessment_control_actions (optional): A list of assessment control actions supported by the platform.
     """
     attempt_number = field()
     start_assessment_url = field(default=None)
--- a/lti_consumer/lti_1p3/consumer.py
+++ b/lti_consumer/lti_1p3/consumer.py
@@ -809,6 +809,17 @@ class LtiProctoringConsumer(LtiConsumer1p3):
 
         return proctoring_claims
 
+    def get_assessment_control_claims(self):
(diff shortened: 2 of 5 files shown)


lti_consumer/lti_1p3/consumer.py
```diff
@@ -843,6 +854,9 @@ def generate_launch_request(
 
         self.set_extra_claim(proctoring_claims)
 
+        if self.proctoring_data.get("assessment_control_url"):
```
MichaelRoytman, line 857: Nit. Should we ensure that `launch_data.message_type == "LtiStartProctoring"` here as well? I think these claims are only valid in the `LtiStartProctoring` message, and we, theoretically, shouldn't send these claims in the `LtiEndAssessment` message. Or maybe this belongs in the [validate_lti_1p3_launch_data](https://github.com/openedx/xblock-lti-consumer/blob/3eaf8aa7ce0f3bd601bf318db48d7f8db50445e8/lti_consumer/api.py#L247) function or as a validator on the `Lti1p3LaunchData` class?

lti_consumer/data.py
```diff
@@ -24,9 +26,19 @@ class Lti1p3ProctoringLaunchData:
         assessment message to after it has completed the proctoring setup and verification. This attribute is required
         if the message_type attribute of the Lti1p3LaunchData instance is "LtiStartProctoring". It is optional and
         unused otherwise.
+    * assessment_control_url (optional): The Platform URL that the Tool will send assessment control messages to.
+    * assessment_control_actions (optional): A list of assessment control actions supported by the platform.
     """
     attempt_number = field()
     start_assessment_url = field(default=None)
+    assessment_control_url = field(default=None)
```
MichaelRoytman, line 34: Should we somehow validate that if one of these is supplied that the other is also necessary? They're both required, and I have done some validation of the data class previously. We do have a [validate_lti_1p3_launch_data](https://github.com/openedx/xblock-lti-consumer/blob/3eaf8aa7ce0f3bd601bf318db48d7f8db50445e8/lti_consumer/api.py#L247) function we could add to. I think that's easier than validating when initializing the `Lti1p3LaunchData` class, unless you know how to do that.

lti_consumer/data.py
```diff
@@ -24,9 +26,19 @@ class Lti1p3ProctoringLaunchData:
         assessment message to after it has completed the proctoring setup and verification. This attribute is required
         if the message_type attribute of the Lti1p3LaunchData instance is "LtiStartProctoring". It is optional and
         unused otherwise.
+    * assessment_control_url (optional): The Platform URL that the Tool will send assessment control messages to.
```
MichaelRoytman, line 29: Did you have any thoughts about whether to add fields like `lti_1p3_proctoring_acs_enabled` and `lti_1p3_acs_actions` to the `LtiConfiguration` model? These data are specific to the integration, so they could be stored on the model, whereas the remainder of the data in this class can't be stored on the model because it isn't meant to be configurable.

That feels like overkill, but I also ask because that matches the pattern we have for the Advantage Services and the field for enabling proctoring on the LTI integration (`lti_1p3_proctoring_enabled`), so I'm mainly curious about our approach to when to add data to the model and when to send data through as part of this data class.

MichaelRoytman (approved):
I left some comments about supplying and validating the data supplied by the consuming library. Otherwise, this looks good to me.

lti_consumer/lti_1p3/consumer.py
Zacharis278: ahh good call, wasn't sure if this was valid in just startProctoring or not. Lemme double check the spec b/c this is easy enough to just move inside the above codeblock

lti_consumer/data.py
Zacharis278: @MichaelRoytman I started down the path of putting this on the model but pulled back and just chose to comment on the inconsistency. In our current implementation enabling ACS isn't optional or configurable (it must match what exams expects) so having this on the model doesn't make much sense right now.

PR #373, opened 2023-05-18
feat: only add acs scope when proctoring

ilee2u:
Relevant ticket: [MST-1897](https://2u-internal.atlassian.net/browse/MST-1897)

### DIFF
--- a/lti_consumer/lti_1p3/consumer.py
+++ b/lti_consumer/lti_1p3/consumer.py
@@ -409,6 +409,26 @@ class LtiConsumer1p3:
         """
         return self.key_handler.get_public_jwk()
 
+    def _get_access_token_scopes(self, scopes):
+        """
+        Helper function to return the list of valid scopes present in the
+        request to the access token endpoint.
+        """
+        # TODO: Make this function should only return the NRPS and AGS scopes if those features are enabled.
+        valid_scopes = []
+        for scope in scopes:
+            # Currently there are no scopes, because there is no use for
+            # these access tokens until a tool needs to access the LMS.
+            # LTI Advantage extensions make use of this.
+            if scope in LTI_1P3_ACCESS_TOKEN_SCOPES:
+                valid_scopes.append(scope)
+            else:
+                log.warning(
+                    f'Scope: {scope} found in the request is not a valid '
+                    f'LTI 1.3 access token or ACS scope'
+                )
+        return valid_scopes
+
     def access_token(self, token_request_data):
         """
         Validate request and return JWT access token.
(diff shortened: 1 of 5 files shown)


ilee2u:
Note for reviewer: I've written a dirty fix for this issue in

Zacharis278:
Good instincts on the issue here, and you're right chopping up the access_token function just makes things messy. I think I see a way out of this. The problem seems to be this line `if scope in LTI_1P3_ACCESS_TOKEN_SCOPES:` where we assume the list of scopes is constant. What if instead we have a getter function for the list of scopes so now we're doing `if scope in self.get_access_token_scopes():` This way we can correctly calculate the list of scopes on LtiProctoringConsumer differently from LtiAdvantageConsumer or LtiConsumer1p3 independent of the access token function which shouldn't really change between different consumer classes.

Actually it seems to me that we're already doing this wrong because I'd think NRPS and AGS would only ever return their respective scopes if those features are enabled. I feel like that's a gap we don't need to solve today but could be worth leaving a comment about it.

ilee2u:
> Good instincts on the issue here, and you're right chopping up the access_token function just makes things messy. I think I see a way out of this. The problem seems to be this line `if scope in LTI_1P3_ACCESS_TOKEN_SCOPES:` where we assume the list of scopes is constant. What if instead we have a getter function for the list of scopes so now we're doing `if scope in self.get_access_token_scopes():` This way we can correctly calculate the list of scopes on LtiProctoringConsumer differently from LtiAdvantageConsumer or LtiConsumer1p3 independent of the access token function which shouldn't really change between different consumer classes.
> 
> Actually it seems to me that we're already doing this wrong because I'd think NRPS and AGS would only ever return their respective scopes if those features are enabled. I feel like that's a gap we don't need to solve today but could be worth leaving a comment about it.

Ok, I went ahead and coded all of this out by making a helper function in LtiConsumer1p3 and overriding said function in LtiProctoringConsumer. I also left a comment regarding the NRPS and AGS scopes in the first version of the function.

ilee2u:
Also, should I add a positive & negative test to TestLtiProctoringConsumer for the ACS scope?

lti_consumer/lti_1p3/consumer.py
```diff
@@ -407,6 +409,26 @@ def get_public_keyset(self):
...
         return self.key_handler.get_public_jwk()
 
+    def _get_access_token_scopes(self, scopes):
+        """
+        Helper function to return the list of valid scopes present in the
+        request to the access token endpoint.
+        """
+        # TODO: Make this function should only return the NRPS and AGS scopes if those features are enabled.
+        valid_scopes = []
+        for scope in scopes:
```
Zacharis278, line 419: small suggestion, we do the looping over requested scopes in both this and the Proctoring consumer `_get_access_token_scopes`. We could just keep that part inside `access_token()` right? Might make more sense because 'scopes' in this function is the requested list and that's not super clear anymore when reading this function alone.

ilee2u:
Note on coverage: I have not written a test to see if `access_token()` returns the scopes in LTI_1P3_ACCESS_TOKEN_SCOPES since those tests would be changed whenever we implement a gate to ensure the NRPS or NRPS are only added when their respective features are enabled as Zach noted in a previous comment.

lti_consumer/lti_1p3/consumer.py
```diff
@@ -877,6 +895,15 @@ def generate_launch_request(
...
         return super().generate_launch_request(preflight_response)
 
+    def _check_if_scope_is_valid(self, scope):
+        """
+        Override of Lti1p3Consumer's _check_if_scope_is_valid
+        that accounts for the ACS scope for proctoring.
+        """
+        if scope in (LTI_1P3_ACCESS_TOKEN_SCOPES, LTI_1P3_ACS_SCOPE):
+            return True
+        return False
```
Zacharis278, line 905: I think you should be able to cover this missing case with a test for a token with a non-valid scope.

Zacharis278 (commented):
small coverage comment if it's a easy fix. Otherwise the only thing we need additionally is an entry in changelog and a version bump to __init__.py so this can get pulled into edx-exams.

PR #376, opened 2023-05-24
feat: allow proctoring start from iframe

Zacharis278:
Allows for start proctoring message result to be rendered in an iframe. This isn't full support for proctoring inside an iframe there are still issues with rendering our learning MFE this way. This does unblock testing the launch flow at minimum.


ilee2u (approved):
LGTM 🚀

PR #382, opened 2023-06-08
feat: redirect to exam

varshamenon4:


lti_consumer/plugin/views.py
```diff
@@ -835,4 +835,10 @@ def start_proctoring_assessment_endpoint(request):
         user_id=request.user.id,
     )
 
-    return render(request, 'html/lti_start_assessment.html', status=HTTP_200_OK)
+    context_url = "/".join([launch_data.context_title, "course", launch_data.context_id, launch_data.context_label])
```
varshamenon4, line 838: Question: I figure that this might be a clean/efficient way to concatenate the url but open to feedback here to make this better/more readable!

lti_consumer/templates/html/lti_start_assessment.html
```diff
@@ -1,19 +1,14 @@
...
     <body>
         <p>
-            <b>{% trans "Sending you back to your exam." %}</b>
-        </p>
-        <script>
-            setTimeout(function () {window.close()}, 5000);
-        </script>
-        <p>
-            <a href="javascript:window.close();">
+            <a href="{{ context_url }}" target="_blank">
```
varshamenon4, line 11: Let me know if I should still keep this. My thought was if it didn't redirect right away. Currently, when testing, it just flashes this before going to the exam.

varshamenon4 (commented):
Just some questions/areas to look at. I will also include a screenshot/screen recording if I can to the PR so you can see how it works currently.

lti_consumer/plugin/views.py
alangsto: I think this is readable. This is where you could grab the MFE URL via settings instead of using `context_tile`

lti_consumer/templates/html/lti_start_assessment.html
alangsto: I think there should be some text on the page saying "redirecting you back to your exam in 5 seconds" or something like that, so the user is aware of what's happening. It might also be useful to provide the user a direct link back to the exam (that renders in the same tab) just in case the redirect doesn't work. That UI pattern is pretty common!

varshamenon4: So I just updated this to basically look the same as it was before (so the copy is the same and there is the same 5 second delay).

varshamenon4: Ah you're right, good catch! Right now, line 6 is opening in the same tab and that link just opens in case that doesn't happen. But that was from the prior experience, so you're right I should update that! Thanks!

ilee2u (approved):
LGTM

PR #386, opened 2023-06-26
POC for course tool launch

Zacharis278:
This endpoint would expose a view a frontend app could load given an lti configuration id and a course id. Loading the view would kick off a basic 1.3 launch.

This is definitely pretty limited in what we could support without building out further modeling for what a platform wide or course wide configuration might look like. I think the long term solution very likely depends on OpenCrafts pending work around remodeling configurations / placements. Launch by placement id seems like the more correct end state for this, we just don't have placements yet. In the meantime maybe we can build something really limited to support proctoring or other tools that are course or platform wide.

### DIFF
--- a/lti_consumer/plugin/views.py
+++ b/lti_consumer/plugin/views.py
@@ -9,7 +9,7 @@ from django.contrib.auth import get_user_model
 from django.core.exceptions import ObjectDoesNotExist, PermissionDenied, ValidationError
 from django.db import transaction
 from django.http import Http404, JsonResponse
-from django.shortcuts import render
+from django.shortcuts import redirect, render
 from django.utils.crypto import get_random_string
 from django.views.decorators.clickjacking import xframe_options_exempt, xframe_options_sameorigin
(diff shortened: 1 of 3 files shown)


lti_consumer/plugin/views.py
```diff
@@ -82,6 +83,45 @@ def has_block_access(user, block, course_key):
...
+    if lti_config.location:
+        raise PermissionDenied
+
+    location = UsageKey.from_string(usage_id)
+    course_key = location.course_key
+
+    # Problem. How can we get the user's role for the passed in course/context?
+    # We have to use a compat layer to do this which gets messy because this isn't
+    # just needed in the LMS but also exams.
+    user_role = None
```
Zacharis278, line 112: this is really problematic when it comes to running this library in anything other than the LMS. We need to ask the running service for user permissions.

Zacharis278:
abandoning this approach for now we don't really have need for a generic course level launch. It's more flexible to build the launch using this libraries python api.

PR #388, opened 2023-06-28
feat: ProctoringConsumer basic launch request

Zacharis278:
Allows a basic LtiResourceLinkRequest launch to be performed by the proctoring consumer class. Since it's the parent class this made sense to me versus initializing a different consumer type based on the launch message. Which, would also be hard to implement anyway.


varshamenon4 (approved):
lgtm!

MichaelRoytman (approved):

PR #394, opened 2023-07-12
feat: added error msg to launch error template

ilee2u:
Ticket: [MST-1931](https://2u-internal.atlassian.net/browse/MST-1931)

### DIFF
--- a/lti_consumer/lti_xblock.py
+++ b/lti_consumer/lti_xblock.py
@@ -1227,8 +1227,6 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
         # Sometimes, it is due to a web crawlers; other times, it is due to actual users of the platform. Regardless,
         # return a 400 response with an appropriate error template.
         try:
-            # ERROR TEST
-            # raise LtiError
             real_user_data = self.extract_real_user_data()
             user_id = self.get_lti_1p1_user_id()
             role = self.role
@@ -1241,7 +1239,8 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
         except LtiError:
             loader = ResourceLoader(__name__)
             error_msg = 'Could not get user data for current request'
-            template = loader.render_django_template('/templates/html/lti_launch_error.html', context={"error_msg": error_msg})
+            template = loader.render_django_template('/templates/html/lti_launch_error.html',
+                                                     context={"error_msg": error_msg})
             return Response(template, status=400, content_type='text/html')
 
         username = None
--- a/lti_consumer/lti_xblock.py
+++ b/lti_consumer/lti_xblock.py
@@ -1235,9 +1235,11 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
             role = LTI_1P1_ROLE_MAP.get(role, 'Student,Learner')
 
             result_sourcedid = self.lis_result_sourcedid
+        # Fails if extract_real_user_data() fails
         except LtiError:
             loader = ResourceLoader(__name__)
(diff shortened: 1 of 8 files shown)


lti_consumer/plugin/views.py
```diff
@@ -138,31 +138,59 @@ def launch_gate_endpoint(request, suffix=None):  # pylint: disable=unused-argume
...
 
     launch_data = get_data_from_cache(lti_message_hint)
     if not launch_data:
-        log.warning(
+        error_msg = (
             f'There was a cache miss trying to fetch the launch data during an LTI 1.3 launch when using the cache'
             f' key {lti_message_hint}. The login hint is {login_hint}.'
         )
-        return render(request, 'html/lti_launch_error.html', status=HTTP_400_BAD_REQUEST)
+        log.warning(error_msg)
```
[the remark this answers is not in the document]
ilee2u: I changed this error message, not exactly sure what change for the others (if at all)

lti_consumer/lti_xblock.py
```diff
@@ -1235,9 +1235,12 @@ def lti_launch_handler(self, request, suffix=''):  # pylint: disable=unused-argu
             role = LTI_1P1_ROLE_MAP.get(role, 'Student,Learner')
 
             result_sourcedid = self.lis_result_sourcedid
+        # Fails if extract_real_user_data() fails
         except LtiError:
             loader = ResourceLoader(__name__)
-            template = loader.render_django_template('/templates/html/lti_launch_error.html')
+            error_msg = 'Could not get user data for current request'
```
Zacharis278, line 1241: this repeats the message we already have on the LtiError object. I think you could just use `except LTIError err:` and than grab `err.message`

Zacharis278 (commented):
one comment otherwise looks pretty good. We'll also need to add a changelog entry and update __init__.py with the new version number before merging.

ilee2u:
Seeing now that codecov fails, but if I recall correctly, it isn't really feasible to run unit tests on some of these error messages due to how dependent on the LTI dataflow each piece of this is.

Zacharis278:
@ilee2u took a look at the missing cases and I think it should be relatively straightforward to add some tests. We can add on to https://github.com/openedx/xblock-lti-consumer/blob/c659e07aa743cb89d8f04a41673ec58e85c9f7fe/lti_consumer/tests/unit/test_lti_xblock.py#L834 that mocks `extract_real_user_data` throwing and error and it looks like we already have error tests for the other missing cases in TestLti1p3LaunchGateEndpoint we just need an invalid context claim launch and another test that mocks out `validate_lti_1p3_launch_data`. Lemme know if you need help on setting these up

lti_consumer/plugin/views.py
```diff
@@ -259,6 +295,9 @@ def launch_gate_endpoint(request, suffix=None):  # pylint: disable=unused-argume
                 custom=dl_params.get('custom')
             )
 
+        # ERROR TEST
```
Zacharis278, line 298: NIT: can you clean these up?

lti_consumer/plugin/views.py
```diff
@@ -234,6 +266,10 @@ def launch_gate_endpoint(request, suffix=None):  # pylint: disable=unused-argume
         # course creators to set up content.
         deep_linking_content_item_id = launch_data.deep_linking_content_item_id
 
+        # ERROR TEST
+        # launch_data.message_type = 'LtiDeepLinkingRequest'
+        # lti_consumer.dl = True
+        # user_role = 'potato'
```
Zacharis278, line 272: same here

Zacharis278 (approved):
just some debugging cleanup to remove but otherwise lgtm

PR #395, opened 2023-07-13
fix: removed get_block_for_descriptor_internal call

Zacharis278:
### [MST-2000](https://2u-internal.atlassian.net/browse/MST-2000)

Fix call to `get_block_for_descriptor_internal` that was removed with https://github.com/openedx/edx-platform/pull/32420


alangsto (approved):
LGTM 👍

PR #425, opened 2023-11-06
feanil/some maintenance

feanil:
- chore: Update pylintrc and .editorconfig from edx-lint
- style: Remove useless pylint supression.


MichaelRoytman (approved):
Thank you!

PR #516, opened 2024-11-25
docs: Add PII annotation

bmtcril:
This class was just missing an annotation.


kdmccormick (approved):
Consider bumping the patch version, but not required

feanil (approved):

PR #527, opened 2025-01-15
fix: ensure that LTI 1.3 launches work

alangsto:
## [COSMO-629](https://2u-internal.atlassian.net/browse/COSMO-629)

Changes introduced in v9.13.0 caused LTI 1.3 launches to not function. This was because the `kid` field was missing from both the JWT header sent to the tool, and from the public keyset endpoint. 

This PR adds the `kid` to the JWT header and ensures that the `kid` is also available in the public keyset endpoint. I've also removed some `pyjwkest` related code that is no longer needed.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,10 @@ Please See the `releases tab <https://github.com/openedx/xblock-lti-consumer/rel
 Unreleased
 ~~~~~~~~~~
 
+9.13.1 - 2025-01-15
+-------------------
+* Fix broken LTI 1.3 launch
+
 9.13.0 - 2025-01-08
 -------------------
 * Removed pyjwkset package and replace with pyjwt package
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -18,7 +18,7 @@ Unreleased
 
 9.13.1 - 2025-01-15
 -------------------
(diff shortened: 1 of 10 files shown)


lti_consumer/lti_1p3/key_handlers.py
```diff
@@ -86,13 +86,6 @@ def _get_keyset(self, kid=None):
                 raise exceptions.NoSuitableKeys() from err
             keyset.extend(keys.keys)
 
-        if self.public_key and kid:
```
alangsto, line 89: I removed this as the `_get_keyset` function was never being called with a non-none `kid` value. The comment also suggests this was related to `pyjwkest` verification, but verification is passing using `pyjwt`

schenedx (approved):
👍

CHANGELOG.rst
```diff
@@ -16,6 +16,10 @@ Please See the `releases tab <https://github.com/openedx/xblock-lti-consumer/rel
 Unreleased
 ~~~~~~~~~~
 
+9.13.1 - 2025-01-15
+-------------------
+* Fix broken LTI 1.3 launch
```
MichaelRoytman, line 21: Could you make this a little more descriptive, so that it's clear what the bug was/what was fixed, please?

MichaelRoytman (approved):

PR #535, opened 2025-02-17
chore: Upgrade Python requirements

Faraz32123:
This PR upgrade python requirements and fixes code([failures](https://github.com/openedx/xblock-lti-consumer/actions/runs/13266356476/job/37034706505?pr=531)) after upgrading `pyjwt` from `2.6.0` to `2.10.1`.

### DIFF
--- a/lti_consumer/lti_1p3/key_handlers.py
+++ b/lti_consumer/lti_1p3/key_handlers.py
@@ -55,7 +55,7 @@ class ToolKeyHandler:
                 public_key = algo_obj.prepare_key(public_key)
                 public_jwk = json.loads(algo_obj.to_jwk(public_key))
                 self.public_key = PyJWK.from_dict(public_jwk)
-            except ValueError as err:
+            except jwt.exceptions.InvalidKeyError as err:
                 log.warning(
                     'An error was encountered while loading the LTI tool\'s key from the public key. '
                     'The RSA key could not parsed.'
@@ -148,7 +148,7 @@ class PlatformKeyHandler:
                 private_jwk = json.loads(algo.to_jwk(private_key))
                 private_jwk['kid'] = kid
                 self.key = PyJWK.from_dict(private_jwk)
-            except ValueError as err:
+            except jwt.exceptions.InvalidKeyError as err:
                 log.warning(
                     'An error was encountered while loading the LTI platform\'s key. '
                     'The RSA key could not be loaded.'
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,10 @@ Please See the `releases tab <https://github.com/openedx/xblock-lti-consumer/rel
 Unreleased
 ~~~~~~~~~~
 
+9.13.4 - 2025-03-21
+-------------------
+* fix: update tests after pyjwt version upgrade
+
 9.13.3 - 2025-03-12
 -------------------
 * Allows LTI Consumer blocks to be used in Libraries v2 learning context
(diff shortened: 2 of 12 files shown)


feanil (approved):
@Faraz32123 thanks for the fix, it seems like this could be a breaking change upstream because upstream repos might be catching the other error, can you bump the major version  and we can merge and release this change?

Faraz32123:
> @Faraz32123 thanks for the fix, it seems like this could be a breaking change upstream because upstream repos might be catching the other error, can you bump the major version and we can merge and release this change?

Sure, I'll do that.

Faraz32123:
@feanil can you check this now?

feanil (changes requested):
@Faraz32123 something was bothering me about why we had to change the test exceptions here so I took a look deeper.  It looks like previously we were using exceptions we defined in our code in the  `lti_consumer.lti_1p3.exceptions` module.  So reading more closely, I think what's going on is that some upstream change is causing a different exception than expected in the initialization of the KeyHandlers and we're no longer catching it correctly.

I think rather than changing the tests here, we need to figure out what is causing the class initialization code to not catch the new exception.  I'm looking at https://github.com/openedx/xblock-lti-consumer/blob/e25638ad37932b995cc1eef658975b2685f0eeb4/lti_consumer/lti_1p3/key_handlers.py#L51-L63 for one example.  In this case, it's only catching a `ValueError` and maybe that was what was being thrown before and now it's more precise?

I think it would be good to see haw the exception is raised with the old version of the library and see if we can make sure we capture the same intended behavior.

Faraz32123:
👍 nice catch. Indeed with newer version, in our code we were not catching newer exception.

I have tried with previous version `pyjwt==2.6.0`, we were getting this exception `ValueError('Unable to load PEM file. See https://cryptography.io/en/latest/faq/#why-can-t-i-import-my-pem-file for more details. MalformedFraming')` and that was caught by `except ValueError as err:` correctly. 
but with version `pyjwt==2.10.1`, exception has been changed and `except ValueError as err:` wasn't catching it. Newer exception is `InvalidKeyError('Could not parse the provided public key.')`.
I have changed the code accordingly.

Thanks @feanil, can you check this again for me.

lti_consumer/lti_1p3/key_handlers.py
```diff
@@ -55,12 +55,12 @@ def __init__(self, public_key=None, keyset_url=None):
                 public_key = algo_obj.prepare_key(public_key)
                 public_jwk = json.loads(algo_obj.to_jwk(public_key))
                 self.public_key = PyJWK.from_dict(public_jwk)
-            except ValueError as err:
+            except jwt.exceptions.InvalidKeyError as err:
                 log.warning(
                     'An error was encountered while loading the LTI tool\'s key from the public key. '
                     'The RSA key could not parsed.'
                 )
-                raise exceptions.InvalidRsaKey() from err
```
feanil, line 63: I think we don't want to change this,  This `InvalidRsaKey` exception is one we define in this library.  You can think of it as a contract between this library and services that use the library.  ie. edx-platform is going to expect this exception in it's code and handle it however it needs to in its application logic.  If we change it to the upstream exception, this would actually be a breaking change and we would need to update handling code downstream as well.   We made our own exception classes to avoid that sort of thing.

feanil (changes requested):
Sorry for the slow review on this.  I think the change is even simpler and we shouldn't need to update the tests.

lti_consumer/lti_1p3/key_handlers.py
Faraz32123: Yes, this makes sense as we are catching the correct exception and we can raise the same previous exception which edx-platform or any other service expects. 👍

feanil (approved):

feanil:
@Faraz32123 thanks so much for iterating on this with me!

PR #551, opened 2025-04-14
feat: add django 5.2 support

### LINKED ISSUE #549: Add Django 5.2 support to xblock-lti-consumer
## Context
Per the Open edX roadmap, this library has been identified as a Django dependency that needs to be updated to support Django 5.2. This update is part of routine platform maintenance and is a prerequisite for upgrading the services that use this library. All Django-dependent libraries must be updated by May 14, 2025, as service upgrades cannot proceed until all library dependencies are compatible.

## Required Actions
1. [ ] Add Django 5.2 to the test matrix
2. [ ] Fix compatibility issues
3. [ ] Update documentation
4. [ ] Release to PyPI

## Resources
- [Django 5.2 release notes](https://docs.djangoproject.com/en/5.2/releases/5.2/)
- [Django Upgrade Guide](https://docs.djangoproject.com/en/5.2/howto/upgrade-version/)
- [django-upgrade tool](https://github.com/adamchainz/django-upgrade) - Automatically upgrade your Django code
- [Django Upgrade Wiki Page](https://openedx.atlassian.net/wiki/spaces/AC/pages/4914774042/Django+5.2+Upgrade)
- For assistance: [#wg-maintenance](https://openedx.slack.com/archives/C03R320AFJP) on Slack

Faraz32123:
In this PR:
- add django 5.2 support
- fix make upgrade command to compile test.txt correctly and Let tox control the Django version for tests, also upgrade requirments again after the fix.
  - This will fix this: https://github.com/openedx/xblock-lti-consumer/actions/runs/14445151146/job/40503912866?pr=551
- fix timezone import, as utc attribute was removed in Django 5.0.
  - This will fix this: https://github.com/openedx/xblock-lti-consumer/actions/runs/14446482262/job/40508199546?pr=551
- close #549

### DIFF
--- a/lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
+++ b/lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
@@ -1,7 +1,7 @@
 """
 Serializers for LTI-related endpoints
 """
-from django.utils import timezone
+from datetime import timezone
 from rest_framework import serializers, ISO_8601
 from rest_framework.reverse import reverse
 from opaque_keys import InvalidKeyError
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
         python-version:
         - '3.11'
         - '3.12'
-        toxenv: [django42, quality]
+        toxenv: [django42, django52, quality]
 
     steps:
     - uses: actions/checkout@v2
@@ -38,8 +38,8 @@ jobs:
       run: tox
 
     - name: Run Coverage
-      if: matrix.python-version == '3.11' && matrix.toxenv=='py311-django42'
-      uses: codecov/codecov-action@v1
+      if: matrix.python-version == '3.12' && matrix.toxenv=='django52'
+      uses: codecov/codecov-action@v5
       with:
         flags: unittests
         fail_ci_if_error: true
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,10 @@ Please See the `releases tab <https://github.com/openedx/xblock-lti-consumer/rel
 Unreleased
 ~~~~~~~~~~
 
+9.14.0 - 2025-04-14
+--------------------
+* Added support for Django 5.2
+
 9.13.4 - 2025-03-21
 -------------------
 * fix: update tests after pyjwt version upgrade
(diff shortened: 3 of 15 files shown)


lti_consumer/lti_xblock.py
```diff
@@ -1093,7 +1093,7 @@ def is_past_due(self):
             close_date = due_date + self.graceperiod
         else:
             close_date = due_date
-        return close_date is not None and timezone.now() > close_date
+        return close_date is not None and datetime.now(timezone.utc) > close_date
```
feanil, line 1096: timezone.now is still valid, it's just the django.utils.timezone.utc alias to datetime.timezone.utc is removed.  Don't use datetime.now because django could set the timezone to a different zone than utc and we still want to support that.

lti_consumer/tests/unit/plugin/test_views_lti_ags.py
```diff
@@ -37,7 +36,7 @@ def setUp(self):
             'lti_1p3_tool_public_key': self.public_key,
             'lti_advantage_ags_mode': 'programmatic',
             # xblock due date related attributes
-            'due': timezone.now(),
+            'due': datetime.now(timezone.utc),
```
feanil, line 39: Again, timezone.now is still valid you don't need to make this change.

lti_consumer/lti_1p3/extensions/rest_framework/serializers.py
```diff
@@ -1,7 +1,7 @@
 """
 Serializers for LTI-related endpoints
 """
-from django.utils import timezone
+from datetime import timezone
```
feanil, line 4: I think this is the only timezone related change that is actually necessary.

feanil (changes requested):

lti_consumer/lti_xblock.py
Faraz32123: can u please check these CI failures: https://github.com/openedx/xblock-lti-consumer/actions/runs/14447517599/job/40511662956

- `timezone` is no more in `django.utils` for `django52`
- That's why imported `timezone` from `datetime`, but `datetime.timezone` doesn't have `now` attr. So I had to use `datetime.now(timezone.utc)`. I can remove `timezone.utc` from it and use datetime.now() to get different timezone if needed. Let me know what you think.

Let me know if I am missing something 🧐

feanil: Right, so `django.utils.timezone.now()` still exists.  It's only, `ddjango.utils.timezone.utc` that was removed.  So you only need to fix the import for timezone.utc, where it is mentioned.  All uses of `django.utils.timezone.now` are fine.

Faraz32123: Make sense 👌. I'll work on it soon.

Issue #552, opened 2025-04-15
Remove pyjwkest dependency from xblock-lti-consumer


iamsobanjaved:
The usage of the `pyjwkest` library was removed and replaced with `pyjwt` in https://github.com/openedx/xblock-lti-consumer/pull/349. However, the `pyjwkest` dependency itself was not removed in that PR.

This issue tracks the cleanup of that unused dependency. Please remove `pyjwkest` from the project and verify that there are no remaining imports or references to it in the codebase.

### Tasks

 - [ ] Remove `pyjwkest` from dependencies 
 - [ ] Search for and remove any leftover imports or usage
- [ ] Make sure CI is green and nothing is broken

jristau1984:
This was previously attempted and ran into issues.  Please work this locally and ensure there is ample review of any PRs before merging any changes.  Note that there are several related tickets across different repos for this replacement.

Faraz32123:
closing this issue in favor of #563

PR #572, opened 2025-08-06
chore: update version to 9.14.2

ktyagiapphelix2u:
# Description

* Deprecation/Removal of pyjwkest
* Various version updates

**Related PR**
https://github.com/openedx/xblock-lti-consumer/pull/563 
 
**Related Issues**
https://github.com/openedx/xblock-lti-consumer/issues/552
https://github.com/edx/edx-arch-experiments/issues/186
 
 **More Context**
This also will need a tag/release created.

Why we are need it.

If we don't bump the version, other dependencies such as edx-platform [Link To Package Reference](https://github.com/openedx/edx-platform/blob/13944afc91ad11ae5b279db0b07551bdeade3931/requirements/edx/base.txt#L602)
and few others will bring pyjwkest in subdependencies.

As the package is removed from here, we also don't need it in other repos and hence it needs to be removed which can be done with new version push.


Faraz32123:
> This also will need a tag/release created.
> 
> Adding a bit more context as to why we are need it.
> 
> If we don't bump the version, other dependencies such as edx-platform [Link To Package Reference ](https://github.com/openedx/edx-platform/blob/13944afc91ad11ae5b279db0b07551bdeade3931/requirements/edx/base.txt#L602) and few others will bring pyjwkest in subdependencies.
> 
> As the package is removed from here, we also don't need it in other repos and hence it needs to be removed which can be done with new version push.

It makes sense. We can add this in PR description OR commit message.

Issue #594, opened 2025-12-03
LTI XBlock Reusability Bug: Copy/Paste, Libraries, and Cross-Course Import Fail Due to DB-Stored Settings


usmanpm:
**Summary**

LTI Consumer (LTI XBlock) has a systemic configuration storage issue that breaks reusability in multiple workflows: copy/paste, use in content libraries, and import/export into different courses. The root cause is that configuration is split between the Studio/module store and a separate database table, and the DB-backed configuration is not kept in sync across these operations.

### Affected scenarios

- Duplicating/copy-pasting an LTI XBlock
- Using LTI XBlocks in content libraries
- Importing a course that contains LTI XBlocks into a different course ID

Note: Export + import into the same course ID appears to work, but this is misleading (explained below).

### Root cause

Per Feanil’s investigation into the LTI Consumer code:

- The LTI block stores:
  - Some configuration in the normal Studio/module store backend
  - Some configuration in a separate database table created by the block.
- During copy/paste, library usage, and import/export:
  - The Studio/module store data is copied and behaves correctly.
  - The separate DB table entries are not synchronized with these operations.
- As a result, after these operations the block’s Studio-side configuration points to nothing or the wrong thing in the external DB, and the LTI block breaks.

### Clarification on import/export behavior

- When a course is exported and re-imported into the same course ID, the LTI block appears to work.
- Reason: the backing DB configuration is keyed by course ID, so re-importing into the same course ID “auto-connects” to the existing DB row.
- When importing into a new course ID, that auto-connection does not exist, and the block is expected to fail.

This explains why export/import can appear to succeed in some tests while the underlying bug still exists.

### Proposed direction for a fix

- Move / duplicate problematic settings that live only in the external DB table into the Studio/module store representation of the block.
- There are already some settings that are tracked in both places and kept in sync; the idea is to treat the problematic settings similarly.
- Key questions to validate:
  - What are the implications of moving or duplicating these settings into Studio/module store?
  - What might this change break (e.g., existing deployments, assumptions in other parts of the system, external integrations)?

This area has a lot of externalities, so further testing and validation is required before finalizing the approach.

### Value, effort, and timeline

- Fixing this issue would:
  - Enable reliable copy/paste, library usage, and import/export across courses for LTI blocks.
  - Remove the need to recommend the “LTI store” workaround

PR #606, opened 2026-02-16
feat: decouple xblock and lti config to fix copy-paste

navinkarkera:
**Description:**

Decouples config ID and xblock usage_key such that the same configuration can be used by more than one xblock.

* [x] We add a new table that links xblock to a specific lti configuration object in database.
* [x] Create a migration that generates entries in this new table based on existing lti configurations.
* [x] Add `config_id` field to LTI xblock so that copy-paste works, when a lti xblock with config id is pasted it just creates a new link entry in the new table pointing to the existing lti configuration.
* [ ] **Limitation:** This method only works in the same instance, so importing and exporting the course in new instance that doesn't have the related lti configuration already in place will not work.
* [x] This PR also adds signal handlers to handle delete or duplicate events related to lti xblocks. These handlers delete unlinked configuration objects and remove linked entries of deleted xblocks.
* [x] Create a migration script that removes location and direct links to LTI configuration from different tables.

**Related to:** https://github.com/openedx/xblock-lti-consumer/issues/594

**Concerns:**

We need to extensively test this PR with all LTI features including proctoring, deep linking, scoring etc. The existing architecture tightly couples config_id and location, and this resulted in urls with config_id that only link to a single xblock, however this PR decouples config_id and xblocks resulting in multiple xblocks being related to a single config_id. I have tried to narrow down to single xblock using `resource_link_id` wherever possible but defaulted to first instance of link entry in cases where it is not possible (In theory this should work as the linked configuration will be correct in any case).

**Questions/Suggestions:**

* [x] What should happen when author starts editing pasted/duplicated xblock settings? Specifically `Configuration Type` field, as this change should be same in database as well as xblock to make sure that it works properly.
  Problem: We cannot update the configuration in database as it is shared between original and the copy.
  Possible solution: When the author changes `config_type`, we create a new configuration entry for this xblock with a new config_id.

**Test instructions:**

* I have created a sandbox PR: https://github.com/openedx/openedx-platform/pull/38144 with latest changes.
* [This page](https://github.com/openedx/openedx-platform/pull/38144/checks?check_run_id=66949617415) contains all details required to login into the sandbox.
* Test all LTI related features there.


openedx-webhooks:
Thanks for the pull request, @navinkarkera!


This repository is currently maintained by `@Faraz32123`.


Once you've gone through the following steps feel free to tag them in a comment and let them know that your changes are ready for engineering review.

<details><summary>:radio_button: Get product approval</summary>

If you haven't already, [check this list](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3875962884/How+to+submit+an+open+source+contribution+for+Product+Review#Does-my-contribution-require-Product-Review%3F) to see if your contribution needs to go through the product review process.

- If it does, you'll need to submit a product proposal for your contribution, and have it reviewed by the [Product Working Group](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3449028609/Product+Working+Group).
    - This process (including the steps you'll need to take) is documented [here](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3875962884/How+to+submit+an+open+source+contribution+for+Product+Review#Product-Review-Process).
- If it doesn't, simply proceed with the next step.
</details>

<details><summary>:radio_button: Provide context</summary>

To help your reviewers and other members of the community understand the purpose and larger context of your changes, feel free to add as much of the following information to the PR description as you can:

- Dependencies
  > This PR must be merged before / after / at the same time as ...
- Blockers
  > This PR is waiting for OEP-1234 to be accepted.
- Timeline information
  > This PR must be merged by XX date **because** ...
- Partner information
  > This is for a course on edx.org.
- Supporting documentation
- Relevant [Open edX discussion forum](https://discuss.openedx.org/) threads
</details>





<details><summary>:radio_button: Get a green build</summary>

If one or more checks are failing, continue working on your changes until this is no longer the case and your build turns green.
</details><details>

<summary>:radio_button: Update the status of your PR</summary>

Your PR is currently marked as a draft.  After completing the steps above, update its status by clicking "Ready for Review", or removing "WIP" from the title, as appropriate.
</details>



---

<details><summary>Where can I find more information?</summary>

If you'd like to get more details on all aspects of the review process for open source pull requests (OSPRs), check out the following resources:

- [Overview of Review Process for Community Contributions](https://docs.openedx.org/en/latest/developers/references/developer_guide/process/FAQ-about-pull-requests.html)
- [Pull Request Status Guide](https://docs.openedx.org/en/latest/developers/references/developer_guide/process/pull-request-statuses.html)
- [Making changes to your pull request](https://docs.openedx.org/en/latest/documentors/how-tos/make_changes_to_your_pull_request.html)
</details>

<details><summary>When can I expect my changes to be merged?</summary>

Our goal is to get community contributions seen and reviewed as efficiently as possible.

However, the amount of time that it takes to review and merge a PR can vary significantly based on factors such as:

- The size and impact of the changes that it introduces
- The need for product review
- Maintenance status of the parent repository

:bulb: *As a result it may take up to several weeks or months to complete a review and merge your PR.*
</details>

navinkarkera:
@rpenido There are some problems that I can think of with that approach:

* The urls currently have `config_id` which is used to fetch the configuration objects. Example: `https://pr-38144-885846.axim-sandboxes.opencraft.hosting/api/lti_consumer/v1/token/102e736f-cf9d-4836-a315-900a45bb73b1`.
  * The UUID at the end is used by the platform to get configuration.
  * If we remove the table, we'll need to fetch the xblock based on the config_id field and then use it. I think it is not very efficient to do so. (But I might be wrong though).
  * Before config_id was introduced, usage_key was used in the URLs, so if a author copies and pastes such a block it will always point to the original block and ignore the copied block configuration. (We have same issue here, but the linked configuration object is same). 

We can probably overcome above issues but it will take a lot more time to implement. However another problem is that the decoupling of LTI config and xblock was done to allow launching LTI outside of xblock context, see https://github.com/openedx/xblock-lti-consumer/blob/master/docs/decisions/0007-lti-launch-decoupling.rst. So removing LTI config table would undo the decision.

Issue #605, opened 2026-02-12
Grade passback not working with H5P


ayub02:
### Background

H5P is a tool that allows users to create interactive content and then embed it into their LMSes using LTI 1.1 and 1.3. 

This github issue is about a problem i've encountered while testing H5P integration with [Open edX Ulmo sandbox](https://apps.ulmo.openedx.io/learner-dashboard/) via LTI 1.3 + Advantage. 

### Problem

When a user (learner) attempts as assessment in H5P tool from within Open edX, the tool displays a warning "Your gradebook did not confirm receiving this attempt". 

<img width="819" height="365" alt="Image" src="https://github.com/user-attachments/assets/f5ba87ca-ca9c-4f46-a72a-ea513688e814" />

As expected, the grades in Progress tab do not change which confirms that the platform was unable to receive grades from the tool.

I shared a HAR file with H5P technical support team (attached). Correspond screen recording is attached.

H5P then shared a response to HAR file from their devs which i'm copying below (see email conversation for context):

Here's what our devs found: 
 
We are making the request to retrieve the line items: 
 
GET "https://ulmo.openedx.io/api/lti_consumer/v1/lti/3/lti-ags?resource_link_id=block-v1:edx+LTI101+2026+type@lti_consumer+block@ff4f12d57abc480eb2a3ec89eacb165a"
Authorization: "Bearer {$access_token}"
Accept: "application/vnd.ims.lis.v2.lineitemcontainer+json"
 
The result we get is:
 
500 Internal Server Error
The response body is the default HTML page for the OpenEdX front page. No specifics of the error are visible.
 
We believe that you'll need to check your application or web server error log to determine what’s wrong here; it’s not exposed to us.

[har file as text.txt](https://github.com/user-attachments/files/25255242/har.file.as.text.txt)

https://github.com/user-attachments/assets/6302907c-a77a-40a4-b352-dfd69c74318e

[Conversation with H5P.pdf](https://github.com/user-attachments/files/25255071/Conversation.with.H5P.pdf)

### Goal

Goal of this ticket is to identify the exact issue and suggest remedies. 

### Steps to reproduce

Please follow the integration procedure outlined in this document. Connect with @ayub02 for any help.

https://openedx.atlassian.net/wiki/spaces/COMM/pages/5542707201/Guide+H5P+1.3+integration+Needs+update

navinkarkera:
[...]

PR #607, opened 2026-02-18
fix: quote resource_link_id properly as it is used in url by tools

navinkarkera:
The resource-link-id is used by the tool in a GET request (included in the query params) like so:

```
GET "https://ulmo.openedx.io/api/lti_consumer/v1/lti/3/lti-ags?resource_link_id=block-v1:edx+LTI101+2026+type@lti_consumer+block@ff4f12d57abc480eb2a3ec89eacb165a"
```

The `+` symbols are removed when the platform receives this request and the resource_link_id cannot be parsed with below error:

```
opaque_keys.InvalidKeyError: <class 'opaque_keys.edx.locator.CourseLocator'>: UNIX UNX1 2026_T1 type@lti_consumer block@a07db98b8d074e4b977ea4cbfbbd20a8
```

This PR quotes the resource_link_id properly to avoid this issue.

Issue: https://github.com/openedx/xblock-lti-consumer/issues/605
`Private-ref`: https://tasks.opencraft.com/browse/FAL-4327

**Test instructions:**

* Follow steps described in https://openedx.atlassian.net/wiki/spaces/COMM/pages/5542707201/Guide+H5P+1.3+integration+Needs+update
* Before saving the xblock in step 4, update following settings
  * Deep linking: True
  * Deep linking launch url: Same as launch url
  * LTI Assignment and Grades Service: programmatic (**Note:** This is required to reproduce the bug, but even with this fix the grade pass back still fails. You can set this to `declarative` to make grade pass back work but the tool doesn't send the above GET request in this case)
  * Scored: True
* Go to <studio_url>/admin/lti_consumer/courseallowpiisharinginltiflag/ and enable pii sharing for your course.
* Go back to the xblock settings, you'll notice three new fields:
  * Request users' username
  * Request users' full name
  * Request users' email
  Set all of them to True
* Save the xblock
* Click on deep linking lauch - configure tool
* Select any problem from h5p and insert it.
* Publish the block and test it in LMS.

**The bug only appears if you set the `LTI Assignment and Grades Service` to `programmatic` the first time without saving the block even once.** 

**Reason:** The [LtiAgsLineItem](https://github.com/open-craft/xblock-lti-consumer/blob/99a3bbebdc848e651a47c1b0547a98eb035dfb82/lti_consumer/models.py#L470-L503) is not created when `programmatic` is selected and the tool for some reason calls the above GET request and our platform fails with 500 error.

**Note:** Even after fixing this 500 error the grade pass back fails due to missing `LtiAgsLineItem` when `programmatic` is selected, you need to select `declarative` for it to work.

### DIFF
--- a/lti_consumer/tests/unit/test_lti_xblock.py
+++ b/lti_consumer/tests/unit/test_lti_xblock.py
@@ -1896,7 +1896,7 @@ class TestLtiConsumer1p3XBlock(TestCase):
             "user_id": 1,
             "user_role": "instructor",
             "config_id": config_id_for_block(self.xblock),
-            "resource_link_id": str(urllib.parse.quote(str(self.xblock.scope_ids.usage_id))),
+            "resource_link_id": urllib.parse.quote(str(self.xblock.scope_ids.usage_id)),
             "external_user_id": "external_user_id",
             "launch_presentation_document_target": "iframe",
             "message_type": "LtiResourceLinkRequest",
--- a/lti_consumer/lti_xblock.py
+++ b/lti_consumer/lti_xblock.py
@@ -1669,7 +1669,7 @@ class LtiConsumerXBlock(StudioEditableXBlockMixin, XBlock):
             user_role=self.role,
             config_id=config_id,
             # resource_link_id is used in the url params by the tool, so it should be url encoded.
-            resource_link_id=str(urllib.parse.quote(str(location))),
+            resource_link_id=urllib.parse.quote(str(location)),
             external_user_id=self.external_user_id,
             preferred_username=username,
             name=full_name,


lti_consumer/tests/unit/test_lti_xblock.py
```diff
@@ -1896,7 +1896,7 @@ def test_get_lti_1p3_launch_data(
             "user_id": 1,
             "user_role": "instructor",
             "config_id": config_id_for_block(self.xblock),
-            "resource_link_id": str(self.xblock.scope_ids.usage_id),
+            "resource_link_id": str(urllib.parse.quote(str(self.xblock.scope_ids.usage_id))),
```
rpenido: Nit: remove extra `str` cast
```suggestion
            "resource_link_id": urllib.parse.quote(str(self.xblock.scope_ids.usage_id)),
```

rpenido (approved):
LGTM :+1:
Thank you for your work, @navinkarkera!

- [x] I tested this using the instructions from the PR
- [x] I read through the code
- [x] Includes documentation

PR #627, opened 2026-03-19
feat: LTI 1.3 Passport Refactor + Database Cleanup Support 

navinkarkera:
#### Split LTI 1.3 Configuration into Passport Model

- Introduce `Lti1p3Passport` model to centralize LTI 1.3 keys and credentials
- Move `lti_1p3_internal_private_key`, `lti_1p3_internal_private_key_id`, `lti_1p3_internal_public_jwk`, `lti_1p3_client_id`, `lti_1p3_tool_public_key`, and `lti_1p3_tool_keyset_url` fields from `LtiConfiguration` to `Lti1p3Passport`
- Add `ForeignKey` relationship from `LtiConfiguration` to `Lti1p3Passport`
- Implement passport-based key generation and retrieval
- Add `clean()` validation to `Lti1p3Passport` to ensure at least one of `lti_1p3_tool_public_key` or `lti_1p3_tool_keyset_url` is set
- Update validation in `LtiConfiguration.clean()` to check for passport presence instead of tool key fields
- Refactor `get_or_create_local_lti_config()` to handle passport creation and sync block/passport key configurations
- Update API endpoints to work with passport ID instead of configuration ID
- Add admin interface for `Lti1p3Passport` model
- Refactor `access_token_endpoint` and `public_keyset_endpoint` to use passport ID
- Update API and views to work with the new passport model
- Generate migration to remove fields from `LtiConfiguration` table
- Update data migration to copy existing configurations to the new Passport model
- Update XBlock to store passport ID instead of config ID
- Fix copy-paste issue in `resource_link_id` generation

#### Support Database Cleanup for Deleted Blocks
- New signal handlers to:
    - Automatically delete orphaned configurations during `pre_item_delete` (block/children).
    - Purge unused passport objects from the LTI 1.3 Passport model.

### Related

* Depends on https://github.com/openedx/openedx-platform/pull/38289
* Related to: https://github.com/openedx/xblock-lti-consumer/issues/594\
* Previous iteration: https://github.com/openedx/xblock-lti-consumer/pull/606

### Test instructions

* Use [sandbox](https://github.com/openedx/openedx-platform/pull/38289/checks?check_run_id=71245177950) to test.
* Create a working 1.3 LTI xblock.
* Copy and paste the xblock.
* Make sure both xblocks are working.
* Go to <studio_url>/admin/lti_consumer/ and check configuration and passport entries.
* Both configuration should use a single passport.
* Now go and modify `lti_1p3_tool_keyset_url` in one of the xblocks and save. This should create a new passport entry instead of modifying the original one to avoid changing other block.
* Finally test all LTI 1.3 features to make sure that we have not broken existing features.


openedx-webhooks:
Thanks for the pull request, @navinkarkera!


This repository is currently maintained by `@Faraz32123`.


Once you've gone through the following steps feel free to tag them in a comment and let them know that your changes are ready for engineering review.

<details><summary>:radio_button: Get product approval</summary>

If you haven't already, [check this list](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3875962884/How+to+submit+an+open+source+contribution+for+Product+Review#Does-my-contribution-require-Product-Review%3F) to see if your contribution needs to go through the product review process.

- If it does, you'll need to submit a product proposal for your contribution, and have it reviewed by the [Product Working Group](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3449028609/Product+Working+Group).
    - This process (including the steps you'll need to take) is documented [here](https://openedx.atlassian.net/wiki/spaces/COMM/pages/3875962884/How+to+submit+an+open+source+contribution+for+Product+Review#Product-Review-Process).
- If it doesn't, simply proceed with the next step.
</details>

<details><summary>:radio_button: Provide context</summary>

To help your reviewers and other members of the community understand the purpose and larger context of your changes, feel free to add as much of the following information to the PR description as you can:

- Dependencies
  > This PR must be merged before / after / at the same time as ...
- Blockers
  > This PR is waiting for OEP-1234 to be accepted.
- Timeline information
  > This PR must be merged by XX date **because** ...
- Partner information
  > This is for a course on edx.org.
- Supporting documentation
- Relevant [Open edX discussion forum](https://discuss.openedx.org/) threads
</details>





<details><summary>:radio_button: Get a green build</summary>

If one or more checks are failing, continue working on your changes until this is no longer the case and your build turns green.
</details><details>


---

<details><summary>Where can I find more information?</summary>

If you'd like to get more details on all aspects of the review process for open source pull requests (OSPRs), check out the following resources:

- [Overview of Review Process for Community Contributions](https://docs.openedx.org/en/latest/developers/references/developer_guide/process/FAQ-about-pull-requests.html)
- [Pull Request Status Guide](https://docs.openedx.org/en/latest/developers/references/developer_guide/process/pull-request-statuses.html)
- [Making changes to your pull request](https://docs.openedx.org/en/latest/documentors/how-tos/make_changes_to_your_pull_request.html)
</details>

<details><summary>When can I expect my changes to be merged?</summary>

Our goal is to get community contributions seen and reviewed as efficiently as possible.

However, the amount of time that it takes to review and merge a PR can vary significantly based on factors such as:

- The size and impact of the changes that it introduces
- The need for product review
- Maintenance status of the parent repository

:bulb: *As a result it may take up to several weeks or months to complete a review and merge your PR.*
</details>

navinkarkera:
@ayub02 Please use the [sandbox](https://github.com/openedx/openedx-platform/pull/38192/checks?check_run_id=67783278470) to test.

feanil (commented):
@navinkarkera I like the approach of adding a new model for storing lti credentials independent of the blocks, this would obviate the need for external storage and make it easier to re-use storage.  I've got a few questions specific to the implementation but I think this is the idea that we should try to land.

@ayub02 a question for you: Should import/export work for LTI blocks from one open edx instance to another? This has not worked before but this change will not really fix that either.

lti_consumer/utils.py
```diff
@@ -369,7 +369,7 @@ def model_to_dict(model_object, exclude=None):
 
         # Remove private and excluded fields.
         for key in list(object_fields):
-            if key.startswith('_') or key in exclude:
+            if key.startswith('_') or key in exclude or (include and key not in include):
```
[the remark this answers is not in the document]
navinkarkera: Just to allow us to include some fields instead of excluding lot of fields incase we only need few of them.

lti_consumer/signals/signals.py
```diff
@@ -82,4 +84,69 @@ def publish_grade_on_score_update(sender, instance, **kwargs):  # pylint: disabl
             raise exc
 
 
+@receiver(post_save, sender=LtiConfiguration, dispatch_uid='create_lti_1p3_passport')
+def create_lti_1p3_passport(sender, instance: LtiConfiguration, **kwargs):  # pylint: disable=unused-argument
+    instance.get_or_create_lti_1p3_passport()
```
feanil, line 91: When you call this function as a post-save function for `LtiConfiguration` for it also calls `lti_configuration.save()` conditionally.  So you have a situation where we get a double save and a double call to the get_or_create function at the creation of each new LtiConfiguration instance.  Take a look at the suggestion in the get_or_create function to see how we could avoid the double save firing.

lti_consumer/models.py
```diff
@@ -315,65 +401,99 @@ def sync_configurations(self):
...
+                log.info("Created new LTI 1.3 Passport %s for %s", passport, self)
+                if block:
+                    passport.name = f"Passport of {block.display_name}"
+                    passport.context_key = block.context_id
+                    passport.save()
+                    block.lti_1p3_passport_id = str(passport.passport_id)
+                    block.save()
+                    compat.save_xblock(block)
+            self.lti_1p3_passport = passport
+            self.save()
```
feanil:
```suggestion
            LtiConfiguration.objects.filter(pk=self.pk).update(lti_1p3_passport=passport)
```

The `self.lti_1p3_passport = passport` keeps the in-memory instance in sync. The `update()` writes the FK directly to the DB without going through `save()`, so the signal doesn't re-fire. The sync_configurations() bypass is fine here since we're only updating this FK field.  But we should add a comment explaining this here.

lti_consumer/migrations/0023_lti1p3passport_context_key_lti1p3passport_name_and_more.py
```diff
@@ -0,0 +1,58 @@
...
+        if not configuration.location:
+            continue
+        try:
+            block = load_enough_xblock(configuration.location)
+        except Exception as e:  # pylint: disable=broad-exception-caught
+            print(f"Error loading xblock {configuration.location}: {e}")
+        passport = configuration.lti_1p3_passport
+        if passport and not passport.name:
+            passport.name = f"Passport of {block.display_name}"
+            passport.context_key = block.context_id
```
feanil: If the try/catch above could throw an exception in which case `block` might not be set.  In that case we'll get further exceptions. Do we need a continue in the `Exception` clause so that we don't error out here?

lti_consumer/api.py
feanil: This save() fires on every call to get_or_create_local_lti_config, which includes every LTI 1.3 author view render in Studio (via get_lti_1p3_launch_data → config_id_for_block) and every LTI launch. That means an author simply opening a block to check the client ID triggers a DB write even when nothing has changed.                                                             
                                                                                                                                
The fix is to only save when something actually changed:
```                                                         
  dirty = (                                               
      lti_config.config_store != config_store or                                                                                
      lti_config.external_id != block.external_config or
      lti_config.version != lti_version or                                                                                      
      lti_config.lti_1p3_passport != passport                                                                                   
  )                                                                                                                             
  if dirty:                                                                                                                     
      lti_config.save()                                                                                                         
```

The longer-term fix would be to move the reconciliation logic into an override of submit_studio_edits so it only fires when an author actually saves changes in Studio. That doesn't cover all cases though (imports, duplicates) so a dirty-check fallback would still be needed. Worth a follow-up ticket.

lti_consumer/signals/signals.py
```diff
@@ -82,4 +84,69 @@ def publish_grade_on_score_update(sender, instance, **kwargs):  # pylint: disabl
...
+
+
+@receiver(SignalHandler.pre_item_delete if SignalHandler else [])
```
feanil, line 94: I don't see a pre_item_delete signal in the [modulestore SignalHandler](https://github.com/openedx/openedx-platform/blob/master/xmodule/modulestore/django.py#L187-L191). Is this defined elsewhere?

lti_consumer/api.py
```diff
@@ -6,25 +6,28 @@
...
+    get_lti_deeplinking_content_url,
 )
-from .filters import get_external_config_from_filter
 
+log = logging.getLogger(__name__)
 
-def _get_or_create_local_lti_config(lti_version, block_location,
-                                    config_store=LtiConfiguration.CONFIG_ON_XBLOCK, external_id=None):
+
+def get_or_create_local_lti_config(lti_version, block, config_store=LtiConfiguration.CONFIG_ON_XBLOCK):
```
feanil: nit: Does this need to be a public function?  Seems like it's only called from a different private function and should remain an internal function (prefix with an underscore?)

lti_consumer/migrations/0023_lti1p3passport_context_key_lti1p3passport_name_and_more.py
navinkarkera: @feanil Nice catch! I was fighting with sandbox deployment, most probably this is the issue.

lti_consumer/signals/signals.py
navinkarkera: It is part of this PR: https://github.com/openedx/openedx-platform/pull/38192, I have added this in the description as a dependency.

lti_consumer/api.py
navinkarkera: Yes, at some point I used it outside but not needed anymore.

lti_consumer/signals/signals.py
navinkarkera: Nice catch!

lti_consumer/models.py
navinkarkera: Same here!

lti_consumer/api.py
navinkarkera: Yes, even I am not happy about the DB changes on each render.

For now I have refactored the whole function with some help from AI. I still need to test it a bit more.

feanil:
@navinkarkera let me know when you want me to take another Pass at reviewing this.

navinkarkera:
@feanil Yes, it is ready for another round.

lti_consumer/signals/signals.py
```diff
@@ -82,4 +85,87 @@ def publish_grade_on_score_update(sender, instance, **kwargs):  # pylint: disabl
...
+    """
+    Duplicate LTI configuration from the source to the target xblock.
+    """
+    xblock_data = kwargs.get("xblock_info", None)
+    if not xblock_data or not isinstance(xblock_data, DuplicatedXBlockData):
+        log.error("Received null or incorrect data for event")
+        return
+
+    src_lti_config = LtiConfiguration.objects.get(location=str(xblock_data.source_usage_key))
+    copy = src_lti_config
```
feanil: Nit: The `pk=None` will make this into a new object but it's a bit of a trick. Add a comment to say we're using it to duplicate this object without having to enumerate all the fields here and making this function more brittle.

There's also a risk that if we introduce a new generated key in the future like `config_id` that this function would not update that correctly.  This is hypothetical so no need to code to defensively around it unless you can think of an easy way to do so.  Nothing obvious comes to mind for me.

feanil (commented):
PR generally looks good, I'm going to do some more testing of it next week before final approval in-case I find any issues but no major new issues from my review.

What's the status of the openedx-platform PR? Are we ready to review/land that?

lti_consumer/signals/signals.py
navinkarkera: I refactored it to use `model_to_dict` and also handle possible errors. c2c7ec70875b

navinkarkera:
> What's the status of the openedx-platform PR? Are we ready to review/land that?

@feanil Yes.

lti_consumer/migrations/0021_create_lti_1p3_passport.py
```diff
@@ -0,0 +1,67 @@
...
+            include=[
+                'lti_1p3_internal_private_key',
+                'lti_1p3_internal_private_key_id',
+                'lti_1p3_internal_public_jwk',
+                'lti_1p3_client_id',
+                'lti_1p3_tool_public_key',
+                'lti_1p3_tool_keyset_url',
+            ],
+        )
+        if block.config_type == "new":
```
feanil: Ran in to the following issue when I tried to run this migration on my local dev setup:

```
Applying lti_consumer.0021_create_lti_1p3_passport...Failed to copy passport_id for configuration LtiConfiguration object (25): lti_consumer:66fc7df97e3948e0a6b39ebe20f19280                                                                                                                        
Traceback (most recent call last):                                       
  File "/openedx/edx-platform/./manage.py", line 93, in <module>                                                                                                                                                                                                                                       
    execute_from_command_line([sys.argv[0]] + django_args)                                                                                                                                                                                                                                             
  File "/openedx/venv/lib/python3.12/site-packages/django/core/management/__init__.py", line 442, in execute_from_command_line                                                                                                                                                                         
    utility.execute()                                                    
  File "/openedx/venv/lib/python3.12/site-packages/django/core/management/__init__.py", line 436, in execute                                       
    self.fetch_command(subcommand).run_from_argv(self.argv)                                                                                                                                                                                                                                            
  File "/openedx/venv/lib/python3.12/site-packages/django/core/management/base.py", line 420, in run_from_argv                                     
    self.execute(*args, **cmd_options)                                   
  File "/openedx/venv/lib/python3.12/site-packages/django/core/management/base.py", line 464, in execute                                           
    output = self.handle(*args, **options)                                                                                                                                                                                                                                                             
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                                             
  File "/openedx/venv/lib/python3.12/site-packages/django/core/management/base.py", line 111, in wrapper                                           
    res = handle_func(*args, **kwargs)                                   
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                                                                                                                                                                 
  File "/openedx/venv/lib/python3.12/site-packages/django/core/management/commands/migrate.py", line 353, in handle                                                                                                                                                                                    
    post_migrate_state = executor.migrate(                               
                         ^^^^^^^^^^^^^^^^^                               
  File "/openedx/venv/lib/python3.12/site-packages/django/db/migrations/executor.py", line 135, in migrate                                         
    state = self._migrate_all_forwards(                                  
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^                                  
  File "/openedx/venv/lib/python3.12/site-packages/django/db/migrations/executor.py", line 167, in _migrate_all_forwards
    state = self.apply_migration(                                                                                                                  
            ^^^^^^^^^^^^^^^^^^^^^                                                                                                                  
  File "/openedx/venv/lib/python3.12/site-packages/django/db/migrations/executor.py", line 255, in apply_migration
    state = migration.apply(state, schema_editor)                                                                                                  
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                  
  File "/openedx/venv/lib/python3.12/site-packages/django/db/migrations/migration.py", line 127, in apply                                                                                                                                                                                              
    operation.database_forwards(                                         
  File "/openedx/venv/lib/python3.12/site-packages/django/db/migrations/operations/special.py", line 196, in database_forwards                     
    self.code(from_state.apps, schema_editor)                                                                                                                                                                                                                                                          
  File "/mnt/xblock-lti-consumer/lti_consumer/migrations/0021_create_lti_1p3_passport.py", line 37, in create_lti_1p3_passport                                                                                                                                                                         
    if block.config_type == "new":                                                                                                                                                                                                                                                                     
       ^^^^^                                                                                                                                                                                                                                                                                           
UnboundLocalError: cannot access local variable 'block' where it is not associated with a value
```

Couple of things here.

1. we don't handle the case where block doesn't exist because of the exception thrown above that was caught and silenced.  We should re-raise the exception probably.

2. The config this is failing on is as follows. Not sure exactly what's going on yet but I'll have a closer look at this soon.  In the meantime passing it along.:
```
mysql> select id, version, config_store, location, config_id, external_id, lti_1p3_passport_id from lti_consumer_lticonfiguration where id=25;
+----+---------+-----------------+-------------------------------------------------------------------------------+----------------------------------+--------------------------------+---------------------+
| id | version | config_store    | location                                                                      | config_id                        | external_id                    | lti_1p3_passport_id |
+----+---------+-----------------+-------------------------------------------------------------------------------+----------------------------------+--------------------------------+---------------------+
| 25 | lti_1p3 | CONFIG_EXTERNAL | block-v1:lti+lti+lti+type@lti_consumer+block@66fc7df97e3948e0a6b39ebe20f19280 | 757a640e464e4cf99cec3bd788ff6124 | lti_store:local_reference_tool |                NULL |
+----+---------+-----------------+-------------------------------------------------------------------------------+----------------------------------+--------------------------------+---------------------+
1 row in set (0.00 sec)
```

navinkarkera: @feanil Ahh, yes. My bad, it is possible to have empty location in configuration table when the LTI is not under course context.

Updated it: 3a4115062d6b to handle such cases and added tests for these custom migration logic with the help of GPT 5.4

lti_consumer/models.py
```diff
@@ -315,65 +401,102 @@ def sync_configurations(self):
...
+        Create an LTI 1.3 Passport configuration for this course instance.
+        """
+        passport = self.lti_1p3_passport
+        block = None
+        if not passport:
+            if self.location:
+                block = compat.load_enough_xblock(self.location)
+            if self.location and block and block.lti_1p3_passport_id:
+                passport, created = Lti1p3Passport.objects.get_or_create(
+                    passport_id=block.lti_1p3_passport_id,
```
feanil, line 415: In testing locally, this is always creating a new passport instead of re-using the existing one on copy paste.  Not sure fully what's going on yet but continuing to investigate.

navinkarkera: That is weird, I tried the exact flow that you posted on slack, i.e., created a LTI block on master branch and then changed to this branch and ran migration. Next copy and paste the block, it seems to work fine.

Did you happen to copy the block before the migration?

lti_consumer/migrations/0021_create_lti_1p3_passport.py
```diff
@@ -0,0 +1,68 @@
...
+                'lti_1p3_client_id',
+                'lti_1p3_tool_public_key',
+                'lti_1p3_tool_keyset_url',
+            ],
+        )
+        if configuration.location:
+            try:
+                block = load_enough_xblock(configuration.location)
+                block.lti_1p3_passport_id = str(configuration.config_id)
+                block.save()
```
ormsbee: Doing modulestore save operations in a database migration like this is likely going to cause a lot of operational issues, as well as issues around draft/publish workflow. Is there any way to do this by making the XBlock read these values as needed from the DB instead of doing a data migration into modulestore?

navinkarkera: @ormsbee Removed save operations from migration. Instead extended `add_xml_to_node` to include passport id in the xml as suggested by @feanil

lti_consumer/models.py
navinkarkera: @feanil Can you try once now, passport_id should not be now fetched from DB and included in xml on copy.

lti_consumer/migrations/0021_create_lti_1p3_passport.py
```diff
@@ -0,0 +1,65 @@
...
+        passport, _ = Lti1p3Passport.objects.update_or_create(
+            # Use config_id as passport_id to preserve existing urls that use it.
+            passport_id=configuration.config_id,
+            defaults=values,
+        )
+        configuration.lti_1p3_passport = passport
+        configuration.save()
+
+
+def backwards(*_):
```
feanil: We should add a proper backwards migration here.  Dropping all the LTI creds when we reveres this is not great if people find they are having issues with the new version of this block and want to back up until we fix them.

lti_consumer/migrations/0021_create_lti_1p3_passport.py
```diff
@@ -0,0 +1,65 @@
...
+                'lti_1p3_client_id',
+                'lti_1p3_tool_public_key',
+                'lti_1p3_tool_keyset_url',
+            ],
+        )
+        if configuration.location:
+            try:
+                block = load_enough_xblock(configuration.location)
+                if block.config_type == "new":
+                    # Data is stored xblock
```
feanil:
```suggestion
                    # Data is stored on the xblock
```

lti_consumer/migrations/0021_create_lti_1p3_passport.py
```diff
@@ -0,0 +1,65 @@
...
+"""
+This migration will copy config_id from LtiConsumer database to LtiConsumerXBlock.
+
+This will help us link xblocks with LtiConsumer database rows without relying on the location or usage_key of xblocks.
+"""
+from django.db import migrations
+
+
+def create_lti_1p3_passport(apps, _):
+    """Copy config_id from LtiConsumer to LtiConsumerXBlock."""
```
feanil:
```suggestion
    """Create a new LtiPassport entry for each existing LtiConfiguration
    
    We use the `config_id` of the existing LtiConfiguration as the `passport_id` for the newly
    created LtiPassort entries to maintain backwards compatability and to ensure urls built
    from LtiPassoport passport_ids continue to work for previously created LtiConfiguration
    which correspond to existing instances of the LTI Consumer XBlock.
    """
```

lti_consumer/migrations/0021_create_lti_1p3_passport.py
```diff
@@ -0,0 +1,65 @@
+# Generated migration for copying config_id into modulestore from database (Django 6.2)
```
feanil:
```suggestion
# Generated migration for copying config_id to the LtiPassport table
```

lti_consumer/migrations/0021_create_lti_1p3_passport.py
```diff
@@ -0,0 +1,65 @@
...
+"""
+This migration will copy config_id from LtiConsumer database to LtiConsumerXBlock.
```
feanil:
```suggestion
This migration will copy config_id from LtiConsumer database to LtiPassport table.
```

lti_consumer/lti_xblock.py
```diff
@@ -1833,3 +1847,20 @@ def index_dictionary(self):
...
+
+        try:
+            from .models import LtiConfiguration  # pylint: disable=import-outside-toplevel
+
+            configuration = LtiConfiguration.objects.select_related("lti_1p3_passport").get(
+                location=self.scope_ids.usage_id,
+            )
+            if configuration.lti_1p3_passport:
+                node.set("lti_1p3_passport_id", str(configuration.lti_1p3_passport.passport_id))
+        except (LtiConfiguration.DoesNotExist, ObjectDoesNotExist):
```
feanil:
```suggestion
        except (LtiConfiguration.DoesNotExist):
```
I don't think we need the generic Object Does not exist here do we?

lti_consumer/lti_xblock.py
```diff
@@ -1833,3 +1847,20 @@ def index_dictionary(self):
...
 
         return xblock_body
+
+    def add_xml_to_node(self, node):
+        """
+        Export XBlock XML and include passport_id from database when available.
+        """
+        super().add_xml_to_node(node)
+
+        try:
```
feanil, line 1862:
```suggestion
        # The lti_1p3_passport_id XBlock field may be empty on blocks that existed before
        # the Lti1p3Passport model was introduced (migration 0021). Rather than backfilling
        # the field in the migration (which requires the XBlock runtime and can fail silently),
        # we read the authoritative passport_id from the DB at export time. This ensures that
        # when a block is duplicated or exported/imported, the receiving block's
        # lti_1p3_passport_id field is populated and can be used to find the shared passport
        # instead of creating new credentials.
        try:
```

lti_consumer/api.py
```diff
@@ -6,44 +6,95 @@
...
+    get_lti_1p3_context_types_claim,
+    get_lti_deeplinking_content_url,
 )
-from .filters import get_external_config_from_filter
 
+log = logging.getLogger(__name__)
+
+
+def _ensure_lti_passport(block, lti_config):
+    """Ensure passport is synced with block fields, creating a new one if needed."""
```
feanil: Can you add more docs here about what this function is intended to do. I think a human readable explanation of when new passports are created vs when existing passports are used would be useful.

feanil (commented):
The PR is looking really good. Mostly I think we need to add context and comments. The one big code-change that I think is worth adding is a backward migration for 0021.  Once that's in, I think we can merge this and release it.

Oh we also need to bump the version.

lti_consumer/lti_xblock.py
navinkarkera: Yes

lti_consumer/migrations/0021_create_lti_1p3_passport.py
navinkarkera: Done. f577680c158c

lti_consumer/api.py
navinkarkera: Done. 895563421769

navinkarkera:
@feanil Applied all your suggestions, also bumped the version and updated changelog.

lti_consumer/lti_xblock.py
```diff
@@ -1833,3 +1846,26 @@ def index_dictionary(self):
...
+    def add_xml_to_node(self, node):
+        """
+        # The lti_1p3_passport_id XBlock field may be empty on blocks that existed before
+        # the Lti1p3Passport model was introduced (migration 0021). Rather than backfilling
+        # the field in the migration (which requires the XBlock runtime and can fail silently),
+        # we read the authoritative passport_id from the DB at export time. This ensures that
+        # when a block is duplicated or exported/imported, the receiving block's
+        # lti_1p3_passport_id field is populated and can be used to find the shared passport
+        # instead of creating new credentials.
+        """
```
feanil, line 1859:
```suggestion
        """
        The lti_1p3_passport_id XBlock field may be empty on blocks that existed before
        the Lti1p3Passport model was introduced (migration 0021). Rather than backfilling
        the field in the migration (which requires the XBlock runtime and can fail silently),
        we read the authoritative passport_id from the DB at export time. This ensures that
        when a block is duplicated or exported/imported, the receiving block's
        lti_1p3_passport_id field is populated and can be used to find the shared passport
        instead of creating new credentials.
        """
```

We don't need comment hashes inside the docstring.

lti_consumer/migrations/0021_create_lti_1p3_passport.py
```diff
@@ -0,0 +1,88 @@
+# Generated migration for copying config_id to the LtiPassport table
+"""
+This migration will copy config_id from LtiConfiguration database to LtiPassport table.
+
+This will help us link xblocks with LtiConsumer database rows without relying on the location or usage_key of xblocks.
```
feanil:
```suggestion
Existing config_id values are reused as passport_id so that the keyset and access token URLs already registered with LTI tools continue to work without reconfiguration.
```

feanil (approved):
Last few things, then I think this is good to merge and release!

ISSUE.md
feanil, line 1: This file looks like it shouldn't be in here?

navinkarkera: Ahh, yes! Sorry, I think I saved a git issue buffer while working on another task in neovim and accidentally committed it here.

navinkarkera:
@feanil Thanks for the in depth review! I made a small change after your review here: 8e5b92616dbc 
We don't need to modify xblock fields now that we add passport_id directly to the xml.

CHANGELOG.rst
```diff
@@ -16,6 +16,12 @@ Please See the `releases tab <https://github.com/openedx/xblock-lti-consumer/rel
 Unreleased
 ~~~~~~~~~~
 
+11.0.1 - 2026-04-20
```
feanil:
```suggestion
11.0.0 - 2026-04-20
```

feanil (approved):

feanil:
@navinkarkera I'm not actually sure that we should remove the save logic completely.  Right now, the id does not get saved at all on the original block which seems like it wolud lead to future confusion.  I think it's worth making sure that when we instantiate the LtiPassport, that it's `passport_id` makes it back to the block.  The add_xml_to_node fix generally bypasses this now but if someone makes changes there assuming the id is also on the block, it could lead to errors in the future.

What do you think? I think this is okay to land as is also if you disagree but given how hard it is to reason through this code right now, my inclination is to make it more robust and unsurprising where possible.

navinkarkera:
@feanil I wanted to avoid calling save_xblock at that point as it is also called in LMS but I agree that it could be confusing in future. I'll revert the changes.

feanil:
Right, it makes sense to not have that call in the LMS and I guess it's really hard right now to differentiate the call from the LMS vs an instantiation of the block in studio?  The critical thing is to save the id to the block when we create the new instance of the block in studio.  This happens on copy paste but not on other instantiations.

feanil:
Merged and is being released now: https://github.com/openedx/xblock-lti-consumer/releases/tag/v11.0.0

PR #639, opened 2026-04-06
build: Fail when we can't load common common_constraints.

feanil:
If we can't get the common_constraints.txt file from edx-lint it is
safer to fail than to pretend the file is empty as there are global
constraints which we want to make sure don't get missed.

See https://github.com/openedx/xblock-lti-consumer/pull/635 for what can happen without this fix.


navinkarkera (approved):
@feanil Thanks!

Faraz32123 (approved):

PR #701, opened 2026-09-10
fix: publish zero value AGS scores to the gradebook

### LINKED ISSUE #695: Zero value LTI AGS scores don't get published to Open edX gradebook
### Description

When an LTI tool submits a valid AGS score with ``scoreGiven: 0.0``, Open edX accepts the score request (201 Created) but does not publish the result to the LMS gradebook.

The grade publishing condition currently checks ``instance.score_given`` as a truthy value. Since 0.0 is falsy in Python, a valid zero score is treated as though no score was supplied.

Expected ### behavior

A score with:
{
  "scoreGiven": 0.0,
  "scoreMaximum": 100.0,
  "gradingProgress": "FullyGraded"
}
should be written to the Open edX gradebook as 0%, provided the line item is linked to an LTI block that accepts grades.

### Actual behavior
- AGS score request returns 201.
- No LMS grade-publishing event is emitted.
- The learner’s gradebook score is not updated to zero.

### Logs
```
lms-1  | 2026-08-24 12:29:52,595 INFO 12 [tracking] [user None] [ip 50.28.39.106] logger.py:41 - {"name": "/api/lti_consumer/v1/lti/73/lti-ags/19/scores", "context": {"user_id": null, "path": "/api/lti_consumer/v1/lti/73/lti-ags/19/scores", "course_id": "", "org_id": "", "enterprise_uuid": ""}, "username": "", "session": "", "ip": "50.28.39.106", "agent": "saLTIre/3.3.2 (from ceLTIc Project)", "host": "edl-stage.edly.io", "referer": "", "accept_language": "", "event": "{"GET": {}, "POST": {}}", "time": "2026-08-24T12:29:52.594858+00:00", "event_type": "/api/lti_consumer/v1/lti/73/lti-ags/19/scores", "event_source": "server", "page": null}
lms-1  | 2026-08-24 12:29:52,601 INFO 12 [lti_consumer.plugin.views] [user None] [ip 50.28.39.106] views.py:791 - LTI AGS request received: action=scores method=POST lti_config_id=73 line_item_id=19 payload_keys=['timestamp', 'userId', 'comment', 'activityProgress', 'gradingProgress', 'scoreGiven', 'scoreMaximum'] resourceId=missing resourceLinkId=missing scoreMaximum=present userId=present scoreGiven=present activityProgress=present gradingProgress=present comment=blank.
lms-1  | 2026-08-24 12:29:52,874 INFO 12 [lti_consumer.plugin.views] [user None] [ip 50.28.39.106] views.py:821 - LTI AGS response returned: action=scores method=POST lti_config_id=73 line_item_id=19 status_code=201 error=None.
lms-1  | [pid: 12|app: 0|req: 2044/5084] 172.18.0.5 () {44 vars in 1674 bytes} [Mon Aug 24 12:29:52 2026] POST /api/lti_consumer/v1/lti/73/lti-ags/19/scores => generated 204 bytes in 306 msecs (HTTP/1.1 201) 8 headers in 569 bytes (1 switches on core 0)
```

taimoor-ahmed-1:
## Description

An LTI tool that posts a valid AGS score of `scoreGiven: 0.0` gets a `201 Created`, but the zero never reaches the Open edX gradebook. `publish_grade_on_score_update` gated grade publishing on `instance.score_given` being *truthy*, so `0.0` was indistinguishable from a missing score and the signal returned without publishing anything.

The same falsy-zero pattern turns out to appear in three more places, all fixed here:

- **`LtiAgsScore.clean()`** skipped its "`scoreMaximum` is required" check for a zero score, and only ever rejected a `scoreMaximum` of `None`. A `scoreMaximum` of `0` is not a usable denominator either: it could be saved through the ORM or the Django admin, was then silently skipped by the publishing signal (the normalized-score division would otherwise raise), yet was still reported as a successful result by `LtiAgsResultSerializer.get_resultMaximum`.
- **`LtiAgsScoreSerializer.validate_scoreMaximum`** rejected `scoreMaximum: 0` only incidentally, through a truthiness check, and reported it as "is a required field" when the field was in fact present.
- **`LtiConsumerXBlock._result_service_get`** omitted a `module_score` of `0` from the LTI 1.1 result response, so a learner scored zero was reported as ungraded. Same bug, LTI 1.1 side.

While in here, the publishing signal also gained logging for *why* a final score was not published — grading progress, resource link id, score given/maximum, and the block's `has_score`/past-due state. Previously only a successful publish was ever logged, which is what made this class of bug invisible in production logs. Interim `Pending`/`InProgress` saves are still skipped silently, since those are routine and would flood the logs.

## Fixes

Fixes #695

## Testing

New tests, all failing on `master` and passing here:

- `test_grade_publish_with_zero_score` — a `FullyGraded` score of `0` out of `100` is published to the block as `0`.
- `test_xblock_grade_publish_with_zero_score` — the same thing end-to-end through the AGS scores endpoint.
- `test_grade_publish_not_done_when_score_maximum_zero` — `scoreMaximum: 0` with a set `scoreGiven` is rejected at save time and never reaches the signal.
- `test_create_score_with_zero_score_maximum` — the API rejects it with the "must be a positive number" message.
- `test_no_score_max_fails_when_setting_zero_score`, `test_score_max_fails_when_zero_with_score_given_set` — model-level assertions for the two `clean()` gaps.
- `test_consumer_get_result_called_with_zero_score` — LTI 1.1 result service reports a score of `0`.

Plus `test_grade_publish_not_done_when_score_given_missing`, a lock-in test that an AGS "erase score" request (explicit `null` `scoreGiven`) is still skipped rather than published as zero.

Full suite (773 tests) passes; `pycodestyle` and `pylint` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -16,6 +16,25 @@ Please See the `releases tab <https://github.com/openedx/xblock-lti-consumer/rel
 Unreleased
 ----------
 
+11.4.2 - 2026-09-10
+--------------------
+* Publish AGS scores with a ``scoreGiven`` of ``0`` to the LMS gradebook. The
+  grade publishing signal tested ``scoreGiven`` for truthiness, so a valid zero
+  score was treated as a missing one and never reached the gradebook.
+* Reject a ``scoreMaximum`` that is unset or not positive whenever ``scoreGiven``
+  is present. ``LtiAgsScore.clean()`` had the same truthiness bug, and a
+  ``scoreMaximum`` of ``0`` is not a usable denominator: it could be saved through
+  the ORM or admin, was then silently skipped by the publishing signal, yet was
+  still reported as a successful result by ``LtiAgsResultSerializer``.
+* Reject an AGS ``scoreMaximum`` of ``0`` at the API too, with a "must be a positive
+  number" error rather than the misleading "is a required field".
+* Report an LTI 1.1 ``module_score`` of ``0`` from the result service instead of
+  omitting it, which reported a learner scored zero as ungraded. Same falsy-zero
+  pattern, on the LTI 1.1 side.
+* Log the reason a grade publish was skipped -- grading progress, resource link id,
+  score given/maximum, and the block's ``has_score``/past-due state -- instead of
+  only ever logging a successful publish.
+
 11.4.1 - 2026-09-02
 --------------------
 * Stop writing the LTI 1.3 passport id back to the XBlock from the
--- a/lti_consumer/models.py
+++ b/lti_consumer/models.py
@@ -947,8 +947,15 @@ class LtiAgsScore(models.Model):
     def clean(self):
         super().clean()
 
-        # 'scoreMaximum' represents the denominator and MUST be present when 'scoreGiven' is present
-        if self.score_given and self.score_maximum is None:
+        # 'scoreMaximum' represents the denominator and MUST be a usable, positive value when
+        # 'scoreGiven' is present -- `score_maximum <= 0` can't be published either (see
+        # `publish_grade_on_score_update`), and letting it through here made it possible for
+        # `LtiAgsResultSerializer.get_resultMaximum` to report a fabricated "successful" result
+        # for a score that was never actually published to the gradebook.
+        # Note: this must be an `is not None` check on `score_given`, not a truthiness check,
+        # since a `scoreGiven` of 0 is falsy but a legitimate value that still requires a
+        # `scoreMaximum`.
+        if self.score_given is not None and not self.score_maximum:
             raise ValidationError({'score_maximum': 'cannot be unset when score_given is set'})
 
     def save(self, *args, **kwargs):
--- a/lti_consumer/signals/signals.py
+++ b/lti_consumer/signals/signals.py
@@ -44,17 +44,54 @@ def publish_grade_on_score_update(sender, instance, **kwargs):  # pylint: disabl
         )
         return
 
-    # Before starting to publish grades to the LMS, check that:
-    # 1. The grade being submitted in the final one - `FullyGraded`
-    # 2. This LineItem is linked to a LMS grade - the `LtiResouceLinkId` field is set
-    # 3. There's a valid grade in this score - `scoreGiven` is set
-    if instance.grading_progress == LtiAgsScore.FULLY_GRADED \
-            and line_item.resource_link_id \
-            and instance.score_given:
-        try:
-            # Load block using LMS APIs and check if the block is graded and still accept grades.
-            block = compat.load_block_as_user(line_item.resource_link_id)
-            if block.has_score and (not block.is_past_due() or block.accept_grades_past_due):
+    # The grade being submitted must be the final one - `FullyGraded`. This is routine, expected
+    # tool behavior to not be the case yet (interim `Pending`/`InProgress` saves are far more
+    # common than the final one), so it's silently skipped here, not logged, to avoid flooding
+    # the logs.
+    if instance.grading_progress != LtiAgsScore.FULLY_GRADED:
+        return
+
+    # From here on the score is final, so failing to publish it is the actionable case worth
+    # logging below. Before publishing to the LMS, check that:
+    # 1. This LineItem is linked to a LMS grade - the `LtiResouceLinkId` field is set
+    # 2. There's a grade present in this score - `scoreGiven` is present
+    # 3. `scoreMaximum` is a usable, positive denominator for the normalized score below
+    # Note: (2) and (3) must be `is not None`/range checks, not truthiness checks, since a
+    # `scoreGiven` or `scoreMaximum` of 0 is falsy but a legitimate value for the former.
+    can_publish = (
+        bool(line_item.resource_link_id) and
+        instance.score_given is not None and
+        instance.score_maximum is not None and
+        instance.score_maximum > 0
+    )
+    if not can_publish:
+        log.info(
+            "LTI AGS grade publish skipped: score=%r grading_progress=%s resource_link_id=%s "
+            "score_given=%s score_maximum=%s.",
+            instance,
+            instance.grading_progress,
+            line_item.resource_link_id,
+            instance.score_given,
+            instance.score_maximum,
+        )
+        return
+
+    try:
+        # Load block using LMS APIs and check if the block is graded and still accept grades.
+        block = compat.load_block_as_user(line_item.resource_link_id)
+        if not block.has_score:
+            log.info(
+                "LTI AGS grade publish skipped: score=%r resource_link_id=%s has_score=%s.",
+                instance,
+                line_item.resource_link_id,
+                block.has_score,
+            )
+        else:
+            # Computed once and reused below: previously `is_past_due()`/`accept_grades_past_due`
+            # were never evaluated at all when `has_score` was False (short-circuited by `and`),
+            # so neither is touched here unless `has_score` is True.
+            is_past_due = block.is_past_due()
+            if not is_past_due or block.accept_grades_past_due:
                 # Map external ID to platform user
                 user = compat.get_user_from_external_user_id(instance.user_id)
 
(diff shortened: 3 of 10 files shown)


CHANGELOG.rst
```diff
@@ -16,6 +16,25 @@ Please See the `releases tab <https://github.com/openedx/xblock-lti-consumer/rel
...
 
+11.4.2 - 2026-09-10
+--------------------
+* Publish AGS scores with a ``scoreGiven`` of ``0`` to the LMS gradebook. The
+  grade publishing signal tested ``scoreGiven`` for truthiness, so a valid zero
+  score was treated as a missing one and never reached the gradebook.
+* Reject a ``scoreMaximum`` that is unset or not positive whenever ``scoreGiven``
+  is present. ``LtiAgsScore.clean()`` had the same truthiness bug, and a
+  ``scoreMaximum`` of ``0`` is not a usable denominator: it could be saved through
+  the ORM or admin, was then silently skipped by the publishing signal, yet was
```
feanil: `scoreMaximum: 0` was not silently skipped on master.  There is no `score_maximum` guard in `publish_grade_on_score_update` there, so `min(score_given, 0) / 0` raises `ZeroDivisionError`, the broad `except` re-raises it out of `post_save`, and the AGS POST 500s.

I confirmed this by running `test_grade_publish_not_done_when_score_maximum_zero` with the four source files reverted to master `66ee48da`: it errors with `ZeroDivisionError: float division by zero` at `signals.py:64`, it does not skip.

Say that instead.  The same claim is in the PR description and in the new comment at `models.py:951-953`.

CHANGELOG.rst
```diff
@@ -16,6 +16,25 @@ Please See the `releases tab <https://github.com/openedx/xblock-lti-consumer/rel
...
+  still reported as a successful result by ``LtiAgsResultSerializer``.
+* Reject an AGS ``scoreMaximum`` of ``0`` at the API too, with a "must be a positive
+  number" error rather than the misleading "is a required field".
+* Report an LTI 1.1 ``module_score`` of ``0`` from the result service instead of
```
feanil: Name the handler.  An LTI 1.1 block exposes two grade-return surfaces and "the result service" does not say which one changed: `outcome_service_handler`, the Basic Outcomes XML service in `outcomes.py`, and `result_service_handler`, the LTI 2.0 JSON/REST result service that `_result_service_get` backs.  This fix is in the second one; the first passes the score through unconditionally at `outcomes.py:206` and never had the bug.

lti_consumer/models.py
```diff
@@ -947,8 +947,15 @@ class LtiAgsScore(models.Model):
     def clean(self):
         super().clean()
 
-        # 'scoreMaximum' represents the denominator and MUST be present when 'scoreGiven' is present
-        if self.score_given and self.score_maximum is None:
+        # 'scoreMaximum' represents the denominator and MUST be a usable, positive value when
```
feanil: Nine lines of comment for a two-line check, and most of it is the history of the bug rather than the rule the code enforces.  The history belongs in the commit message, where it stays useful after "previously" stops meaning anything.  Something like:

```python
        # 'scoreMaximum' is the denominator, so it has to be present and positive whenever
        # 'scoreGiven' is present.  'scoreGiven' is tested with `is not None` because 0 is
        # a valid grade; 'scoreMaximum' is tested for falsiness because both None and 0
        # are rejected.
        if self.score_given is not None and not self.score_maximum:
```

Same pass on `signals.py:47-50`, `signals.py:54-60`, `signals.py:90-92`, and the new docstrings in `test_signals.py` and `test_models.py`.  Several of those docstrings are longer than the tests under them.

lti_consumer/models.py
```diff
@@ -947,8 +947,15 @@ class LtiAgsScore(models.Model):
...
+        # 'scoreGiven' is present -- `score_maximum <= 0` can't be published either (see
+        # `publish_grade_on_score_update`), and letting it through here made it possible for
+        # `LtiAgsResultSerializer.get_resultMaximum` to report a fabricated "successful" result
+        # for a score that was never actually published to the gradebook.
+        # Note: this must be an `is not None` check on `score_given`, not a truthiness check,
+        # since a `scoreGiven` of 0 is falsy but a legitimate value that still requires a
+        # `scoreMaximum`.
+        if self.score_given is not None and not self.score_maximum:
             raise ValidationError({'score_maximum': 'cannot be unset when score_given is set'})
```
feanil: Update the message.  It now fires for `score_maximum=0`, which is set, not unset.  `serializers.py:155` got "must be a positive number" for the same condition; use that wording here too.

lti_consumer/signals/signals.py
```diff
@@ -44,17 +44,54 @@ def publish_grade_on_score_update(sender, instance, **kwargs):  # pylint: disabl
...
+    # `scoreGiven` or `scoreMaximum` of 0 is falsy but a legitimate value for the former.
+    can_publish = (
+        bool(line_item.resource_link_id) and
+        instance.score_given is not None and
+        instance.score_maximum is not None and
+        instance.score_maximum > 0
+    )
+    if not can_publish:
+        log.info(
+            "LTI AGS grade publish skipped: score=%r grading_progress=%s resource_link_id=%s "
```
feanil: Three of these fields are already in `score=%r`.  `LtiAgsScore.__str__` (`models.py:966-971`) renders as `LineItem 7: score 0.0 out of 100.0 - FullyGraded`, so `grading_progress`, `score_given` and `score_maximum` each land in the line twice.  `grading_progress` is also `FullyGraded` by construction here, since the check above returned otherwise.  Drop the three duplicates and keep `resource_link_id`.

lti_consumer/tests/unit/test_signals.py
```diff
@@ -118,6 +119,106 @@ def test_grade_publish(self):
...
+            activity_progress=LtiAgsScore.COMPLETED,
+            grading_progress=LtiAgsScore.FULLY_GRADED,
+            user_id="test",
+            timestamp=datetime.now(),
+        )
+
+        self._block_mock.set_user_module_score.assert_not_called()
+        self._signals_compat_mock.load_block_as_user.assert_not_called()
+
+    def test_grade_publish_not_done_when_score_maximum_zero(self):
```
feanil: This asserts model validation, not signal behavior.  `LtiAgsScore.objects.create` raises inside `save()` -> `full_clean()`, before `post_save` fires, so the signal never runs and `test_models.py::test_score_max_fails_when_zero_with_score_given_set` already covers the same thing.  Move it or rename it for what it checks.

Related: because `save()` calls `full_clean()`, the new `clean()` makes the `score_maximum is not None and score_maximum > 0` half of `can_publish` unreachable.  Keep it as defence in depth if you want, but the docstring describing it as a workaround master already had is wrong.

CHANGELOG.rst
feanil, line 27: `scoreMaximum: 0` was not silently skipped on master.  There is no `score_maximum` guard in `publish_grade_on_score_update` there, so `min(score_given, 0) / 0` raises `ZeroDivisionError`, the broad `except` re-raises it out of `post_save`, and the AGS POST 500s.

I confirmed this by running `test_grade_publish_not_done_when_score_maximum_zero` with the four source files reverted to master `66ee48da`: it errors with `ZeroDivisionError: float division by zero` at `signals.py:64`, it does not skip.

Say that instead.  The same claim is in the PR description and in the new comment at `models.py:951-953`, and the docstring at `test_signals.py:191-192` goes further and says master already had a `score_maximum > 0` guard in the signal.

CHANGELOG.rst
feanil, line 31: Name the handler.  An LTI 1.1 block exposes two grade-return surfaces and "the result service" does not say which one changed: `outcome_service_handler`, the Basic Outcomes XML service in `outcomes.py`, and `result_service_handler`, the LTI 2.0 JSON/REST result service that `_result_service_get` backs.  This fix is in the second one; the first passes the score through unconditionally at `outcomes.py:206` and never had the bug.

lti_consumer/models.py
feanil, line 950: Nine lines of comment for a two-line check, and most of it is the history of the bug rather than the rule the code enforces.  The history belongs in the commit message, where it stays useful after "previously" stops meaning anything.  Something like:

```python
        # 'scoreMaximum' is the denominator, so it has to be present and positive whenever
        # 'scoreGiven' is present.  'scoreGiven' is tested with `is not None` because 0 is
        # a valid grade; 'scoreMaximum' is tested for falsiness because both None and 0
        # are rejected.
        if self.score_given is not None and not self.score_maximum:
```

Same pass on `signals.py:47-50`, `signals.py:54-60`, `signals.py:90-92`, and the new docstrings in `test_signals.py` and `test_models.py`.  Several of those docstrings are longer than the tests under them.

lti_consumer/models.py
feanil, line 959: Update the message.  It now fires for `score_maximum=0`, which is set, not unset.  `serializers.py:155` got "must be a positive number" for the same condition; use that wording here too.

lti_consumer/signals/signals.py
feanil, line 69: Three of these fields are already in `score=%r`.  `LtiAgsScore.__str__` (`models.py:966-971`) renders as `LineItem 7: score 0.0 out of 100.0 - FullyGraded`, so `grading_progress`, `score_given` and `score_maximum` each land in the line twice.  `grading_progress` is also `FullyGraded` by construction here, since the check above returned otherwise.  Drop the three duplicates and keep `resource_link_id`.

PR #703, opened 2026-09-13
fix: deny NRPS memberships endpoint when NRPS is disabled

### LINKED ISSUE #690: NRPS can get response without NRPS being allowed by the XBlock
Tools can call and get response from `memberships` endpoint even when the XBlock has NRPS disabled. 

All they need to know if the exact endpoint. 

In the example below, saltire is acting as a tool (1st screenshot). If i provide it with the endpoint `https://master.openedx.io/api/lti_consumer/v1/lti/31/memberships`, even as NRPS is disabled in the XBlock (2nd screenshot), it still gets back the memberships (3rd screenshot).

Image

Image

Image

cpruijsen:
The NRPS `memberships` endpoint returns data even when NRPS is disabled on the XBlock. A tool holding
a correctly scoped token can read the course membership list from a block whose configuration says
that service is off, so the setting does not actually gate access to the roster.

`LtiNrpsContextMembershipsPermissions` in `lti_consumer/lti_1p3/extensions/rest_framework/permissions.py` only checked token scopes. Scope says what the token is allowed
to ask for; it says nothing about whether this configuration offers the service. This adds a
`has_permission` that returns False when `get_lti_advantage_nrps_enabled()` is False, before
delegating to the scope check, so both have to hold.

The test asserts the 403 and that `get_course_members` is never called, so it fails if the endpoint
starts returning early for some other reason while still reaching the roster.

Fixes #690


mphilbrick211:
Hi @cpruijsen! Welcome, and thank you for this contribution! In order for your CLA check to turn green, you'll need to submit a CLA form. If you are contributing as an individual, please fill out the individual [CLA form here](https://docs.openedx.org/en/latest/developers/quickstarts/so_you_want_to_contribute.html#id175).

If you are contributing on behalf of an organization, please have your manager reach out to oscm@axim.org so you may be added to your org's existing entity agreement.

Please let me know if you have any questions. Thanks!

e0d:
@ayub02 FYI.

